PR #2, opened 2022-01-28
paymentservice: code polishes

marcelosalloum:
### What

Polished code from the `paymentservice` module in order to make it ready for the next features.

Changes:
- Removed `AccountType` to use the new object `Account.Capabilities`, for improved clarity.
- Fix async test `testGetAccount_isNotMainAccount` that was failing intermittently.
- Update `sendPayment` signature so it can represent payments from one network to another, e.g. `Circle->Stellar`.
- Rename package `paymentsservice` to use the singular form `paymentservice`.
- Add or update comments to improve clarity.


src/main/java/org/stellar/anchor/paymentsservice/Balance.java
```diff
@@ -5,8 +5,20 @@
 @Data
 public class Balance {
     String amount;
+    /**
+     * The name of the currency that will be ultimately credited into the beneficiary user account. It should obey the
+     * {scheme}:{identifier} format described in <a href="https://stellar.org/protocol/sep-38#asset-identification-format">SEP-38</a>.
+     */
```
JakeUrban, line 11: ❓ Are Java comments HTML?

src/main/java/org/stellar/anchor/paymentsservice/PaymentsService.java
```diff
@@ -80,15 +80,17 @@
...
+     *                           addressTag) will be mandatory.
+     * @param currencyName       the name of the currency used in the payment. It should obey the {scheme}:{identifier}
+     *                           format described in <a href="https://stellar.org/protocol/sep-38#asset-identification-format">SEP-38</a>.
+     * @param amount             the payment amount.
      * @return asynchronous stream with the payment object.
+     * @throws HttpException If the provided input parameters are invalid.
      * @throws HttpException If the http response status code is 4xx or 5xx.
      */
-    Mono<Payment> sendPayment(String sourceAccountId, String destinationAccountId, String currencyName, String amount) throws HttpException;
+    Mono<Payment> sendPayment(Account sourceAccount, Account destinationAccount, String currencyName, Float amount) throws HttpException;
```
JakeUrban, line 93: Amounts where precision matters are generally not handling using floating point numbers. Does Java have a BigNumber or Decimal type that is better suited for this variable?

src/main/java/org/stellar/anchor/paymentsservice/Account.java
```diff
@@ -16,7 +16,7 @@
 
     public Network network;
 
-    public AccountType type;
+    public AccountLevel level;
```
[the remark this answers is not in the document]
marcelosalloum: My rationale was influenced by how Circle deals with accounts: there are distribution accounts and default ones, where the distribution account is the one connected with a bank account (and can make/receive bank payments) while the default ones are regular virtual accounts.

My understanding was that one of them had a higher level than the other, that's why I renamed it globally, but I might be wrong here. Maybe AccountType is more generic and scalable, WDYT? My intention here is to have a way to differentiate the regular accounts from the outliers, like distribution accounts.

src/main/java/org/stellar/anchor/paymentsservice/Balance.java
marcelosalloum: They support HTML to some extent. I've only used it for external links `<a href=...</a>` and for [code snippets](https://github.com/stellar/java-stellar-anchor-dev/blob/1a7268657a335397961629ac19e1dee73e575301/src/main/java/org/stellar/anchor/paymentsservice/PaymentsService.java#L99-L113).

In [this](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html) blog post there are some examples of HTML being used to organize the text with paragraphs `<p>` and bullet points `<li>` as well.

src/main/java/org/stellar/anchor/paymentsservice/Account.java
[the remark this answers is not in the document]
JakeUrban: I think `Type` is more generic, but I don't feel strongly about this.

Will the payments service package eventually support sending and receiving from bank accounts, and if so, will we limit this functionality to a subset of `AccountLevel`'s? I can see this service having many more account levels/types than Circle, since this service will include more payment systems with differing sets of functionality.

src/main/java/org/stellar/anchor/paymentsservice/PaymentsService.java
JakeUrban: Looks like [BigDecimal](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html) is an option.

lijamie98:
LGTM since there is no Circle specific implementations. I think Jake can do a better review on the payment gateway interface.

lijamie98 (approved):

src/main/java/org/stellar/anchor/paymentsservice/Account.java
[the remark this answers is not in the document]
marcelosalloum: > Will the payments service package eventually support sending and receiving from bank accounts

Yes.

> and if so, will we limit this functionality to a subset of AccountLevel's?

Firstly, it will be limited by network, accounts from the Stellar network can't send bank payments for instance.
Within Circle, it's limited by account level as well, but the limitation could be different in other networks.

> I can see this service having many more account levels/types than Circle, since this service will include more payment systems with differing sets of functionality.

Giving it another thought... it could be more intuitive to have a `capabilities` object with values like these:
```js
{ (...)
  "capabilities": {
    "send": {
      "circle": true,
      "stellar": true,
      "wire": true,
      "foo": false
    },
    "receive": {
      "circle": true,
      "stellar": true,
      "wire": true,
      "foo": false
    }
  }
(...) }
```

WDYT?

src/main/java/org/stellar/anchor/paymentsservice/PaymentsService.java
lijamie98: Yes, BigDecimal is the right way to go.

marcelosalloum: Done in https://github.com/stellar/java-stellar-anchor-sdk/pull/2/commits/75e3e6ffb8cbe1066b5dc46027391ef9c3d2009c

src/main/java/org/stellar/anchor/paymentsservice/Account.java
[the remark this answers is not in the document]
marcelosalloum: Replacing in https://github.com/stellar/java-stellar-anchor-sdk/pull/2/commits/38ca76c3fb8ceb0d136b85558fbd491a7a47c568.

src/main/java/org/stellar/anchor/paymentsservice/Account.java
```diff
@@ -1,9 +1,12 @@
 package org.stellar.anchor.paymentsservice;
```
lijamie98, line 1: Can we use org.stellar.anchor.paymentservice?

src/main/java/org/stellar/anchor/paymentsservice/Account.java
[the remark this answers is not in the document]
JakeUrban: Sorry I've been in meetings. This looks great, much more flexible 👍

JakeUrban (approved):

src/main/java/org/stellar/anchor/paymentsservice/Account.java
marcelosalloum: Thanks for pointing that out! Renamed in https://github.com/stellar/java-stellar-anchor-sdk/pull/2/commits/4d89efa47b70e283b7b29830658ade547009e7ef

PR #3, opened 2022-01-28
paymentsservice: implement sendPayment to send `Circle->Circle` or `Circle->Stellar` payments

marcelosalloum:
### What
Implement `POST /v1/transfer` request that supports `Circle->Circle` and `Circle->Stellar` payments.

### DIFF
--- a/src/main/java/org/stellar/anchor/paymentservice/Account.java
+++ b/src/main/java/org/stellar/anchor/paymentservice/Account.java
@@ -27,15 +27,15 @@ public class Account {
     @NonNull
     public Account.Capabilities capabilities;
 
+    @NonNull
+    public List<Balance> balances = new ArrayList<>();
+
     /**
      * The list of not-yet-available balances that are expected to settle shortly. These balances could be cancelled or
      * returned, in which cases they may never become available in the user account.
      */
     @NonNull
     public List<Balance> unsettledBalances = new ArrayList<>();
-
-    @NonNull
-    public List<Balance> balances = new ArrayList<>();
 
     public Account(@NonNull Network network, @NonNull String id, @Nullable String idTag, @NonNull Account.Capabilities capabilities) {
         this.network = network;
@@ -82,15 +82,7 @@ public class Account {
             this(List.of(sendAndReceive), List.of(sendAndReceive));
         }
 
-        public void setSendSupport(Network network, Boolean supportEnabled) {
-            this.send.put(network, supportEnabled);
-        }
-
-        public void setReceiveSupport(Network network, Boolean supportEnabled) {
-            this.receive.put(network, supportEnabled);
-        }
-
-        public void setFullSupport(Network network, Boolean supportEnabled) {
+        public void setCapabilities(Network network, Boolean supportEnabled) {
             this.send.put(network, supportEnabled);
             this.receive.put(network, supportEnabled);
         }
--- a/src/main/java/org/stellar/anchor/paymentservice/Account.java
+++ b/src/main/java/org/stellar/anchor/paymentservice/Account.java
@@ -82,17 +82,20 @@ public class Account {
             this(List.of(sendAndReceive), List.of(sendAndReceive));
         }
 
-        public void setCapabilities(Network network, Boolean supportEnabled) {
-            this.send.put(network, supportEnabled);
-            this.receive.put(network, supportEnabled);
-        }
-
         @Override
         public boolean equals(Object o) {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
             Capabilities that = (Capabilities) o;
-            return this.send.keySet().equals(that.send.keySet()) && this.receive.keySet().equals(that.receive.keySet());
+            return this.send.equals(that.send) && this.receive.equals(that.receive);
+        }
+
+        @Override
+        public String toString() {
+            return "Capabilities{" +
+                    "send=" + send +
+                    ", receive=" + receive +
+                    '}';
         }
     }
 }
--- a/src/main/java/org/stellar/anchor/paymentservice/Account.java
+++ b/src/main/java/org/stellar/anchor/paymentservice/Account.java
@@ -2,6 +2,8 @@ package org.stellar.anchor.paymentservice;
 
 import lombok.Data;
 import lombok.Getter;
+import reactor.util.annotation.Nullable;
+import reactor.util.annotation.NonNull;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


lijamie98:
Marcelo, this PR depends on how the anchor may integrate with the SDK and Circle is MGI specific (although other anchors may choose to use Circle).

It is ok if we merge in the SDK in the beginning. I am working on some code structural change and we will need to move this implementation to another module.

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
```diff
@@ -257,6 +260,73 @@ private HttpClient getWebClient(boolean authenticated) {
...
+        if (sourceAccount.network != Network.CIRCLE) {
+            throw new HttpException(400, "the only supported network for the source account is circle");
+        }
+        if (!List.of(Network.CIRCLE, Network.STELLAR).contains(destinationAccount.network)) {
+            throw new HttpException(400, "the only supported networks for the destination account are circle and stellar");
+        }
+        if (!currencyName.startsWith(destinationAccount.network.getCurrencyPrefix())) {
+            throw new HttpException(400, "the currency to be sent must contain the destination network schema");
+        }
+    }
```
JakeUrban, line 285: Do we want to validate the sending and receiving account addresses here? 

For example, what if we try to send a payment from Circle to a Stellar account that doesn't exist or doesn't have a trustline? In the opposite direction, what if we try to send to a Circle wallet but the we use a memo that wasn't generated using Circle's `POST /addresses` endpoint?

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
```diff
@@ -257,6 +260,73 @@ private HttpClient getWebClient(boolean authenticated) {
...
+                .responseSingle((postResponse, bodyBytesMono) -> {
+                    if (postResponse.status().code() >= 400) {
+                        return handleCircleError(postResponse, bodyBytesMono);
+                    }
+
+                    return bodyBytesMono.asString();
+                }).flatMap(body -> {
+                    CircleTransferResponse transfer = gson.fromJson(body, CircleTransferResponse.class);
+                    return Mono.just(transfer.getData().toPayment());
+                });
```
JakeUrban, line 327: Are we certain that all responses with a status code below 400 contains the expected json body? We may want to have some logic that handles when an unexpected response is returned.

src/main/java/org/stellar/anchor/paymentservice/circle/model/CirclePaymentStatus.java
```diff
@@ -0,0 +1,41 @@
+package org.stellar.anchor.paymentservice.circle.model;
+
+import com.google.gson.annotations.SerializedName;
+import lombok.Getter;
+import org.stellar.anchor.paymentservice.Payment;
+
+@Getter
+public enum CirclePaymentStatus {
```
JakeUrban, line 8: Do we want to store any kind of error message or result code? When I use Circle's APIs and get failed transactions, there is usually an error code associated with it.

src/main/java/org/stellar/anchor/paymentservice/Account.java
```diff
@@ -60,8 +78,21 @@ public Capabilities(List<Network> send, List<Network> receive) {
...
+            this.send.put(network, supportEnabled);
+            this.receive.put(network, supportEnabled);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Capabilities that = (Capabilities) o;
+            return this.send.keySet().equals(that.send.keySet()) && this.receive.keySet().equals(that.receive.keySet());
```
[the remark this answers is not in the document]
marcelosalloum: Thanks for spotting that! Fixed in https://github.com/stellar/java-stellar-anchor-sdk/pull/3/commits/b62a8e30f1bfe9353ab2de44d5e71fa9161d12dd.

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
marcelosalloum: 🤔  These are great points! My rationale was to avoid overloading the SDK with unnecessary complexity and simply handle an error when it occurs due to a misconfiguration.

> For example, what if we try to send a payment from Circle to a Stellar account that doesn't exist or doesn't have a trustline? 

That's a great point, depending on how much we want to validate the accounts before sending a payment, this could scale further:
* For a Circle sender, should we validate it exists in the Circle network? Should we verify if the balance is enough?
* For a Circle receiver, should we check if it exists?
* For a Stellar receiver, should we check if it exists and if it has the trustline needed?
* As we expand the support for other networks, we will need to add to these verifications.

I'd address this in another PR if we decide to go through this route.

> In the opposite direction, what if we try to send to a Circle wallet but the we use a memo that wasn't generated using Circle's POST /addresses endpoint?

I don't think this is a use case we need to handle because these memos are used to receive payments from outside Circle. if we need to send a payment from a Circle wallet to another circle wallet, we would use an [internal transfer](https://developers.circle.com/reference#accounts-transfers-create).

src/main/java/org/stellar/anchor/paymentservice/circle/model/CirclePaymentStatus.java
marcelosalloum: The error message is being parsed from the transfer response and included in the Payment object:
https://github.com/stellar/java-stellar-anchor-sdk/blob/8f5f0bc22a75745b485c5032d32c5f8b8599cc46/src/main/java/org/stellar/anchor/paymentservice/circle/model/CircleTransfer.java#L20-L32

Whoever has access to the Payment object (another application or an API layer) will have access to this information.

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
marcelosalloum: You're right, we're not handling the case when the response status code is "fine" but the response body is totally unexpected.

Creating an issue for that: https://github.com/stellar/java-stellar-anchor-sdk/issues/14

lijamie98 (approved):

marcelosalloum:
@JakeUrban: I'm merging it to unblock Jamie. I can address further feedback in other PRs.

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
JakeUrban: > I'd address this in another PR if we agree to go through this route.

I think we should eventually, but we can plan to do this later in the package's development.

> I don't think this is a use case we need to handle because these memos are used to receive payments from outside Circle.

I see, so we're not supporting `Stellar->Circle` payments? Why is that?

JakeUrban:
@marcelosalloum fine by me. I added one more follow up question but otherwise it looked good.

src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentsService.java
marcelosalloum: Alright, I created this issue to follow-up on the validation: https://github.com/stellar/java-stellar-anchor-sdk/issues/16

> I see, so we're not supporting Stellar->Circle payments? Why is that?

Let's sync offline about this one, it should be faster.

marcelosalloum: Just to register what we discussed offilne: this PR is restricted to implementing the Circle API integration capable of sending `Circle->Circle` and `Circle->Stellar` transfers.

In the future, we will add a Stellar driver (likely) at `paymentservice.stellar` that's capable of doing `Stellar->Circle` payments.

PR #19, opened 2022-02-01
Gradle: multi-project layout

lijamie98:
<details>
  <summaryConvert to Gradle multi-project layout.</summary>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The project is split to `core` and `payment-service` sub-projects to separate the implementation and the sdk-core business logic. 

### Why

This is to simplify the project build process and to separate the SDK to reduce the possibility of transitive dependency conflicts.


JakeUrban (changes requested):
I would rename the `circle-payment` project directory to just `paymentservice`, since the payment service will include other rails soon like Stellar, PIX, etc.

lijamie98:
> I would rename the `circle-payment` project directory to just `paymentservice` / `payment-service`, since the payment service will include other rails soon like Stellar, PIX, etc.

This is on purpose. When we have other payment implementations, they will be `payment-stellar` and `payment-PIX`. I will rename the current one to `payment-circle`.

JakeUrban:
> When we have other payment implementations, they will be payment-stellar and payment-PIX

Oh ok, so each `payment-*` directory will have an inner `paymentservice` directory and then a directory for the type of rail? Looking at this branch, it looks like `payment-circle/.../paymentservice/circle`.

So when we add a `payment-stellar` submodule, it will have its own `paymentservice.stellar` package?

This all sounds good to me, I'm just trying to understand how the directory structure relates to package naming and structure. In python, package structure maps directly to the file structure.

lijamie98:
> > When we have other payment implementations, they will be payment-stellar and payment-PIX
> 
> Oh ok, so each `payment-*` directory will have an inner `paymentservice` directory and then a directory for the type of rail? Looking at this branch, it looks like `payment-circle/.../paymentservice/circle`.
> 
> So when we add a `payment-stellar` submodule, it will have its own `paymentservice.stellar` package?
> 
> This all sounds good to me, I'm just trying to understand how the directory structure relates to package naming and structure. In python, package structure maps directly to the file structure.

In Java, the package structure also maps to the package name.  But that only applies under the `src` (https://github.com/stellar/java-stellar-anchor-sdk/tree/main/src) folder. The `payment-circle` is a "project" name. Fortunately, it is treated differently from the package name.

JakeUrban (approved):
Thanks for clarifying. LGTM

marcelosalloum (approved):
@lijamie98: you moved `NettyHttpClient` inside `payment-circle` but my idea is to have it in `core` so it can be reused by future `payment-*` implementations.

It should be ok to move it to `core` later thought, your call.

lijamie98:
> @lijamie98: you moved `NettyHttpClient` inside `payment-circle` but my idea is to have it in `core` so it can be reused by future `payment-*` implementations.
> 
> It should be ok to move it to `core` later thought, your call.

@marcelosalloum core uses SDF's Horizon Java SDK which has transitive dependency on OkHttp3, which duplicates the NettyHttpClient. We should keep the core as small as possible. Once we create another payment-* project, I will create a payment-shared project and help to move the commonly used packages there.

PR #22, opened 2022-02-03
Add google-java-format task to compileJava task.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Apply `google-java-format` when `compileJava` task is run. 
https://github.com/google/google-java-format

Note: `compileJava` is executed when `build` is executed.

### Why

Make sure all subprojects uses the same formatter.

### Known limitations
Kotlin formatter should be added next.


core/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.js.inline.clean.removeUnusedImports
+
```
JakeUrban, line 2: Do we need this import here? I don't see it being used in the file.

JakeUrban (approved):
Love it.

core/build.gradle.kts
lijamie98: Good catch. Thank you!

Updated.

PR #26, opened 2022-02-03
circle-payment: fix `CirclePaymentSystem#sendPayment` json and other things

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Polishes in the `payment-circle` package:
- In the tests, update sendPayment result status to pending to be consistent with the real API.
- Add missing idempotency key when sending the `POST /v1/transfer` request.
- Fix payment balance's `currencyName` to contain the prefix of the destination network, as described in the code documentation ([ref1], [ref2]).
- Rename `*PaymentsService*` to its singular form `*PaymentService*`.
- Replaced unneeded usages of `flatMap` for `map`.

### Why

Regular polishing step before building new features.

[ref1]: https://github.com/stellar/java-stellar-anchor-sdk/blob/0b96af6ca4679ee936b0b888804d079529e7fcf4/core/src/main/java/org/stellar/anchor/paymentservice/Payment.java#L12-L13
[ref2]: https://github.com/stellar/java-stellar-anchor-sdk/blob/0b96af6ca4679ee936b0b888804d079529e7fcf4/core/src/main/java/org/stellar/anchor/paymentservice/Balance.java#L8-L13


payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
```diff
@@ -65,14 +65,13 @@ private HttpClient getWebClient(boolean authenticated) {
   private <T> Mono<T> handleCircleError(HttpClientResponse response, ByteBufMono bodyBytesMono) {
     return bodyBytesMono
         .asString()
-        .flatMap(
+        .map(
```
JakeUrban, line 68: ❓ Can you explain when `.flatMap()` or `.map()` is appropriate?

As I understood it, `Mono.flatMap()` was the correct usage, otherwise you should use `Object.map()`?

JakeUrban (approved):

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
marcelosalloum: In reactor objects that inherit from `Publisher`, both map and flatMap are called after the async part of the publisher finishes, but while `.map` is executed synchronously, `.flatMap` spins off a new Publisher whose execution will be async ([ref]).

[ref]: https://www.baeldung.com/java-reactor-map-flatmap

PR #29, opened 2022-02-04
platform: Create `platform` subproject

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add `platform` subproject to jumpstart the platform development. 

This is a Spring MVC implementation.

### Why

NA

### Known limitations

The JDBC in the `data` package will be extracted to another subproject.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,11 @@
 
 The Stellar Anchor SDK is a collection of projects that make easy to build financial applications on Stellar.
 
-* [core](./core) is the base package for implementing standardized anchor applications.
-* [payment-circle](./payment-circle) is a package that wraps a subset of the [Circle API](https://developers.circle.com/reference).
+* [core](./core): the base package for implementing standardized anchor applications.
+* [config-spring-property](./config-spring-property): Implementation of configuration management with [Spring ConfigurationProperties](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config).
+* [data-spring-jdbc](./data-spring-jdbc): Implementation of data access with [Spring Data JDBC](https://spring.io/projects/spring-data-jdbc). 
+* [payment-circle](./payment-circle): Implementation of payment servie with [Circle API](https://developers.circle.com/reference).
+* [platform](./platform): The anchor platform [Spring Boot Application with WebMVC](https://spring.io/guides/gs/serving-web-content/). 
 
 While the project is in its early stages and in active development, it is used in production today by businesses providing services on Stellar.
 
--- a/config-spring-property/src/main/java/org/stellar/anchor/server/config/PropertyResourceAppConfig.java
+++ b/config-spring-property/src/main/java/org/stellar/anchor/server/config/PropertyResourceAppConfig.java
@@ -1,0 +1,17 @@
+package org.stellar.anchor.server.config;
+
+import java.util.List;
+import lombok.Data;
+import org.stellar.anchor.config.AppConfig;
+
+@Data
+public class PropertyResourceAppConfig implements AppConfig {
+  private String stellarNetworkPassPhrase = "Test SDF Network ; September 2015";
+  private String hostUrl = "http://localhost:9800";
+  private String horizonURI = "https://horizon-testnet.stellar.org";
+
+  private String jwtSecretKey;
+  private String assets = "assets-test.json";
+
+  private List<String> languages;
+}
--- a/data-spring-jdbc/src/main/java/org/stellar/anchor/server/sqlite/SQLiteDialect.java
+++ b/data-spring-jdbc/src/main/java/org/stellar/anchor/server/sqlite/SQLiteDialect.java
@@ -1,0 +1,132 @@
+package org.stellar.anchor.server.sqlite;
+
+import java.sql.Types;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.function.SQLFunctionTemplate;
+import org.hibernate.dialect.function.StandardSQLFunction;
+import org.hibernate.dialect.function.VarArgsSQLFunction;
+import org.hibernate.type.StringType;
+
+public class SQLiteDialect extends Dialect {
+  public SQLiteDialect() {
+    registerColumnType(Types.BIT, "integer");
+    registerColumnType(Types.TINYINT, "tinyint");
+    registerColumnType(Types.SMALLINT, "smallint");
+    registerColumnType(Types.INTEGER, "integer");
+    registerColumnType(Types.BIGINT, "bigint");
+    registerColumnType(Types.FLOAT, "float");
+    registerColumnType(Types.REAL, "real");
+    registerColumnType(Types.DOUBLE, "double");
+    registerColumnType(Types.NUMERIC, "numeric");
+    registerColumnType(Types.DECIMAL, "decimal");
+    registerColumnType(Types.CHAR, "char");
+    registerColumnType(Types.VARCHAR, "varchar");
+    registerColumnType(Types.LONGVARCHAR, "longvarchar");
+    registerColumnType(Types.DATE, "date");
+    registerColumnType(Types.TIME, "time");
+    registerColumnType(Types.TIMESTAMP, "timestamp");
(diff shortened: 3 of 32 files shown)


config-spring-property/src/main/java/org/stellar/anchor/server/config/PropertyResourceAppConfig.java
```diff
@@ -0,0 +1,17 @@
+package org.stellar.anchor.server.config;
+
+import java.util.List;
+import lombok.Data;
+import org.stellar.anchor.config.AppConfig;
+
+@Data
+public class PropertyResourceAppConfig implements AppConfig {
```
JakeUrban, line 8: ❓ Can you briefly explain the relationship between these config classes, the properties files, and the application? How do they interact?

data-spring-jdbc/src/main/java/org/stellar/anchor/server/sqlite/SQLiteDialect.java
```diff
@@ -0,0 +1,132 @@
+package org.stellar.anchor.server.sqlite;
+
+import java.sql.Types;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.function.SQLFunctionTemplate;
+import org.hibernate.dialect.function.StandardSQLFunction;
+import org.hibernate.dialect.function.VarArgsSQLFunction;
+import org.hibernate.type.StringType;
+
+public class SQLiteDialect extends Dialect {
```
JakeUrban, line 10: ❓ I'm surprised this isn't a default hibernate dialect, why do we have to create this on our own?

JakeUrban (commented):
This looks great! Thanks for getting it set up. I left a couple questions.

However, I think we need to remove all SEP-24 related functionality from the platform application. It will not support SEP-24 for the foreseeable future. Happy to discuss this further.

config-spring-property/src/main/java/org/stellar/anchor/server/config/PropertyResourceAppConfig.java
lijamie98: I will be happy to explain on Monday. The document is a little intimidating to read.

data-spring-jdbc/src/main/java/org/stellar/anchor/server/sqlite/SQLiteDialect.java
lijamie98: That’s sad that we have to write the dialect because hibernate does not have the dialect for SQLite.

lijamie98:
> This looks great! Thanks for getting it set up. I left a couple questions.
> 
> However, I think we need to remove all SEP-24 related functionality from the platform application. It will not support SEP-24 for the foreseeable future. Happy to discuss this further.

It’s ok to remove Sep 24 endpoints in following PRs. They are easy to add back when needed. 

It is also possible to dynamically enable/disable sep endpoints based on configurations

JakeUrban:
> It’s ok to remove Sep 24 endpoints in following PRs. They are easy to add back when needed.

@lijamie98 I agree, but I don't understand why we would add SEP-24 logic if we've already decided that we're going to remove it.

> It is also possible to dynamically enable/disable sep endpoints based on configurations

Yes, but we don't want SEP-24 to be an option at this point. We've already decided it will only be SEP-31 during MVP.

lijamie98:
> > It’s ok to remove Sep 24 endpoints in following PRs. They are easy to add back when needed.
> 
> @lijamie98 I agree, but I don't understand why we would add SEP-24 logic if we've already decided that we're going to remove it.
> 
> > It is also possible to dynamically enable/disable sep endpoints based on configurations
> 
> Yes, but we don't want SEP-24 to be an option at this point. We've already decided it will only be SEP-31 during MVP.

JakeUrban (approved):
LGTM. 

One thing I think we should keep in mind is that this now a public project and we should make sure that `main` branch is always runnable and releasable. Since we're in the early phases of the project we can keep the SEP-24 code as a reference until we implement SEP-31 but generally I think we should avoid this pattern.

lijamie98:
> LGTM.
> 
> One thing I think we should keep in mind is that this now a public project and we should make sure that `main` branch is always runnable and releasable. Since we're in the early phases of the project we can keep the SEP-24 code as a reference until we implement SEP-31 but generally I think we should avoid this pattern.

That's good point. Having SEP-24 endpoints there can be confusing for people who are following this repository. I will have following PR to move the SEP-24 code to example subprojects.

README.md
```diff
@@ -7,8 +7,11 @@
 
 The Stellar Anchor SDK is a collection of projects that make easy to build financial applications on Stellar.
 
-* [core](./core) is the base package for implementing standardized anchor applications.
-* [payment-circle](./payment-circle) is a package that wraps a subset of the [Circle API](https://developers.circle.com/reference).
+* [core](./core): the base package for implementing standardized anchor applications.
+* [config-spring-property](./config-spring-property): Implementation of configuration management with [Spring ConfigurationProperties](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config).
+* [data-spring-jdbc](./data-spring-jdbc): Implementation of data access with [Spring Data JDBC](https://spring.io/projects/spring-data-jdbc).
```
marcelosalloum, line 12: 🤔  Is this package meant to use SQLite only? If so we should add `sqlite` to the package name.

platform/src/main/java/org/stellar/anchor/server/SepConfigurationMVC.java
```diff
@@ -0,0 +1,114 @@
...
+
+  @Bean
+  JdbcSep24TransactionStore sep24TransactionStore(JdbcSep24TransactionRepo txnRepo) {
+    return new JdbcSep24TransactionStore(txnRepo);
+  }
+
+  @Bean
+  Sep1Service sep1Service(Sep1Config sep1Config) {
+    return new Sep1Service(sep1Config);
+  }
```
marcelosalloum, line 73: Just so I fully understand what a Bean is: when you define a `Sep1Service sep1Service` in this file, spring will use a new instance of this object to inject the dependencies in any class constructor that contains a `Sep1Service sep1Service` in the signature. 

Is that ☝️  correct?

platform/src/main/resources/application-wks.properties
```diff
@@ -0,0 +1,36 @@
...
+app.stellarNetworkPassPhrase=Test SDF Network ; September 2015
+app.hostUrl=http://localhost:8080
+app.languages=en
+app.horizonURI=https://horizon-testnet.stellar.org
+app.assets=assets-test.json
+app.jwtSecretKey=secret
+
+# sep-1
+sep1.enabled=true
+sep1.stellarFile=sep1/stellar-wks.toml
```
marcelosalloum, line 19: It would be great if we could dynamically build a toml file based on the SEPs and app configurations. Maybe something for later.

marcelosalloum (commented):
Great PR! I can see the platform taking form now 🎉 

I've left a few clarification questions below

platform/src/main/java/org/stellar/anchor/server/SepConfigurationMVC.java
lijamie98: That is correct. 👍

platform/src/main/resources/application-wks.properties
lijamie98: Yes, that's more to come. Sep-1 needs to be re-refactored to work like sep-10 and sep-24.

platform/src/main/java/org/stellar/anchor/server/SepConfigurationMVC.java
marcelosalloum: Cc @JakeUrban, in case he's new to this too.

JakeUrban: Thanks for the tag. I'll have to think about this in a bit and will come with questions offline given this PR is merged.

PR #30, opened 2022-02-04
paymentservice: Cleanup PaymentService Interface

JakeUrban:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Adjusts the `PaymentService` interface based on a discussion offline with @marcelosalloum.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
+++ b/core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
@@ -45,7 +45,8 @@ public interface PaymentService {
   /**
    * API request that creates an account with the given id.
    *
-   * @param accountId is the identifier of the account to be created.
+   * @param accountId is the identifier of the account to be created. It could be mandatory or
+   *     optional depending on the implementation.
    * @return asynchronous stream with the account object.
    * @throws HttpException If the http response status code is 4xx or 5xx.
    */
--- a/core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
+++ b/core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
@@ -56,10 +56,12 @@ public interface PaymentService {
    * API request that returns the history of payments involving a given account.
    *
    * @param accountID the id of the account whose payment history we want to fetch.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
```diff
@@ -9,20 +9,13 @@
  * Contains the interface methods for a payment service. It can be implemented in different
  * networks, like Stellar, Circle, Wyre or others.
  */
-@SuppressWarnings("unused")
```
lijamie98, line 12: Nice! This should be avoided when possible although sometimes, it is not possible if we are building a library)

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
```diff
@@ -27,27 +27,33 @@
 import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 
-@Data
 public class CirclePaymentService implements PaymentService {
   private static final Gson gson = new Gson();
+
+  @Getter()
```
lijamie98, line 33: Lombok is mostly used for object field access to avoid tediously boilerplate code. As a service, it is actually less confusing to use `getNetwork`, `getUrl`, and 'getSecretKey`.

JakeUrban: Oh ok, I can change this.

core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java
JakeUrban: One nice side-effect of using a mono repo is that we can remove when they are used by other subprojects 👍  resolving.

lijamie98:
> add getters, remove unneeded imports 

The `google-java-format` plugin can remove unneeded imports. :-) 

https://github.com/stellar/java-stellar-anchor-sdk/blob/main/build.gradle.kts#L39

lijamie98 (approved):

marcelosalloum (commented):
LGTM 👍

PR #28, opened 2022-02-04
paymentservice: StellarPaymentService

JakeUrban:
<details>
  <summary>PR Checklist</summary>

### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `paymentservice.stellar`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
</details>

### What

[TODO: Short statement about what is changing.]

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #40, opened 2022-02-10
payment-circle: implement getAccountHistory method

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Implement `CirclePaymentService#getAccountHistory` as part of the `PaymentService` implementation. This implementation is a bit non conventional because it aggregates results from multiple endpoints into one, since Circle history is distributed across multiple endpoints:
* `GET /v1/transfers` returns CircleWallet->CircleWallet, CircleWallet->Stellar and Stellar->CircleWallet results.
* `GET /v1/payouts` returns CircleWallet->BankWire results.
* (not implemented) `GET /v1/payments` returns BankWire->CircleWallet and CreditCard->CircleWallet

### Further discussion

When we aggregate results from multiple endpoints, we mess with the consistency of page sizes: if one of the endpoints has more pages than the other, the page sizes will look like this:
* Page 1: 100 results
* Page 2: 53 results
* Page 3...(n-1): 50 results
* Page n: x<50 results

Instead of aggregating the results from multiple endpoints, we could alternatively change the signature of [`getAccountPaymentHistory`](https://github.com/stellar/java-stellar-anchor-sdk/blob/a72b81518d3ca290734f2b76f5e7eece73aed8a1/core/src/main/java/org/stellar/anchor/paymentservice/PaymentService.java#L68-L70) to contemplate which network we want to read status from, something like:

```java
Mono<PaymentHistory> getAccountPaymentHistory(
      String accountID, 
      Network networkOfInterest,
      @Nullable String beforeCursor,
      @Nullable String afterCursor
) throws HttpException;
```

This would increase the complexity on the Anchor side though.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/paymentservice/Payment.java
+++ b/core/src/main/java/org/stellar/anchor/paymentservice/Payment.java
@@ -7,6 +7,7 @@ import lombok.Data;
 @Data
 public class Payment {
   String id;
+  String txHash;
   Account sourceAccount;
   Account destinationAccount;
   /** The balance currency name contains the scheme of the destination network of the payment. */
--- a/core/src/main/java/org/stellar/anchor/paymentservice/Payment.java
+++ b/core/src/main/java/org/stellar/anchor/paymentservice/Payment.java
@@ -7,7 +7,7 @@ import lombok.Data;
 @Data
 public class Payment {
   String id;
-  String txHash;
+  String idTag;
   Account sourceAccount;
   Account destinationAccount;
   /** The balance currency name contains the scheme of the destination network of the payment. */
--- a/core/src/main/java/org/stellar/anchor/paymentservice/PaymentHistory.java
+++ b/core/src/main/java/org/stellar/anchor/paymentservice/PaymentHistory.java
@@ -1,11 +1,17 @@
 package org.stellar.anchor.paymentservice;
 
+import java.util.ArrayList;
 import java.util.List;
+import lombok.Data;
 
-@SuppressWarnings("unused")
+@Data
 public class PaymentHistory {
   Account account;
   String afterCursor;
   String beforeCursor;
-  List<Payment> payments;
+  List<Payment> payments = new ArrayList<>();
+
+  public PaymentHistory(Account account) {
+    this.account = account;
+  }
 }
--- a/payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
+++ b/payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
@@ -280,7 +280,7 @@ public class CirclePaymentService implements PaymentService, StellarReconciliati
             });
   }
 
-  public Mono<CircleTransferListResponse> _getTransfers(
+  public Mono<CircleTransferListResponse> getTransfers(
       String accountID, String beforeCursor, String afterCursor, Integer pageSize)
       throws HttpException {
     // build query parameters for GET requests
@@ -314,6 +314,7 @@ public class CirclePaymentService implements PaymentService, StellarReconciliati
                   gson.fromJson(body, CircleTransferListResponse.class);
 
               for (CircleTransfer transfer : response.getData()) {
+                // TODO: parallelize this call
                 updateStellarSenderAddress(transfer);
               }
 
@@ -321,13 +322,13 @@ public class CirclePaymentService implements PaymentService, StellarReconciliati
             });
   }
 
-  public Mono<CirclePayoutListResponse> _getPayouts(
+  public Mono<CirclePayoutListResponse> getPayouts(
       String accountID, String beforeCursor, String afterCursor, Integer pageSize)
       throws HttpException {
     // build query parameters for GET requests
-    Integer _pageSize = pageSize != null ? pageSize : 50;
+    int _pageSize = pageSize != null ? pageSize : 50;
     LinkedHashMap<String, String> queryParams = new LinkedHashMap<>();
-    queryParams.put("pageSize", _pageSize.toString());
+    queryParams.put("pageSize", Integer.toString(_pageSize));
     queryParams.put("source", accountID);
 
     if (afterCursor != null && !afterCursor.isEmpty()) {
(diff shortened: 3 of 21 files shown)


marcelosalloum:
I used the interface [`StellarReconciliation`] with a default implementation as a strategy to avoid increasing the size of CirclePaymentService indefinitely, so new features can be included as add-ons or class extensions. I'm new to this in Java so I'm not 100% sure this is a good practice, WDYT @lijamie98?

It was implemented in [`CirclePaymentService`] BTW.

[`StellarReconciliation`]: https://github.com/stellar/java-stellar-anchor-sdk/blob/b7a0f0e2c3e3fdbc38e47de97b9236554d886569/payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/StellarReconciliation.java

[`CirclePaymentService`]: https://github.com/stellar/java-stellar-anchor-sdk/blob/b7a0f0e2c3e3fdbc38e47de97b9236554d886569/payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java#L29

core/src/main/java/org/stellar/anchor/paymentservice/Payment.java
```diff
@@ -7,6 +7,7 @@
 @Data
 public class Payment {
   String id;
+  String txHash;
```
JakeUrban, line 10: Can we generalize this to `idTag` like the `Account` class? Or maybe just `secondaryId`?

For payments that don't involve Stellar it doesn't make sense to have this specific `txHash` attribute in my mind.

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
```diff
@@ -1,30 +1,41 @@
...
 import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 
-public class CirclePaymentService implements PaymentService {
-  private static final Gson gson = new Gson();
+public class CirclePaymentService implements PaymentService, StellarReconciliation {
+  private static final Gson gson =
+      new GsonBuilder()
+          .registerTypeAdapter(CircleTransfer.class, new CircleTransfer.Serialization())
+          .registerTypeAdapter(CirclePayout.class, new CirclePayout.Deserializer())
```
JakeUrban, line 33: Whats the point of registering these type adapters if you still have to pass the class to `fromJson()`?

core/src/main/java/org/stellar/anchor/paymentservice/PaymentHistory.java
```diff
@@ -1,11 +1,17 @@
 package org.stellar.anchor.paymentservice;
 
+import java.util.ArrayList;
 import java.util.List;
+import lombok.Data;
 
-@SuppressWarnings("unused")
+@Data
 public class PaymentHistory {
```
JakeUrban, line 8: There is some overlap between your changes in this PR and [mine](https://github.com/stellar/java-stellar-anchor-sdk/pull/28/files) for the StellarPaymentService.

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
```diff
@@ -266,6 +280,95 @@ private HttpClient getWebClient(boolean authenticated) {
...
+    LinkedHashMap<String, String> queryParams = new LinkedHashMap<>();
+    queryParams.put("pageSize", Integer.toString(_pageSize));
+    queryParams.put("walletId", accountID);
+
+    if (afterCursor != null && !afterCursor.isEmpty()) {
+      queryParams.put("pageAfter", afterCursor);
+      // we can't use both pageBefore and pageAfter at the same time, that's why I'm using 'else if'
+    } else if (beforeCursor != null && !beforeCursor.isEmpty()) {
+      queryParams.put("pageBefore", beforeCursor);
+    }
```
JakeUrban, line 297: In my StellarPaymentService PR I got rid of the `beforeCursor` and only support a single `cursor` parameter. Do you think that could work for the CirclePaymentService?

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/StellarReconciliation.java
```diff
@@ -0,0 +1,79 @@
...
+    // chain == "XLM" and type == BLOCKCHAIN
+    if (!transfer.getSource().getChain().equals("XLM")
+        || transfer.getSource().getType() != CircleTransactionParty.Type.BLOCKCHAIN) {
+      throw new HttpException(500, "invalid source account");
+    }
+
+    Page<OperationResponse> responsePage;
+    try {
+      String txHash = transfer.getTransactionHash();
+      responsePage = getHorizonServer().payments().forTransaction(txHash).execute();
```
JakeUrban, line 35: This is a blocking IO call, shouldn't we try to do this using our web client?

JakeUrban (approved):
LGTM. I have a few questions, and there may be some discussion about the PaymentService interface since my StellarPaymentService PR makes some changes as well.

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
JakeUrban: Well Horizon's endpoints only support a single `cursor`, so the issue is that I would have to ignore the `beforeCursor` specified for calls on `StellarPaymentService` objects.

I don't see the need for a `beforeCursor` because that limit can always be imposed by the client consuming the data. This differs from the `afterCursor` which is required to fetch the next page of data.

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/StellarReconciliation.java
JakeUrban: Oh I see, I wasn't aware of this mechanic. At the base layer though, using the SDK's horizon server will block a thread right? Even if its not the main thread?

marcelosalloum: Yes, that's correct.

payment-circle/src/main/java/org/stellar/anchor/paymentservice/circle/CirclePaymentService.java
marcelosalloum: In Horizon, you actually have the `next` and `prev` pages, which work similarly to the `after` and `before` cursors. Although the prev page is returned in inverse order, which can make things a bit messy to translate from one convention to the other.

I'm ok disabling the before cursor though, in order to make implementation simpler. I'd rather do that in a separate PR, if that's ok.

JakeUrban (approved):
LGTM 👍

PR #73, opened 2022-02-22
SEP-12 MVP implementation

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The MVP implementation of SEP-12

### Known limitations
NA


anchor-reference-server/src/main/java/org/stellar/anchor/reference/controller/CustomerController.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.stellar.anchor.exception.SepNotFoundException;
+import org.stellar.anchor.integration.customer.*;
+import org.stellar.anchor.reference.config.AppSettings;
+import org.stellar.anchor.reference.service.CustomerService;
+
+@RestController
+public class CustomerController {
```
JakeUrban, line 13: I just want confirm my understanding before I deep dive into this PR.

This `CustomerController` class is the server-side implementation of the requests made from the platform right?

In that case, this controller should use the endpoints and request / response schemas defined here: https://github.com/stellar/stellar-anchor-platform/blob/main/design/callbacksApi/api.yml#L268

From what I can see now, we're using the `/customers` path, but the Callback API uses `/customer`. Also, I see we're using the `integration.customer` request & response classes here, but those classes don't match the schemas used in the Callback API. For example, the Callback API includes the `lang` parameter.

lijamie98: > I just want confirm my understanding before I deep dive into this PR.

This PR has rough edges. It is not necessary to do a full review at this moment yet. I will continue to implement end-to-end sep-12 implementation. Then we can do a full review. 

> 
> This `CustomerController` class is the server-side implementation of the requests made from the platform right?
> 
Yes. 

> In that case, this controller should use the endpoints and request / response schemas defined here: https://github.com/stellar/stellar-anchor-platform/blob/main/design/callbacksApi/api.yml#L268
Yes, that's right. Can you push commits to the `develop-12` branch? 

> 
> From what I can see now, we're using the `/customers` path, but the Callback API uses `/customer`. Also, I see we're using the `integration.customer` request & response classes here, but those classes don't match the schemas used in the Callback API. For example, the Callback API includes the `lang` parameter.

This PR is a starting point. We should make it conform to the design. https://github.com/stellar/stellar-anchor-platform/blob/main/design/callbacksApi/api.yml#L268

JakeUrban: Once I create the `platform-apis` subproject, the reference controller can import those classes?

platform/src/main/java/org/stellar/anchor/platform/NettyCustomerIntegration.java
```diff
@@ -1,118 +1,172 @@
...
     if (request.getId() != null) {
-      encoder.addParam("id", request.getId());
+      urlBuilder.addQueryParameter("id", request.getId());
     } else {
-      encoder.addParam("account", request.getAccount());
+      urlBuilder.addQueryParameter("account", request.getAccount());
       if (request.getMemo() != null && request.getMemoType() != null) {
-        encoder.addParam("memo", request.getMemo());
-        encoder.addParam("memo_type", request.getMemoType());
+        urlBuilder.addQueryParameter("memo", request.getMemo());
```
lijamie98, line 55: Builder pattern can be cascaded. 

```
urlBuilder
    .addQueryParameter("memo", request.getMemo())
    .addQueryParameter("memo", request.getMemo());
```

platform/src/main/java/org/stellar/anchor/platform/NettyCustomerIntegration.java
```diff
@@ -1,118 +1,172 @@
...
+  private final OkHttpClient httpClient;
 
   public NettyCustomerIntegration(String baseUri, CustomerRepository customerRepository) {
     try {
       new URI(baseUri);
     } catch (URISyntaxException e) {
       throw new IllegalArgumentException("invalid 'baseUri'");
     }
-    this.baseUri = baseUri;
+    HttpUrl baseUrl = HttpUrl.parse(baseUri);
```
lijamie98, line 34: You are using okhttp here. Since we are using Netty, I suggest that you use the Netty way.

platform/src/main/java/org/stellar/anchor/platform/NettyCustomerIntegration.java
```diff
@@ -1,118 +1,172 @@
...
+  public org.stellar.anchor.dto.sep12.PutCustomerResponse putCustomer(
+      org.stellar.anchor.dto.sep12.PutCustomerRequest request) throws AnchorException {
+    RequestBody requestBody =
+        RequestBody.create(
+            gson.toJson(request, PutCustomerRequest.class), MediaType.get("application/json"));
+    Request callbackRequest = new Request.Builder().url(customerUrl).put(requestBody).build();
+    Response callbackResponse = getResponse(callbackRequest);
+    String callbackResponseContent = getResponseContent(callbackResponse);
+
+    if (callbackResponse.code() == 200) {
```
lijamie98, line 95: Use HttpStatus.OK

JakeUrban (approved):
I think merging this PR is ok since this codebase is still pre-release. But we should fast-follow with fixes for these tests.

PR #86, opened 2022-02-24
payment-circle: fix instructions for the Wire deposit

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix the deposit instructions when receiving bank wire deposits.

### Why

[This section](https://developers.circle.com/docs/wire-payments-quickstart#3-create-the-bank-account-you-will-accept-a-payment-from) of the circle documentation states that you need to effectively reach the `/banks/wires/:id/instructions` endpoint for the deposit instructions. We were returning the user bank info instead of the Circle bank account's deposit instructions.


JakeUrban (approved):
Good catch on the Circle API.

I find their deposit instructions endpoint quite interesting. It reminds me a lot of what we try to do in the success response to a SEP-6 `GET /deposit` request, where we have the `how` string and `extra_info` object. I don't have any particular action to take from this but its good to see an example of a similar use case.

PR #91, opened 2022-02-28
platform: remove all SEP-24 related code

JakeUrban:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Removes all SEP-24 related code from the Platform.

### Why

The platform will not support SEP-24 for the foreseeable future and the project no longer needs the code to be used as a reference for implementing other SEPs.


reecexlm (approved):
LGTM

reecexlm:
one question i have, is if this is working code, can we leave it as a disabled?

lijamie98:
@reecexlm It is possible to leave them as disabled or deleted. 
@JakeUrban I had not have time to create example projects yet. Can we wait a little longer?

reecexlm:
i agree that we should not have unfinished code in our "main" branch. if this is working, complete code (code complete)  then i don't see a reason to remove it. its a feature that is ready and can be toggled on/off right? unless i am missing some context about why we would not want a feature available in our code base that might be used long term.  if its still WIP then IMHO, it should live in a branch until it is ready to be part of our main branch.  i might also be missing some context on our branch strategy.  thanks!

PR #107, opened 2022-03-09
add github action to build/push docker image

stfung77:
update dockerfile to use a build stage resulting in reduced size
add github action to build docker image and push to docker hub


Dockerfile
```diff
@@ -1,22 +1,21 @@
...
+# build jar
+ADD . /code
+WORKDIR /code
+RUN ./gradlew clean build --stacktrace
+
 FROM ubuntu:20.04
 
 RUN apt-get update && \
 apt-get install -y --no-install-recommends \
         openjdk-11-jdk
```
leighmcculloch, line 11: For the final image you should only need the JRE and not the full JDK.
```suggestion
        openjdk-11-jre
```

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,32 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ main ]
```
leighmcculloch, line 5: It looks like this workflow runs independently of any tests that run. I think you should add itas a job to the build and test workflow, and you can configure it to run after the tests run so that we only push images that pass tests.

leighmcculloch (commented):
This looks great. 🎉 Thanks for pinning the third party actions, and love the use of multistage builds in the image. A couple suggestions inline.

.github/workflows/docker-image.yml
stfung77: makes sense. i'll do that

.github/workflows/build_and_test.yml
```diff
@@ -26,3 +24,27 @@ jobs:
       run: ./gradlew jar
     - name: Test
       run: ./gradlew test
+      
+  build-and-push-anchor-platform:
+    runs-on: ubuntu-latest
```
leighmcculloch, line 29: ❗ Jobs will run in parallel unless you explicitly link their dependencies. You'll need to add a `needs` to ensure this runs after the `build_and_test`.
```suggestion
  build-and-push-anchor-platform:
    needs: [build_and_test]
    runs-on: ubuntu-latest
```

leighmcculloch (commented):
One thing that needs adding (❗), and one suggestion (💡).

.github/workflows/build_and_test.yml
```diff
@@ -26,3 +24,27 @@ jobs:
...
+    - name: Get image tag (short sha)
+      shell: bash
+      id: get_tag
+      run: echo ::set-output name=TAG::$(git rev-parse --short ${{ github.sha }} )
+
+    - name: Login to DockerHub
+      uses: docker/login-action@bb984efc561711aaa26e433c32c3521176eae55b
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
```
lijamie98, line 43: Where are these secrets stored?

leighmcculloch: In GitHub, they're something Ops will configure.

.github/workflows/build_and_test.yml
```diff
@@ -26,3 +24,27 @@ jobs:
...
+  build-and-push-anchor-platform:
+    runs-on: ubuntu-latest
+    name: Push stellar/anchor-platform:sha to DockerHub
+    steps:
+    - uses: actions/checkout@v3
+    
+    - name: Get image tag (short sha)
+      shell: bash
+      id: get_tag
+      run: echo ::set-output name=TAG::$(git rev-parse --short ${{ github.sha }} )
```
leighmcculloch, line 37: One thing to be aware of is that this image on PRs that gets pushed will have a sha you don't necessarily expect. That's because Github Actions runs workflows on a merge commit with the destination branch merged into the PR branch. This won't affect builds on the main branch, or on the push builds that happen on branches either. If this is going to be a problem let me know and we can talk more about options.

leighmcculloch (approved):

PR #121, opened 2022-03-15
git action, 'cat' results of test

stfung77:
cat the test results file
change distribution for test_and_build to corretto


.github/workflows/build_and_test.yml
```diff
@@ -19,11 +19,10 @@ jobs:
         java-version: '11'
         distribution: 'adopt'
     - name: Clean
```
lijamie98, line 21: The name should be 'Clean, Build and Test`

lijamie98 (approved):

PR #113, opened 2022-03-11
core: implement SEP-38 `GET /prices` endpoint

marcelosalloum:
<details>
  <summary>PR Checklist</summary>

### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `paymentservice.stellar`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
</details>

### What

Implement SEP-38 [`GET /prices`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-prices) endpoint.

### Why

Closes #77.

PR #125, opened 2022-03-16
core: SEP-38's number of decimals can now be set in the assets.json configuration file

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The number of decimals returned in SEP-38's [`GET /prices`], [`POST /quote`] and [`GET /quote/:id`] requests now defaults to **7** and can be configured to a different value in the assets.json config file.

### Why

Addressing Jake's comment in https://github.com/stellar/java-stellar-anchor-sdk/pull/113#discussion_r826102562.

[`GET /prices`]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-prices
[`POST /quote`]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#post-quote
[`GET /quote/:id`]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-quote

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -198,21 +198,22 @@ public class Sep38Service {
     BigDecimal bSellAmount, bBuyAmount;
     if (sellAmount != null) {
       bSellAmount = new BigDecimal(sellAmount);
-      bBuyAmount = bSellAmount.divide(bPrice, buyAsset.getDecimals(), RoundingMode.HALF_UP);
+      bBuyAmount = bSellAmount.divide(bPrice, buyAsset.getDecimals(), RoundingMode.DOWN);
     } else {
       bBuyAmount = new BigDecimal(buyAmount);
       bSellAmount = bBuyAmount.multiply(bPrice);
     }
     builder =
         builder
-            .sellAmount(formatAmount(bSellAmount, sellAsset.getDecimals()))
-            .buyAmount(formatAmount(bBuyAmount, buyAsset.getDecimals()));
+            .sellAmount(formatAmount(bSellAmount, sellAsset.getDecimals(), RoundingMode.UP))
+            .buyAmount(formatAmount(bBuyAmount, buyAsset.getDecimals(), RoundingMode.DOWN));
 
     return builder.build();
   }
 
-  private String formatAmount(BigDecimal amount, Integer decimals) throws NumberFormatException {
-    BigDecimal newAmount = amount.setScale(decimals, RoundingMode.HALF_UP);
+  private String formatAmount(BigDecimal amount, Integer decimals, RoundingMode roundingMode)
+      throws NumberFormatException {
+    BigDecimal newAmount = amount.setScale(decimals, roundingMode);
 
     DecimalFormat df = new DecimalFormat();
     df.setMaximumFractionDigits(decimals);
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -2,6 +2,7 @@ package org.stellar.anchor.sep38;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.text.DecimalFormat;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -194,16 +195,30 @@ public GetPriceResponse getPrice(
     GetPriceResponse.GetPriceResponseBuilder builder =
         GetPriceResponse.builder().price(rateResponse.getRate().getPrice());
     BigDecimal bPrice = new BigDecimal(rateResponse.getRate().getPrice());
+    BigDecimal bSellAmount, bBuyAmount;
     if (sellAmount != null) {
-      BigDecimal bSellAmount = new BigDecimal(sellAmount);
-      BigDecimal bBuyAmount = bSellAmount.divide(bPrice, 4, RoundingMode.HALF_UP);
-      builder = builder.sellAmount(sellAmount).buyAmount(bBuyAmount.toPlainString());
+      bSellAmount = new BigDecimal(sellAmount);
+      bBuyAmount = bSellAmount.divide(bPrice, buyAsset.getDecimals(), RoundingMode.HALF_UP);
```
JakeUrban, line 201: ❓ Shouldn't we round down? If the anchor has specified 2 decimals of precision, and we're dealing with an amount 1.234, shouldn't the service simply not recognize the 3rd decimal and act as if 1.23 was sent?

There are a few other places `HALF_UP` is used.

marcelosalloum: Yeah, great point!

Added https://github.com/stellar/java-stellar-anchor-sdk/pull/125/commits/eb69c3657b34be78a0e8f201cbb5deab0504466d to update the rounding logic:
* sell_amount is rounded UP
* buy_amount is rounded DOWN

let me know if that makes sense to you too.

JakeUrban: Why is `sell_amount` rounded up?

platform/src/main/resources/assets-prod.json
```diff
@@ -62,6 +62,7 @@
                     "stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN"
                 ],
                 "country_codes": ["USA"],
+                "decimals": 4,
```
lijamie98, line 65: Reading from assets-prod is legacy code from MGI. We need to externalize the assets information (not in the code/resource).
Issue [#128](https://github.com/stellar/java-stellar-anchor-sdk/issues/128) is created.

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
marcelosalloum: If we consider 2 parties, CLIENT and ANCHOR, it's expected that the volume of ANCHOR transactions is way higher than CLIENT's. With that in mind, I propose the rounding difference be shifted in favor of the ANCHOR to prevent the risk where a high amount of transactions could cause a significant loss in the long term.

Let's take the following scenarios (considering 2 decimals only)

Rounding sell amount UP will prevent the ANCHOR from receiving less than expected:
* `sell_amount=1.354` --rounded to--> `sell_amount=1.36` (UP)
* `buy_amount=2.400` --rounded to-->`buy_amount=2.40`

Rounding buy amount DOWN will prevent the ANCHOR from paying more than expected:
* `sell_amount=3.620` --rounded to--> `sell_amount=3.62`
* `buy_amount=6.407` --rounded to-->`buy_amount=6.40` (DOWN)

WDYT?

JakeUrban: I'm a bit confused with the examples, but I think I understand what you're saying.

If a client requests a `sell_amount` of `1.2345` (say 4 decimals are allowed for this off-chain asset), and the `price` (exchange rate) is `2.3456`, then `buy_amount` is `0.52630457...`.

Lets say the buy asset uses 5 decimals (even though a Stellar asset would use 7). If we round UP, `buy_amount` is `0.52631.`. If we round DOWN, `buy_amount` is `0.52630`.

In this case, it would be better for the anchor if we rounded DOWN like you suggest. In the same way, when calculating `sell_amount` it would be better for the anchor if we rounded UP.

I think this makes sense. Even though the platform is being opinionated by choosing a rounding strategy, the anchor ultimately has the ability to adjust their `price` if they would rather give the client a favorable rate.

JakeUrban (approved):

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
marcelosalloum: I updated all rounding to HALF_DOWN to follow common practices.

PR #148, opened 2022-03-23
code, platform: implement the logic to retrieve a quote by ID, for `GET {SEP-38}/quote/:id`

### LINKED ISSUE #80: SEP-38: implement the logic to retrieve a quote by ID, for `GET /quote/:id`

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Implement the logic to retrieve a quote by ID, for `GET {SEP-38}/quote/:id`

### Why

Closes #80.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -404,7 +404,7 @@ public class Sep38Service {
     if (!StringUtils.equals(quote.getCreatorAccountId(), account)
         || !StringUtils.equals(memo, quote.getCreatorMemo())
         || !StringUtils.equals(memoType, quote.getCreatorMemoType())) {
-      throw new ServerErrorException("internal server error");
+      throw new NotFoundException("quote not found");
     }
 
     return Sep38QuoteResponse.builder()
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -394,6 +394,11 @@ public class Sep38Service {
       throw new BadRequestException("sep10 token is malformed");
     }
 
+    // empty quote id
+    if (StringUtils.isEmpty(quoteId)) {
+      throw new BadRequestException("quote id cannot be empty");
+    }
+
     Sep38Quote quote = this.sep38QuoteStore.findByQuoteId(quoteId);
     if (!StringUtils.equals(quote.getCreatorAccountId(), account)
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -370,4 +371,50 @@ public Sep38QuoteResponse postQuote(JwtToken token, Sep38PostQuoteRequest reques
...
+      throw new BadRequestException("quote id cannot be empty");
+    }
+
+    // validate consistency between quote and jwt token
+    Sep38Quote quote = this.sep38QuoteStore.findByQuoteId(quoteId);
+    if (!StringUtils.equals(quote.getCreatorAccountId(), account)
+        || !StringUtils.equals(memo, quote.getCreatorMemo())
+        || !StringUtils.equals(memoType, quote.getCreatorMemoType())) {
+      throw new ServerErrorException("internal server error");
+    }
```
JakeUrban, line 408: I believe we should be returning a 404 here, since the client authenticated is the not the client that created the quote.

JakeUrban (changes requested):
I think there is one adjustment needed but otherwise LGTM. Once the fix is pushed I'll approve 👍

lijamie98 (approved):
LGTM other than confirming Jake's comment.

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
marcelosalloum: That's a great point! I'm not so sure what the best approach should be here...
* `404` makes sense since this is like "this quote id could not be found in your quotes"
* `403` could also make sense since the user is not authorized to retrieve that quote, but we would be giving too much info for someone trying to exploit this endpoint
* `500` would mean "something unexpected happened" but you're right, I don't think this is the case here.

marcelosalloum: I'll go ahead and change it to 404.

JakeUrban (approved):

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
lijamie98: I also think `404` is better.

marcelosalloum: Updated to `404` in https://github.com/stellar/java-stellar-anchor-sdk/pull/148/commits/39681c059dd6cebbf1460dd95378c8e25cbe061a.

PR #153, opened 2022-03-24
core: Issue #152 Check if sep38 is enabled in the `/info` endpoing

### LINKED ISSUE #152: Make sure only sep48_enabled assets show up in the `GET {SEP-38}/info` response
### What

In this loop:
https://github.com/stellar/java-stellar-anchor-sdk/blob/1ae01c929c533b3f875f3963b1c4aa7be38171e6/core/src/main/java/org/stellar/anchor/dto/sep38/InfoResponse.java#L14-L17

Make sure we look at the value of sep38_enabled:
https://github.com/stellar/java-stellar-anchor-sdk/blob/816681201d0bbbfcab2fd6fade5370f43efec799/platform/src/main/resources/assets-test.json#L34

and only include the assets where `sep38_enabled==true`

lijamie98:
#closes #152

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Check if sep38 is enabled in the `/info` endpoing

### Why

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/dto/sep38/InfoResponse.java
+++ b/core/src/main/java/org/stellar/anchor/dto/sep38/InfoResponse.java
@@ -12,29 +12,29 @@ public class InfoResponse {
   private List<Asset> assets = new ArrayList<>();
 
   public InfoResponse(List<AssetInfo> assetInfoList) {
-    assetInfoList.forEach(
-        assetInfo -> {
-          Asset newAsset = new Asset();
-          String assetName = assetInfo.getSchema().toString() + ":" + assetInfo.getCode();
-          if (!Objects.toString(assetInfo.getIssuer(), "").isEmpty()) {
-            assetName += ":" + assetInfo.getIssuer();
-          }
-          newAsset.setAsset(assetName);
+    for (AssetInfo assetInfo : assetInfoList) {
+      if (!assetInfo.getSep38Enabled()) continue;
+      Asset newAsset = new Asset();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/stellar/anchor/dto/sep38/InfoResponse.java
```diff
@@ -14,26 +14,28 @@
   public InfoResponse(List<AssetInfo> assetInfoList) {
     assetInfoList.forEach(
         assetInfo -> {
-          Asset newAsset = new Asset();
-          String assetName = assetInfo.getSchema().toString() + ":" + assetInfo.getCode();
-          if (!Objects.toString(assetInfo.getIssuer(), "").isEmpty()) {
-            assetName += ":" + assetInfo.getIssuer();
+          if (assetInfo.getSep38Enabled()) {
```
marcelosalloum, line 17: nitpick: I'd prefer an approach without an extra nesting level, and return early instead. Something like:

```java
for (AssetInfo assetInfo : assetInfoList) {
  if (!assetInfo.getSep38Enabled())
    continue;

  // TODO: continue doing what we were doing.
```

marcelosalloum (approved):
LGTM! I added a small suggestion below.

core/src/main/java/org/stellar/anchor/dto/sep38/InfoResponse.java
lijamie98: I also like this better!

PR #158, opened 2022-03-29
all: replace usage of `java.util.Date` and `java.time.LocalDateTime` with `java.time.Instant`

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Replace usage of java.util.Date and java.time.LocalDateTime with `java.time.Instant`. Also, updated the gson serializer to make sure `java.time.Instant` is serialized/deserialized using a readable format in the UTC timezone, like `"2022-03-01T01:01:01.544Z"` for instance.

### Why

We were using different objects to represent dates, which was pretty inconsistent.

Additionally, Gson does not serialize java.time.* objects to a string (by default) so we had to use a custom Gson instance to make sure we would serialize things accordingly.


core/src/main/java/org/stellar/anchor/util/GsonUtils.java
```diff
@@ -0,0 +1,18 @@
...
+import java.time.Instant;
+
+public class GsonUtils {
+  private static Gson instance = null;
+
+  public static GsonBuilder builder() {
+    return new GsonBuilder().registerTypeAdapter(Instant.class, new InstantConverter());
+  }
+
+  public static Gson getGsonInstance() {
```
lijamie98, line 14: In Java Singleton pattern, this is usually called `getInstance()`
https://www.baeldung.com/java-singleton

anchor-reference-server/src/main/java/org/stellar/anchor/reference/controller/RateController.java
```diff
@@ -12,9 +12,11 @@
 @RestController
 public class RateController {
   private final RateService rateService;
+  private final Gson gson;
 
-  public RateController(AppSettings appSettings, RateService rateService) {
+  public RateController(AppSettings appSettings, RateService rateService, Gson gson) {
```
lijamie98, line 17: I am not sure if it is too much to autowired it? 😄  

Should we use GsonUtils.builder().create()?

lijamie98 (approved):
LGTM. 

I need this PR fo fix my `java.time.Instant` serialization problem. 😄

core/src/main/java/org/stellar/anchor/util/GsonUtils.java
marcelosalloum: https://github.com/stellar/java-stellar-anchor-sdk/pull/158/commits/130707dffdc2cc60462b7b10a8f0bc33f73fb739

anchor-reference-server/src/main/java/org/stellar/anchor/reference/controller/RateController.java
marcelosalloum: https://github.com/stellar/java-stellar-anchor-sdk/pull/158/commits/923fdb5b54648c99affd3eda00fd458c792a217f

PR #179, opened 2022-04-08
platform: Payment observer implementation

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Implement payment observer. 

### Why

N/A.

### Known limitations

1. Events will be triggered when the event API is ready.
2. There are protected fields in Java Stellar SDK. We need to submit PRs to make those fields accessible (by getters) to validate and populate event fields. 
3. TransactionEvent class is mocked before the Event API PR is merged.

### DIFF
--- a/data-spring-jdbc/src/main/java/org/stellar/anchor/server/data/StellarAccountPageToken.java
+++ b/data-spring-jdbc/src/main/java/org/stellar/anchor/server/data/StellarAccountPageToken.java
@@ -1,0 +1,17 @@
+package org.stellar.anchor.server.data;
+
+import com.google.gson.annotations.SerializedName;
+import javax.persistence.*;
+import lombok.Data;
+
+@Data
+@Entity
+@Access(AccessType.FIELD)
+@Table(name = "stellar_account_page_token")
+public class StellarAccountPageToken {
+  @Id
+  @SerializedName("account_id")
+  String accountId;
+
+  String token;
+}
--- a/platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
@@ -1,5 +1,10 @@
 package org.stellar.anchor.platform.paymentobserver;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import org.stellar.anchor.util.Log;
 import org.stellar.sdk.KeyPair;
 import org.stellar.sdk.Server;
 import org.stellar.sdk.requests.EventListener;
@@ -9,11 +14,6 @@ import org.stellar.sdk.responses.operations.OperationResponse;
 import org.stellar.sdk.responses.operations.PaymentOperationResponse;
 import shadow.com.google.common.base.Optional;
 import shadow.com.google.gson.Gson;
-
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
 
 public class StellarPaymentObserver {
   final Server server;
@@ -63,17 +63,24 @@ public class StellarPaymentObserver {
           public void onEvent(OperationResponse transaction) {
             if (transaction instanceof PaymentOperationResponse) {
               PaymentOperationResponse payment = (PaymentOperationResponse) transaction;
-              if (payment.getTo().equals(account)) {
-                observers.forEach(observer -> observer.onReceived(payment));
-              } else if (payment.getFrom().equals(account)) {
-                observers.forEach(observer -> observer.onSent(payment));
+              try {
+                if (payment.getTo().equals(account)) {
+                  observers.forEach(observer -> observer.onReceived(payment));
+                } else if (payment.getFrom().equals(account)) {
+                  observers.forEach(observer -> observer.onSent(payment));
+                }
+              } catch (Throwable t) {
+                Log.errorEx(t);
               }
             }
             pageTokenStore.save(account, transaction.getPagingToken());
           }
 
           @Override
-          public void onFailure(Optional<Throwable> exception, Optional<Integer> statusCode) {}
+          public void onFailure(Optional<Throwable> exception, Optional<Integer> statusCode) {
+            // TODO: The stream seems closed when failure happens. Improve the reliability of the
+            // stream.
+          }
         });
   }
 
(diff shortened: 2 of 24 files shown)


stfung77 (approved):

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /** Start watching the accounts. */
+  public void start() {
+    for (String account : accounts) {
+      SSEStream<OperationResponse> stream = watch(account);
+      this.streams.add(stream);
+    }
+  }
+
```
marcelosalloum, line 44: ❗   One streamer per account would create many streamers in parallel and that could reach the rate limit quicker.

The approach we did for Vibrant was to stream the `{horizon}/payment` endpoint and filter the results based on the accounts we were looking for. Then we made the DB changes based on those filtered results.

Also, in Vibrant we managed to reach a rate limit with a single streamer when we used an old cursor with a limit smaller than `50` (not so sure about the exact number, TBH). That means an anchor with multiple assets  (and distribution accounts) could meet the rate limit soon.

💡  We can create an issue just for this change and handle it in separate as a tech debt.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -0,0 +1,123 @@
...
+  /** Graceful shutdown. */
+  public void shutdown() {
+    for (SSEStream<OperationResponse> stream : streams) {
+      stream.close();
+    }
+  }
+
+  public SSEStream<OperationResponse> watch(String account) {
+    PaymentsRequestBuilder paymentsRequest = server.payments().forAccount(account);
```
marcelosalloum, line 53: 💡  We should configure the streamer with `limit=200` to make the requests more efficient, and `order=desc` to make sure we're streaming from oldest to newest. I think the default order is `desc`, but I suggest making it explicit here _just in case_.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -0,0 +1,123 @@
...
+    if (lastToken != null) {
+      paymentsRequest.cursor(lastToken);
+    }
+
+    return paymentsRequest.stream(
+        new EventListener<>() {
+          @Override
+          public void onEvent(OperationResponse transaction) {
+            if (transaction instanceof PaymentOperationResponse) {
+              PaymentOperationResponse payment = (PaymentOperationResponse) transaction;
```
marcelosalloum, line 65: ❓  Does this include path payments?

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarTransaction.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.google.gson.annotations.SerializedName;
+import java.time.Instant;
+import lombok.Builder;
+import lombok.Data;
+import org.stellar.anchor.paymentservice.Payment;
+
+@Data
+@Builder
+public class StellarTransaction {
```
marcelosalloum, line 11: ❓  Can't we use the Transaction from the Stellar SDK?

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -0,0 +1,95 @@
...
+  PaymentOperationToEventListener(JdbcSep31TransactionStore transactionStore) {
+    this.transactionStore = transactionStore;
+  }
+
+  @Override
+  public void onReceived(PaymentOperationResponse payment) {
+    // Find the matching transaction
+    Sep31Transaction txn = transactionStore.findByStellarAccountId(payment.getTo());
+    if (txn == null) {
+      Log.error(String.format("no transaction(stellarAccountId=%s) is found.", payment.getTo()));
```
marcelosalloum, line 30: nit: maybe this should be a warning.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -0,0 +1,95 @@
...
+      // Asset does not match
+      return;
+    }
+
+    AssetTypeCreditAlphaNum atcPayment = (AssetTypeCreditAlphaNum) payment.getAsset();
+    if (!txn.getAmountInAsset().equals(atcPayment.getCode())) {
+      Log.error(
+          String.format(
+              "Payment asset(%s) does not match the expected asset(%s)",
+              atcPayment.getCode(), txn.getAmountInAsset()));
```
marcelosalloum, line 44: nit: Maybe Log.warn here too? This is not exactly an error.

marcelosalloum (commented):
Added a couple of important fixes as ❗, as well as a few suggestions 💡 , questions ❓, and nitpicks.

data-spring-jdbc/src/main/java/org/stellar/anchor/server/data/StellarAccountPageToken.java
```diff
@@ -0,0 +1,17 @@
...
+@Entity
+@Access(AccessType.FIELD)
+@Table(name = "stellar_account_page_token")
+public class StellarAccountPageToken {
+  @Id
+  @SerializedName("account_id")
+  String accountId;
+
+  String token;
+}
```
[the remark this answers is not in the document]
marcelosalloum: Or PaymentStreamerCursor

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarTransaction.java
marcelosalloum: 🤔  Maybe we should send over the PaymentOperation instead, since a Transaction can contain multiple payments/operations.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
marcelosalloum: I created an issue for this in https://github.com/stellar/java-stellar-anchor-sdk/issues/182.

PR #184, opened 2022-04-12
platform: Integrate payment observer with SEP-31

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Integrate PaymentObserver with SEP-31. 

### Why

N/A.
### Known limitations

The transaction status is changed to COMPLETED after the payment is received. When event queue is implemented, the reference server should call `PATCH /transactions` to update the status from PENDING_RECEIVER to COMPLETED.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
@@ -3,16 +3,16 @@ package org.stellar.anchor.sep31;
 import static org.stellar.anchor.dto.sep31.Sep31InfoResponse.AssetResponse;
 import static org.stellar.anchor.sep31.Sep31Helper.amountEquals;
 import static org.stellar.anchor.util.MathHelper.decimal;
+import static org.stellar.anchor.util.MemoHelper.*;
 import static org.stellar.anchor.util.SepHelper.*;
+import static org.stellar.sdk.xdr.MemoType.*;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.stream.Collectors;
 import lombok.SneakyThrows;
+import org.apache.commons.lang3.StringUtils;
 import org.stellar.anchor.asset.AssetInfo;
 import org.stellar.anchor.asset.AssetInfo.Sep31TxnFieldSpecs;
 import org.stellar.anchor.asset.AssetService;
@@ -76,8 +76,7 @@ public class Sep31Service {
     validateAmount("", request.getAmount());
     validateLanguage(appConfig, request.getLang());
     validateRequiredFields(request.getAssetCode(), request.getFields().getTransaction());
-
-    validateAsset(request);
+    AssetInfo asset = validateAsset(request);
     validateSenderAndReceiver(request);
 
     Sep31Transaction txn =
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -93,17 +92,26 @@ public Sep31PostTransactionResponse postTransaction(
...
   }
 
+  private void generateTransactionMemo(Sep31Transaction txn) throws SepException {
+    String memo = StringUtils.truncate(txn.getId(), 32);
+    memo = StringUtils.leftPad(memo, 32, '0');
+    memo = new String(Base64.getEncoder().encode(memo.getBytes()));
+    txn.setStellarMemo(memo);
+    txn.setStellarMemoType(memoType(MEMO_HASH));
+  }
+
```
marcelosalloum, line 114: I think we should make the memo values less predictable. Nowadays I believe our IDs are incrementally generated, so we should probably update our id generator strategy to UUID by changing [this](https://github.com/stellar/java-stellar-anchor-sdk/blob/9df535a5712b9645a0298a917ee6d8de40641b6d/data-spring-jdbc/src/main/java/org/stellar/anchor/server/data/JdbcSep31Transaction.java#L22) to:

```
@Id
@GeneratedValue
UUID id;
```

If we don't, someone could mess with our watchers by sending 1 stroop payments, for instance.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -120,4 +121,34 @@ public StellarPaymentObserver build() {
...
+                  public void onSent(PaymentOperationResponse payment) {
+                    System.out.println("Sent:" + new Gson().toJson(payment));
+                  }
+                })
+            .build();
+
+    watcher.start();
+    Thread.sleep(300000);
+    watcher.shutdown();
+  }
```
marcelosalloum, line 153: Did you mean to commit this? It's just for tests, right?

Just checking, I'm fine with it being in the code, so we can test it easily.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -24,15 +26,29 @@ public class PaymentOperationToEventListener implements PaymentListener {
...
     if (txn == null) {
-      Log.error(String.format("no transaction(stellarAccountId=%s) is found.", payment.getTo()));
+      Log.info(String.format("no transaction(stellarAccountId=%s) is found.", payment.getTo()));
       return;
     }
 
     if (!(payment.getAsset() instanceof AssetTypeCreditAlphaNum)) {
-      // Asset does not match
+      // Asset does not match. Ignore.
+      Log.info(String.format("unexpected payment type %s", payment.getAsset().getClass()));
```
marcelosalloum, line 51: ‼️   Seems like you're validating the stellar payment `memo` and `asset code`, but you're not validating the amount nor the asset issuer. This way, someone could send one stroop and e approved, which is pretty bad.

I think we should make sure the amount received is `>=` the amount expected.

marcelosalloum (commented):
Looks great! I think we're missing some important validations before marking the transaction as `pending_receiver` or `completed` though:

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
marcelosalloum: I'll address this one in the PR I've been working on, since this is not an issue you introduced but something that was already in the code.

lijamie98: Yes, we should validate the asset issuer and the amount.

marcelosalloum (approved):

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
lijamie98: 😄

PR #195, opened 2022-04-14
core: Add inclusive and additive amount calculation in SEP-31

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add inclusive and additive amount calculation in SEP-31. Default to "inclusive"

### Why

There are anchors who calculate amount in the following equation.
amount_in = amount + fee (additive)

### Known limitations
Demo wallet won't work with `additive` settings.


core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -254,11 +254,65 @@ void validateAndApplyFeeAndQuote(
...
+      txn.setAmountOutAsset(txn.getAmountInAsset());
+    } else {
+      // Quote exists
       txn.setQuoteId(request.getQuoteId());
+      Sep38Quote quote = sep38QuoteStore.findByQuoteId(request.getQuoteId());
+      BigDecimal buyAmount = decimal(quote.getBuyAmount(), asset);
+      if (fee.getAsset().equals(txn.getAmountInAsset())) {
+        // Fee is applied to the amount_in
+        // amount_out = quote.buy_amount
+        txn.setAmountOut(String.valueOf(buyAmount));
```
JakeUrban, line 305: If fee is applied to `amount_in`, `amount_out` will still be less than `buy_amount`, unless `additive` is specified.

JakeUrban (changes requested):
Looks 95% good, but I think there is one issue with the amount / fee calculations.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -370,6 +427,27 @@ static Sep31InfoResponse createFromAssets(List<AssetInfo> assetInfos) {
...
+        context.set(new Context());
+      }
+      return context.get();
+    }
+
+    public static void reset() {
+      context.set(null);
+    }
+  }
+
```
marcelosalloum, line 450: ❗ I don't think this is thread safe. What if we're processing multiple transaction contexts at the same time?

Instead of using a static context variable, we should use different instances of context every time we need to process a different transaction.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -132,13 +136,30 @@ public Sep31PostTransactionResponse postTransaction(
...
         .id(txn.getId())
-        .stellarAccountId(asset.getDistributionAccount())
+        .stellarAccountId(Context.get().getAsset().getDistributionAccount())
         .stellarMemo(txn.getStellarMemo())
         .stellarMemoType(txn.getStellarMemoType())
         .build();
   }
 
-  private void generateTransactionMemo(Sep31Transaction txn) throws SepException {
+  private void applyAmount() {
```
marcelosalloum, line 145: 💡 maybe we should rename it to `calculateAmountIn()`? There are 3 types of amount for in, out and fee, so amount is not specific enough for this method name.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -254,11 +254,65 @@ void validateAndApplyFeeAndQuote(
                 "Quote asset is [%s] different from the transaction sending asset [%s]",
                 quote.getSellAmount(), txn.getAmountIn()));
       }
+    }
+  }
+
+  void queryFee() throws AnchorException {
+    Sep31PostTransactionRequest request = Context.get().getRequest();
+    JwtToken token = Context.get().getJwtToken();
+    org.stellar.platform.apis.shared.Amount fee =
```
marcelosalloum, line 263: nit: can we remove the full qualifier `org.stellar.platform.apis.shared.` and use `Amount` directly?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
marcelosalloum: I see a few problems with this approach:
* This works like a singleton, which is an anti-pattern.
* The name `Context` is very broad, we should use something more specific.
* This class handles POST, PATCH, and GET but all of them have access to this static singleton Context, meaning any request can interfere with the Context values. Multiple POST requests running in parallel certainly will interfere with each other and cause unpredictable results.

My suggestion is to actually create a local instance of this `PostTxContext` and pass it around to the private methods that need access to it. Or even better: only pass over the values that need to be used. For instance, if a `validateSenderAndReceiver` method just needs the sender and receiver you should only pass these properties, this will make the code more predictable and easier to write tests for it.

core/src/main/java/org/stellar/anchor/util/MathHelper.java
```diff
@@ -1,12 +1,18 @@
...
+import java.math.RoundingMode;
+import org.stellar.anchor.asset.AssetInfo;
 
 public class MathHelper {
   public static BigDecimal decimal(String value) {
     return new BigDecimal(value);
   }
 
+  public static BigDecimal decimal(String value, AssetInfo asset) {
+    return new BigDecimal(value).setScale(asset.getSignificantDecimals(), RoundingMode.HALF_UP);
```
marcelosalloum, line 13: 🤔  I think we should actually round HALF_DOWN.

marcelosalloum (changes requested):
The approach looks good! I've found a couple of problems though.

marcelosalloum:
Also, it would be great to add tests to make sure we cover all these cases. This part of the code is a bit complicated and we should have tests in place to make sure a change won't break things here.

config-spring-property/src/main/java/org/stellar/anchor/server/config/PropertySep31Config.java
```diff
@@ -7,4 +7,5 @@
 public class PropertySep31Config implements Sep31Config {
   boolean enabled = false;
   String feeIntegrationEndPoint = "http://localhost:8081";
+  String amountCalculation = "inclusive";
```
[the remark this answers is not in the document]
lijamie98: Yes, this is a default value that would be overwritten by the yaml value. 

I will give it a try to see how enum would work.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: This is actually thread-safe. 😄 Line #437 is an object of ThreadLocal. ([Some intro here](https://www.baeldung.com/java-threadlocal)). It is designed to store context of a thread execution.

1. The singleton is the ThreadLocal and the objects are per-thread. 😃 
2. The class name `Context` is nested in Sep31Service. To access the class, the name would be Sep31Service.Context. 
3. This should be the same as #1. 

I will be happy to call if it is not explained clearly.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: I can refactor the calculation if the diagram is accurate. https://lucid.app/lucidchart/55d25dfd-e895-4347-bea3-15a05bb6ccfa/edit?page=ZXw_~8n.uJ7c&invitationId=inv_7ae1fca7-04d1-4023-a650-8606c5627ed8#

core/src/main/java/org/stellar/anchor/util/MathHelper.java
lijamie98: Use HALF_DOWN.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -128,13 +138,109 @@ public Sep31PostTransactionResponse postTransaction(
...
+          // amount_in=req.amount
+          // amount_out=quote.buy_amount
+          amountIn = reqAmount;
+          amountOut = buyAmount;
+          // Validate:
+          //   quote.sell_amount = amount_in - fee
+          if (sellAmount.compareTo(amountIn.subtract(fee)) != 0) {
+            throw new BadRequestException(
+                "Bad quote. quote.sell_amount does not equal to (amount_in - amount_fee)");
+          }
```
marcelosalloum, line 198: This is tricky, right? This is part of why we're having these discussions 😥

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -128,13 +137,109 @@ public Sep31PostTransactionResponse postTransaction(
...
+          //   quote.sell_amount = req.amount
+          if (sellAmount.compareTo(reqAmount) != 0) {
+            throw new BadRequestException(
+                "Bad quote. quote.sell_amount does not equal to amount_in");
+          }
+        } else {
+          // amount_in=req.amount
+          // amount_out=quote.buy_amount -fee
+          amountIn = sellAmount;
+          amountOut = buyAmount.subtract(fee);
```
marcelosalloum, line 228: I think you're missing a validation here, `amount_in=req.amount` is not being validated.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -250,11 +334,31 @@ void validateAndApplyFeeAndQuote(
...
+  void queryFee() throws AnchorException {
+    Sep31PostTransactionRequest request = Context.get().getRequest();
+    JwtToken token = Context.get().getJwtToken();
+    Amount fee =
+        feeIntegration
+            .getFee(
+                GetFeeRequest.builder()
+                    .sendAmount(request.getAmount())
+                    .sendAsset(request.getAssetCode())
+                    .receiveAsset(request.getAssetCode())
```
marcelosalloum, line 349: ❗ I think this is supposed to be the destination asset.

```suggestion
                    .receiveAsset(request.getDestinationAsset())
```

core/src/test/kotlin/org/stellar/anchor/sep31/Sep31ServiceTest.kt
```diff
@@ -0,0 +1,332 @@
...
+import org.stellar.anchor.event.EventService
+import org.stellar.anchor.integration.customer.CustomerIntegration
+import org.stellar.anchor.integration.fee.FeeIntegration
+import org.stellar.anchor.sep10.JwtService
+import org.stellar.anchor.sep38.PojoSep38Quote
+import org.stellar.anchor.sep38.Sep38QuoteStore
+import org.stellar.anchor.util.GsonUtils
+import org.stellar.platform.apis.shared.Amount
+
+internal class Sep31ServiceTest {
```
marcelosalloum, line 29: Great to see some tests here 🎉

marcelosalloum (approved):
Good job! Plz fix the comment with "❗" before merging

🎉

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: Yeah.... hope we can take the "Quote includes fee and tax" approach. Then we don't have these tricky calculations.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: I had it and took it out. But I cannot remember the reason why I took it out. Added it back.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: Cool catch. Thanks so much.

JakeUrban (approved):
I haven't taken another look but want to remove my previous blocking review.

PR #196, opened 2022-04-14
Reference server config

stfung77:
update docker image for reference server to use config file
make kafka/event variables configurable in the reference server


anchor-reference-server/src/main/java/org/stellar/anchor/reference/AnchorEventConsumerService.java
```diff
@@ -103,7 +101,7 @@ public void run() {
                                             .build()))
                                 .build();
                         PlatformApiClient platformClient =
-                            new PlatformApiClient("http://localhost:8080");
+                            new PlatformApiClient(eventSettings.getSep31Endpoint());
```
lijamie98, line 104: Can we rename to `eventSettings.getPlatformApiEndpoint()`?

lijamie98 (approved):
LGTM except one minor comment.

PR #204, opened 2022-04-21
core: Add signature algorithm validation

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Added signature algorithm validation.

### Why

This is to avoid the security vulnerability https://www.oracle.com/security-alerts/cpuapr2022.html

We should make sure the signature is accepted only when it is signed by H256. 

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep10/JwtService.java
+++ b/core/src/main/java/org/stellar/anchor/sep10/JwtService.java
@@ -1,11 +1,13 @@
 package org.stellar.anchor.sep10;
 
 import io.jsonwebtoken.*;
+import io.jsonwebtoken.impl.DefaultJwsHeader;
+import org.apache.commons.codec.binary.Base64;
+import org.stellar.anchor.config.AppConfig;
+
 import java.nio.charset.StandardCharsets;
 import java.util.Calendar;
 import java.util.Map;
-import org.apache.commons.codec.binary.Base64;
-import org.stellar.anchor.config.AppConfig;
 
 public class JwtService {
   final String jwtKey;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/stellar/anchor/sep10/JwtService.java
```diff
@@ -46,6 +48,16 @@ public JwtToken decode(String cipher) {
...
     Jwt jwt = jwtParser.parse(cipher);
+    Header header = jwt.getHeader();
+    if (!(header instanceof DefaultJwsHeader)) {
+      // This should not happen
+      throw new IllegalArgumentException("Bad token");
+    }
+    DefaultJwsHeader defaultHeader = (DefaultJwsHeader) header;
+    if (!defaultHeader.getAlgorithm().equals(SignatureAlgorithm.HS256.getValue())) {
+      // Not signed by the JWTService.
+      throw new IllegalArgumentException("Bad token");
```
leighmcculloch, line 59: 💡 Suggest distinguishing these exceptions with something like the following:
```suggestion
      throw new IllegalArgumentException("Unrecognized algorithm");
```

leighmcculloch (approved):

leighmcculloch (commented):
❗ There does not appear to be any tests for this. Can you add tests?

core/src/main/java/org/stellar/anchor/sep10/JwtService.java
lijamie98: This is purposely hidden not to reveal the algorithm that we are using for the attackers. 

Probably not too meaningful for they can tell from the signature. But just some extra precausion.

lijamie98:
> ❗ There does not appear to be any tests for this. Can you add tests?

Great catch. Was in an unnecessary hurry and missed the tests.

core/src/main/java/org/stellar/anchor/sep10/JwtService.java
leighmcculloch: The exception details make it up into the API response? We should probably revisit that, as exception details can contain information we don't want to leak in APIs.

lijamie98: Yes, the exception description does bubble up to the exception handler unless we catch and wrap. The default exception handler does include the reason of the failure. So, we pay extra attention to the security related messages. 

I think this is probably ok for most API errors. But for security, it would help to have an extra pair of eyes to see if we miss anything.

C0x41lch0x41: In this specific case the exception is caught and 403 Forbidden is returned which is good in terms of security.
I think even the generic exception handler should not return the exception message. I'll open an issue to discuss this.

PR #205, opened 2022-04-22
Start Circle payment observer

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Start listening and processing circle notifications.

### Future work

The next steps include:
* Create and register event listeners
* Trigger an event from the Circle notification observer
* Write tests
* Update the SEP-31 code that generates the memo so it uses Circle's infrastructure to generate a memo.


platform/src/main/java/org/stellar/anchor/platform/controller/CirclePaymentObserverController.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.reflect.Type;
+import java.util.Map;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.*;
+import org.stellar.anchor.platform.paymentobserver.CirclePaymentObserverService;
+import shadow.com.google.common.reflect.TypeToken;
+
+@RestController
+@RequestMapping("/circle-observer")
+public class CirclePaymentObserverController {
```
lijamie98, line 13: What is the purpose of this controller? Is this used for Circle to push notification to the `platform`?

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/CirclePaymentObserverService.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  /**
+   * Handle incoming circle notifications of type "transfers". A transfer notification can contain circle<>circle or circle<>stellar events.
+   * @param circleNotification is the circle notification object.
+   */
+  public void handleTransferNotification(CircleNotification circleNotification) {
+    TransferNotificationBody transferNotification =
+        gson.fromJson(circleNotification.getMessage(), TransferNotificationBody.class);
```
lijamie98, line 104: Should we add a handler (controller or service) when message is mal-formed?

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/CirclePaymentObserverService.java
```diff
@@ -0,0 +1,217 @@
...
+
+    CircleTransfer circleTransfer = transferNotification.getTransfer();
+    if (circleTransfer == null) {
+      Log.error("Missing \"transfer\" value in notification of type \"transfers\".");
```
lijamie98, line 108: This can be treated the same way as line#113. I would suggest that we use Log.info

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/CirclePaymentObserverService.java
```diff
@@ -0,0 +1,217 @@
...
+    boolean isSourceOnStellar =
+        source.getType().equals(CircleTransactionParty.Type.BLOCKCHAIN)
+            && source.getChain().equals("XLM");
+    CircleTransactionParty destination = circleTransfer.getDestination();
+    boolean isDestinationOnStellar =
+        destination.getType().equals(CircleTransactionParty.Type.BLOCKCHAIN)
+            && destination.getChain().equals("XLM");
+
+    if (!isSourceOnStellar && !isDestinationOnStellar) {
+      Log.warn(
```
lijamie98, line 128: Log.info?

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/CirclePaymentObserverService.java
```diff
@@ -0,0 +1,217 @@
...
+          "Neither source nor destination are stellar accounts. Giving up on processing this transfer.");
+      return;
+    }
+
+    if (!circleTransfer.getAmount().getCurrency().equals("USD")) {
+      Log.error("The only circle currency supported is USDC.");
```
lijamie98, line 134: Log.info?

lijamie98 (approved):
LGTM.

Left a couple of comments.

platform/src/main/java/org/stellar/anchor/platform/controller/CirclePaymentObserverController.java
marcelosalloum: Yes, exactly! In the future, I believe we should move it to a separate service though. We can do it when we move the Stellar observer as well.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/CirclePaymentObserverService.java
marcelosalloum: 🤔  how exactly would that be used?

lijamie98: Something like in [Sep38Controller](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/platform/src/main/java/org/stellar/anchor/platform/controller/Sep38Controller.java#L99)

Or we may wrap it and throw like:
```java
TransferNotificationBody transferNotification 
try {
transferNotification =
        gson.fromJson(circleNotification.getMessage(), TransferNotificationBody.class);
} catch (JsonParseException jpex) {
        throw new BadRequestException("description here...");
}
```

marcelosalloum: Ok, I'll incorporate that in a future PR.

PR #234, opened 2022-05-04
SEP-31 + Circle: allow using circle for SEP-31 transaction memo and incoming address information

### LINKED ISSUE #220: Update SEP-31 to be able to use Circle to generate the transaction memos
A complementary part of Circle Payment Detection
GanttStart: 2022-04-27
GanttDue: 2022-04-29

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Allow configuring SEP-31 to be used along with Circle.

### Why

When using Circle, the SEP-31 transaction memo, memo type and destination address need to be created using the Circle API, so Circle can receive the incoming payment and notify the platform.

Close #220


api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31DepositInfo.java
```diff
@@ -0,0 +1,12 @@
+package org.stellar.anchor.api.sep.sep31;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class Sep31DepositInfo {
```
lijamie98, line 8: @marcelosalloum class `Sep31DepositInfo` seems to belong to Circle implementation and not part of the SEP-31 API. I think this class should belong to `platform` sub-project unless the deposit info is required for all SEP-31 implementations. 

If that's the case, we do not need to define the generator interface. We can make the generator part of the `JdbcSep31TransactionStore` where we update the deposit information.

core/src/main/java/org/stellar/anchor/sep31/Sep31DepositInfoGenerator.java
```diff
@@ -0,0 +1,13 @@
+package org.stellar.anchor.sep31;
+
+import org.stellar.anchor.api.sep.sep31.Sep31DepositInfo;
+
+public interface Sep31DepositInfoGenerator {
```
lijamie98, line 5: If the deposit info is specific to `platform`, we can move this class to `platform` subproject. But I start to see that you plan to make the deposit info generation customizable. 👍

platform/example.anchor-config.yaml
```diff
@@ -70,23 +70,24 @@ app-config:
     #     STRICT_SEND: amount_in = amount
     #     STRICT_RECEIVE: amount_in = amount + fee
     paymentType: STRICT_SEND
+    depositInfoGeneratorType: self # self or circle
```
lijamie98, line 73: Can we have some comments for `self`?

lijamie98 (approved):
LGTM with some comments.

api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31DepositInfo.java
marcelosalloum: I don't think it should be part of `JdbcSep31TransactionStore` because when using Circle we need to reach an external API, and API calls don't really fit inside database entities.

platform/example.anchor-config.yaml
marcelosalloum: Done in https://github.com/stellar/java-stellar-anchor-sdk/pull/234/commits/178cf4281c87d08ea762b15a71c0215aa6d29d81.

core/src/main/java/org/stellar/anchor/sep31/Sep31DepositInfoGenerator.java
marcelosalloum: Alright, this is what I ended up doing:
* The `Sep31DepositInfo` class and the `Sep31DepositInfoGenerator` interface stayed in core because they are used for the Sep31Service.
* I moved the `Sep31DepositInfoGeneratorSelf` implementation to platform.
* I created a `Sep31DepositInfoGeneratorCircle` implementation class, also in platform.
* `payments` doesn't hold any reference to `Sep31DepositInfo` nor `Sep31DepositInfoGenerator` anymore.

lijamie98: I think this is a good interim solution. 

After MVP, What I am thinking is to implement filter or observer interface pattern to generalize the interface and reduce the need for modifications when we have more use cases. 

I will come up with a simple architecture design and discuss with you.

PR #246, opened 2022-05-12
platform: Support reading asset JSON file from classpath, file and url

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Support reading asset JSON file from classpath, file and url

### Why

Allows anchor to define the list of supported assets.

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/asset/Assets.java
```diff
@@ -0,0 +1,10 @@
+package org.stellar.anchor.asset;
+
+import java.util.List;
+import lombok.Data;
+import org.stellar.anchor.api.sep.AssetInfo;
+
+@Data
+public class Assets {
+  List<AssetInfo> assets;
+}
```
marcelosalloum, line 10: Do you need this class at all?

I think we could just work with `List<AssetInfo>` instead.

marcelosalloum (approved):
LGTM!

core/src/main/java/org/stellar/anchor/asset/Assets.java
lijamie98: Thanks for pushing the improvement which is really appreciated. 

I do not like the current implementation either. However, This is needed for the [assets-test.json](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/platform/src/main/resources/assets-test.json#L2) 

I wish to refactor the format and schema of the asset file. I had created [issue #253](https://github.com/stellar/java-stellar-anchor-sdk/issues/253) with some thoughts to follow up.

PR #247, opened 2022-05-12
platform: Implement `HealthCheckService` and add `StellarPaymentObserver` checks.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

1. Added StellarPaymentObserver check

Here is an example of the health check. 

```JSON
{
    "started": "2022-05-13T13:58:15.591763Z",
    "elapsed_time_ms": 1,
    "number_of_checks": 1,
    "checks": {
        "stellar_payment_observer": {
            "streams": [
                {
                    "account": "GA7FYRB5VREZKOBIIKHG5AVTPFGWUBPOBF7LTYG4GTMFVIOOD2DWAL7I",
                    "threadShutdown": false,
                    "threadTerminated": false,
                    "stopped": false,
                    "lastEventId": "4064443516329985"
                }
            ]
        }
    }
}
```

### Why

N/A

### Known limitations

More health checks should be added.


platform/src/main/java/org/stellar/anchor/platform/service/HealthCheckContext.java
```diff
@@ -0,0 +1,3 @@
+package org.stellar.anchor.platform.service;
+
+public interface HealthCheckContext {}
```
marcelosalloum, line 3: What do we need this for? Seems like it could be deleted.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -156,34 +171,60 @@ public StellarPaymentObserver build() {
...
+
+      AtomicReference<String> lastEventId = getField(stream, "lastEventId", null);
+      if (lastEventId != null) {
+        builder.lastEventId(lastEventId.get());
+      }
+
+      results.add(builder.build());
+    }
+
+    return new SPOHealthCheckResult(getName(), results);
```
marcelosalloum, line 207: I think it'd be more efficient to use the `onFailure` and `onEvent` (with its try/catches) to periodically update a static variable.

Despite providing extra information, accessing the private variables seems too hacky.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AnchorPlatformIntegrationTest.kt
```diff
@@ -94,10 +93,10 @@ class AnchorPlatformIntegrationTest {
...
     val responseBody = response.body?.string()
     assertNotNull(responseBody)
-
-    val gotHealthCheck = SepClient.gson.fromJson(responseBody, HealthCheck::class.java)
-    val wantHealthCheck = HealthCheck(true)
-    assertEquals(wantHealthCheck, gotHealthCheck)
+    assertEquals(200, response.code)
+//    val gotHealthCheck = SepClient.gson.fromJson(responseBody, HashMap::class.java)
+//    assertNotNull(gotHealthCheck["number_of_checks"])
+//    assertNotNull(gotHealthCheck["checks"])
```
marcelosalloum, line 99: ❗  Could you add some test here?

We're deploying `/health` as a way for other services to check if Platform is up, so it makes sense to me that we have at least smoke tests to make sure it is working and exposing the info as expected. This way we can guarantee some reliability to this endpoint others will be relying on.

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheck.java
```diff
@@ -0,0 +1,40 @@
...
+  Map<String, HealthCheckResult> checks = new HashMap<>();
+
+  public HealthCheck() {
+    this.started = Instant.now();
+  }
+
+  public HealthCheck finish(List<HealthCheckResult> results) {
+    for (HealthCheckResult result : results) {
+      checks.put(result.name(), result);
+    }
```
marcelosalloum, line 30: 👀 if there are keys in `checks` that are not in `results`, we would pass the wrong information that they were checked when they actually weren't.

We should start a new `check` map here instead of simply calling `check.put`.

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheck.java
```diff
@@ -0,0 +1,40 @@
...
+  @SerializedName("number_of_checks")
+  int numberOfChecks;
+
+  Map<String, HealthCheckResult> checks = new HashMap<>();
+
+  public HealthCheck() {
+    this.started = Instant.now();
+  }
+
+  public HealthCheck finish(List<HealthCheckResult> results) {
```
marcelosalloum, line 27: What about renaming it to `wrap` or `complete` or `consolidate`? Finish sends the idea this service is being completely disabled, when in fact it's just wrapping up the sub-checks.

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheckResult.java
```diff
@@ -0,0 +1,5 @@
+package org.stellar.anchor.api.platform;
+
+public interface HealthCheckResult {
+  String name();
+}
```
marcelosalloum, line 5: What about adding a `List<Any> getStatuses()`? I think it would be more representative of how it'll be used.

marcelosalloum (commented):
I'm excited to see this landing 🎉

platform/src/main/java/org/stellar/anchor/platform/service/HealthCheckContext.java
lijamie98: The `HealthCheckContext` was for providing the context for each health check, such as environment variables, configuration profiles, etc. However, it is not used at this moment. 

I had removed the interface for now. We can add it back when needed.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
lijamie98: One of important check is to get the status of the SSEStream. It is possible that the thread died and the process is still running. 

I wish those variables are accessible from the class. However, the SDK was not designed to provide the status. That's why I added a function to access those variables. 

I had added an [issue #432 ](https://github.com/stellar/java-stellar-sdk/issues/432)in the Stellar Java SDK to make those variables non-private so that we may avoid the change of accessibility.

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheck.java
lijamie98: Cool suggestions. I had changed the name to `complete`.

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheckResult.java
lijamie98: I think we can use `List<String>` to reduce the complexity for now like the way ElasticSearch does [here.](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html#cluster-health-api-response-body) 

I had added two functions in the interface: `List<String> getStatuses` and `String getStatus()` with some comments in the interface.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AnchorPlatformIntegrationTest.kt
lijamie98: The tests are actually quite tricky because `Gson` throws the following exception while we call 
```java
gson.fromJson(responseBody, HealthCheckResponse::class.java)
```

Exception:
```
Caused by: java.lang.UnsupportedOperationException: Interface can't be instantiated! Interface name: org.stellar.anchor.api.platform.HealthCheckResult
	at com.google.gson.internal.UnsafeAllocator.assertInstantiable(UnsafeAllocator.java:117)
	at com.google.gson.internal.UnsafeAllocator$1.newInstance(UnsafeAllocator.java:49)
	at com.google.gson.internal.ConstructorConstructor$14.construct(ConstructorConstructor.java:225)
	... 80 more
```
That was the reason they were commented out. 

Nevertheless, I had added some tedious, verbose but necessary tests.

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
```diff
@@ -156,34 +172,96 @@ public StellarPaymentObserver build() {
...
-
-    watcher.start();
-    Thread.sleep(300000);
-    watcher.shutdown();
+  @Override
+  public HealthCheckResult check() {
+    List<StreamHealth> results = new ArrayList<>();
+    Status status = GREEN;
+    for (SSEStream<OperationResponse> stream : streams) {
+      StreamHealth.StreamHealthBuilder builder = StreamHealth.builder();
```
marcelosalloum, line 190: 💡 
```suggestion
      StreamHealth.StreamHealthBuilder healthBuilder = StreamHealth.builder();
```

platform/src/main/java/org/stellar/anchor/platform/service/HealthCheckService.java
```diff
@@ -0,0 +1,42 @@
...
+  HealthCheckService(List<HealthCheckable> checkables) {
+    for (HealthCheckable checkable : checkables) {
+      for (String tag : checkable.getTags()) {
+        List<HealthCheckable> checksOfTag = mapCheckable.get(tag);
+        if (checksOfTag == null) {
+          checksOfTag = List.of(checkable);
+          mapCheckable.put(tag, checksOfTag);
+        } else {
+          checksOfTag.add(checkable);
+        }
```
marcelosalloum, line 21: 💡 this suggestion is just a personal preference of mine, so no need to make the change if you don't want:

```suggestion
        if (checksOfTag == null) {
          checksOfTag = List.of();
          mapCheckable.put(tag, checksOfTag);
        }
        checksOfTag.add(checkable);
```

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheckResponse.java
```diff
@@ -0,0 +1,41 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class HealthCheckResponse {
+  Instant started;
```
marcelosalloum, line 13: 💡 
```suggestion
  final Instant started;
```

platform/src/main/java/org/stellar/anchor/platform/paymentobserver/StellarPaymentObserver.java
lijamie98: updated.

platform/src/main/java/org/stellar/anchor/platform/service/HealthCheckService.java
lijamie98: Good one. I used to do your way. But for some reason, I was doing the other way.
```java
        if (checksOfTag == null) {
          mapCheckable.put(tag, List.of());
        }
        checksOfTag.add(checkable);
```

api-schema/src/main/java/org/stellar/anchor/api/platform/HealthCheckResponse.java
lijamie98: updated.

Issue #254, opened 2022-05-15
sep-31 create transaction fails


reecexlm:
anchor validation tool is failing for SEP-31 

![image](https://user-images.githubusercontent.com/98427531/168486861-be18cc2a-9e15-4ced-a314-00ef74041a51.png)
![image](https://user-images.githubusercontent.com/98427531/168486872-f0439445-148e-4864-9e56-98d05b675885.png)

`Network Calls:
Request:
POST https://www.stellaranchordemo.com/sep31/transactions
Headers:
Authorization: Bearer..REMOVED
Content-Type: application/json
Body:
{
"sender_id":"8010f370-b222-4b92-99e0-fac497dcff7f"
"receiver_id":"8dad2c3c-89a2-479a-9ef7-6272a0cb658f"
"amount":100
"asset_code":"stellar:JPYC:GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP"
"fields":{
"transaction":{
"receiver_account_number":"123"
"receiver_routing_number":"456"
"type":"SWIFT"
}
}
}`

Exceptions seen in sep server
`17:40:56.583 INFO  - o.s.a.f.BaseTokenFilter - token created. account=GDOH...G75E url=https://www.stellaranchordemo.com/sep31/transactions
17:40:56.590 ERROR - o.s.a.p.c.GlobalControllerExceptionHandler - java.lang.NullPointerException
	at org.stellar.anchor.sep31.Sep31Service.postTransaction(Sep31Service.java:85)
	at org.stellar.anchor.platform.controller.Sep31Controller.postTransaction(Sep31Controller.java:45)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:150)
[...]

PR #256, opened 2022-05-16
core: Fix NullPointerException when the asset in the SEP-31 request not valid

lijamie98:
… found

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix issue #254 . 

When the asset code is invalid, we should return 400 instead of status code 500, which is caused by an NullPointerException.

### Why

N/A

### Known limitations

N/A


marcelosalloum (approved):
LGTM!

PR #262, opened 2022-05-17
core: Improve SEP-1, SEP-10 logging

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Improve the logging of SEP-1 and SEP-10

### Why

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
@@ -1,5 +1,12 @@
 package org.stellar.anchor.sep10;
 
+import static org.stellar.anchor.util.Log.*;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.*;
+import java.util.stream.Collectors;
 import org.stellar.anchor.api.exception.SepException;
 import org.stellar.anchor.api.exception.SepValidationException;
 import org.stellar.anchor.api.sep.sep10.ChallengeRequest;
@@ -14,14 +21,6 @@ import org.stellar.anchor.util.Sep1Helper.TomlContent;
 import org.stellar.sdk.*;
 import org.stellar.sdk.requests.ErrorResponse;
 import org.stellar.sdk.responses.AccountResponse;
-
-import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.*;
-import java.util.stream.Collectors;
-
-import static org.stellar.anchor.util.Log.*;
 
 /** The Sep-10 protocol service. */
 public class Sep10Service {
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -34,27 +33,35 @@ public class Sep10Service {
...
       throw new SepValidationException(
           String.format("home_domain [%s] is not supported.", challengeRequest.getHomeDomain()));
     }
 
     if (sep10Config.isClientAttributionRequired()) {
       if (challengeRequest.getClientDomain() == null) {
-        infoF("ALERT: client domain required and not provided");
+        info("client_domain is required but not provided");
         throw new SepValidationException("client_domain is required");
       }
```
marcelosalloum, line 67: Please allow me to discuss a concept here: I'd prefer to use `error`, rather than `info` logs when we're inside a `catch`, WDYT?

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -34,27 +33,35 @@ public class Sep10Service {
...
     this.serverAccountId = KeyPair.fromSecretSeed(sep10Config.getSigningSeed()).getAccountId();
   }
 
   public ChallengeResponse createChallenge(ChallengeRequest challengeRequest) throws SepException {
+    info("Creating challenge");
     //
     // Validations
     //
     if (challengeRequest.getHomeDomain() == null) {
+      infoF("home_domain is not specified. Use {}", sep10Config.getHomeDomain());
```
marcelosalloum, line 54: What's the reason for using `info` instead of `debug` here (and in other parts of the project)? It's more intuitive to me that debug is used when we're walking through the code, while Info is something more high-level. For instance:

```text
INFO:  Creating challenge
DEBUG: home domain is not specified
DEBUG: fetching account from Stellar network
ERROR: something went wrong when fetching the Stellar account.
```

Not a blocking change, nor it's written in stone... it's a concept that we will eventually reach a common definition as we work together.

core/src/test/kotlin/org/stellar/anchor/util/LogTest.kt
```diff
@@ -66,6 +77,55 @@ internal class LogTest {
...
+  @Test
+  fun testInfoBPII() {
+    val slot = slot<String>()
+    every { logger.info(capture(slot)) } answers {}
+
+    val testBeanPII = TestBeanPII()
+    Log.infoB("Hello", testBeanPII)
+    verify(exactly = 2) { logger.info(ofType(String::class)) }
+    assertFalse(slot.captured.contains("fieldPII"))
+  }
```
marcelosalloum, line 89: I've noticed the `slot` is not being used here... according with [this link](https://github.com/mockk/mockk/issues/352#issuecomment-652409698), you could use a list instead of a slot. It would look like this:

```suggestion
  @Test
  fun testInfoBPII() {
    val slotIMessages = mutableListOf<String>()
    every { logger.info(capture(slotIMessages)) } answers {}

    val testBeanPII = TestBeanPII()
    Log.infoB("Hello", testBeanPII)

    verify(exactly = 2) { logger.info(ofType(String::class)) }
    assertEquals("Hello", slotIMessages[0])
    val wantBean = """{
    "fieldNoPII": "no secret"
    }""".trimMargin()
    assertEquals(wantBean, slotIMessages[1])
  }
```

marcelosalloum (approved):
LGTM! I've left a few suggestions below ⬇️

PR #263, opened 2022-05-17
Postgres support + liquibase database migration

stfung77:
### What
1) Adding support for postgres (tested AWS aurora postgres and local postgres, example configs are in anchor-config-defaults.yaml)
2) Added Liquibase database migration and documentation on how to connect to DB and generate changlogs using JPA Buddy


anchor-reference-server/src/main/java/org/stellar/anchor/reference/AnchorReferenceServer.java
```diff
@@ -28,7 +28,8 @@ public static void start(int port, String contextPath) {
             .bannerMode(OFF)
             .properties(
                 String.format("server.port=%d", port),
-                String.format("server.contextPath=%s", contextPath));
+                String.format("server.contextPath=%s", contextPath),
+                "spring.liquibase.enabled=false");
```
marcelosalloum, line 32: Do we really need this line here? It'd be better to leave it in an existent yaml file, of possible.

platform/src/main/resources/anchor-config-defaults.yaml
```diff
@@ -250,10 +253,34 @@ data-spring-jdbc-sqlite:
...
+data-spring-jdbc-local-postgres:
+  spring.jpa.generate-ddl: true
+  spring.jpa.database: POSTGRESQL
+  spring.jpa.show-sql: false
+  spring.datasource.driver-class-name: org.postgresql.Driver
+  spring.datasource.url: jdbc:postgresql://localhost:5432/anchorplatform
+  spring.datasource.username: anchorplatform
+  spring.datasource.password: <PASSWORD>
+  spring.mvc.converters.preferred-json-mapper: gson
+  spring.liquibase.change-log: classpath:/db/changelog/db.changelog-master.yaml
```
marcelosalloum, line 278: 🤔 Is that ok to leave all of these database configurations uncommented? Just wondering if that could bring any side effects.

marcelosalloum (approved):
LGTM!

anchor-reference-server/src/main/java/org/stellar/anchor/reference/AnchorReferenceServer.java
stfung77: the unit tests specifically dont seem to pick up the value when set in the yaml

platform/src/main/resources/anchor-config-defaults.yaml
lijamie98: I think we can leave them un-commented. The user may do cherry picking by enabling different database configurations.

PR #277, opened 2022-05-23
all: Add/fix reference server health check and integration test.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
Add/fix reference server health check and integration test.

### Why

N/A

### Known limitations

Only KafkaListener is checked. We should add SQS checks.

### DIFF
--- a/anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
+++ b/anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
@@ -3,6 +3,7 @@ package org.stellar.anchor.reference.event;
 import static org.stellar.anchor.api.platform.HealthCheckStatus.GREEN;
 import static org.stellar.anchor.api.platform.HealthCheckStatus.RED;
 
+import com.google.gson.annotations.SerializedName;
 import java.time.Duration;
 import java.util.List;
 import java.util.Properties;
@@ -37,8 +38,7 @@ public class KafkaListener extends AbstractEventListener implements HealthChecka
     this.executor.submit(this::listen);
   }
 
-  public void listen() {
-    Log.info("Kafka event consumer server started ");
+  Consumer<String, AnchorEvent> createKafkaConsumer() {
     Properties props = new Properties();
 
     props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaListenerSettings.getBootStrapServer());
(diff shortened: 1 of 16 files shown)


anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
```diff
@@ -87,4 +96,50 @@ public void destroy() {
...
+    return "kafka_listener";
+  }
+
+  @Override
+  public List<String> getTags() {
+    return List.of("all", "kafka");
+  }
+
+  @Override
+  public HealthCheckResult check() {
```
stfung77, line 116: should we test if the service is able to connect to kafka as well? ex: call consumer.listTopics() just to check the connection.

https://stackoverflow.com/questions/41016075/check-if-system-is-connected-to-kafka

anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
```diff
@@ -87,4 +96,50 @@ public void destroy() {
...
+        .build();
+  }
+}
+
+@Data
+@Builder
+class KafkaHealthCheckResult implements HealthCheckResult {
+  transient String name;
+
+  List<String> statuses = List.of("green", "red");
```
marcelosalloum, line 135:
```suggestion
  List<String> statuses = List.of(GREEN.getName(), RED.getName());
```

integration-tests/src/main/kotlin/org/stellar/anchor/platform/ReferenceServerClient.kt
```diff
@@ -0,0 +1,24 @@
+package org.stellar.anchor.platform
+
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import okhttp3.Request
+
+class ReferenceServerClient(private val endpoint: String) : SepClient() {
+  fun health(checks: List<String>): java.util.HashMap<*, *> {
```
marcelosalloum, line 8: 💡 You could default the health check test to `"all"`:

```suggestion
  fun health(checks: List<String> = listOf("all")): java.util.HashMap<*, *> {
```

marcelosalloum (approved):

anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
lijamie98: This is great suggestion. The check is added.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/event/KafkaListener.java
lijamie98: Thanks for catching this. Corrected.

integration-tests/src/main/kotlin/org/stellar/anchor/platform/ReferenceServerClient.kt
lijamie98: Nice. Changed.

PR #286, opened 2022-05-31
SEP-31 + SEP-38: update SEP-31 flow to use SEP-38 quotes correctly

### LINKED ISSUE #193: event - update event models with transaction_status_changed field
update anchor event models with the latest schema described here:

https://github.com/stellar/stellar-anchor-platform/pull/31/files

update event producer/consumer to use these changes

### LINKED ISSUE #199: Collectively review SEP-31 document amount calculations. It should be simpler and contain examples to make it clearer how the implementation should be
### What

Collectively review SEP-31 document amount calculations. The documentation should become clearer and contain examples on how the anchors should calculate the amounts and apply the fees, especially when we have non-equivalent assets, and we're using quotes.

### Why

When trying to calculate the amounts and apply the fees, we came across a lot of questions:
* Are fees inclusive or additive?
* If the fees are inclusive, should the quotes be created based on the amounts post-fee?
* Can the quote.sell_amount differ from the `GET /transaction/:id` amount_in?
* Can the `GET /transaction/:id` amount_in differ from the `POST /transaction` amount?
* In a nutshell, we have a lot of amounts that are mutually connected and it's not clear how they should be calculated:
  * `POST /transaction` amount
  * `GET /transaction/:id` amount_in*, amount_out*, amount_fee and amount_fee_asset
  * `GET|POST /quote(/:id)` sell_amount* and buy_amount*.

_* indicates the amounts whose calculation is pretty confusing right now._

### What SEP-31 doc says

The documentation of amount-relevant fields says:

**amount (number)** - `POST /transaction`
Amount of the Stellar asset sent to the Receiving Anchor.

--
**amount_in** - `GET /transaction/:id`
(optional) The amount of the Stellar asset received or to be received by the Receiving Anchor. Excludes any fees charged after Receiving Anchor receives the funds.

**amount_fee** - `GET /transaction/:id`
(optional) The amount of fee charged by the Receiving Anchor.

**amount_out** - `GET /transaction/:id`
(optional) The amount sent by the Receiving Anchor to Receiving Client.

--
**sell_amount** - `POST|GET /quote(/:id)`
The amount of sell_asset the client would exchange for buy_asset.

**buy_amount** - `POST|GET /quote(/:id)`
The amount of buy_asset the client would like to purchase with sell_asset.

### Suggestion:

We should look at how Horizon solved this with strict-send and strict-receive path payments and make sure to follow a similar approach here. What matters to the client is either to define how much they are gonna send or how much the other party is going to receive.

### LINKED ISSUE #284: Update SEP-31 + SEP-38 flow based on the changes adopted in stellar/stellar-protocol#1204
Currently, when you create a SEP-31 transaction using a quote_id, the amounts are not being calculated/validated properly.
[These formulas](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#amount-formulas) that were added to SEP-31 when quote_id is used are new.  This user story uses the correct formulas validate & calculate the SEP31 amounts when a quote_id is provided. (

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Update SEP-31 flow to use SEP-38 quotes correctly.

### Why

Close #284.
Close #199.
Close #193.


core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -77,42 +75,65 @@ public Sep31InfoResponse getInfo() {
...
             .id(generateSepTransactionId())
             .status(SepTransactionStatus.PENDING_SENDER.toString())
-            .stellarAccountId(asset.getDistributionAccount())
-            .amountInAsset(assetInfo.getAssetName())
-            .amountIn(request.getAmount())
+            .statusEta(null)
+            .amountFee(fee.getAmount())
+            .amountFeeAsset(fee.getAsset())
+            .startedAt(Instant.now())
+            .completedAt(null)
```
lijamie98, line 113: Can we set the builder default values to `null` to reduce the verbosity.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -122,11 +143,23 @@ public Sep31PostTransactionResponse postTransaction(
...
+            .statusChange(
+                new TransactionEvent.StatusChange(null, TransactionEvent.Status.PENDING_SENDER))
+            .amountExpected(new Amount(txn.getAmountIn(), txn.getAmountInAsset()))
             .amountIn(new Amount(txn.getAmountIn(), txn.getAmountInAsset()))
             .amountOut(new Amount(txn.getAmountOut(), txn.getAmountOutAsset()))
             .amountFee(new Amount(txn.getAmountFee(), txn.getAmountFeeAsset()))
             .quoteId(txn.getQuoteId())
             .startedAt(txn.getStartedAt())
+            .updatedAt(txn.getStartedAt())
+            .completedAt(null)
```
lijamie98, line 156: Set the builder default values to `null` ?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
marcelosalloum: My intention here was to be verbose, actually. There were some fields that should be set but were not. Since there are so many fields, it's easy to miss something so I wanted to make it explicit which fields we're setting and which ones we're purposely not setting (the null ones).

This approach helps me answer the following question just by glancing at the code: "are we addressing all the fields?".

marcelosalloum: I'm open to discussing it further, just let me know.

PR #307, opened 2022-06-07
Merge core-release-1.0 commits by reworking.

### LINKED ISSUE #275: Merge `core-release-1.0` branch onto `main`
1. Merge core-release-1.0 onto main branch by redoing all the commits in the core-release-1.0 branch to main.
2. Publish `org.stellar.anchor-sdk:core:1.0.x` available at Maven central. 
3. Update MGI `StellarAdapterService` with `org.stellar.anchor-sdk:core:1.0.x`

lijamie98:
Closes #275 

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This is to redo the 8 commits ahead of the `main` branch pushed to `core-release-1.0` branch. 

### Why

Merge the commits. 

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -110,24 +106,30 @@ public InteractiveTransactionResponse withdraw(
...
             .assetIssuer(withdrawRequest.get("asset_issuer"))
             .startedAt(Instant.now().getEpochSecond())
             .stellarAccount(token.getAccount())
+            .stellarAccountMemo(token.getAccountMemo())
             .fromAccount(sourceAccount)
             .protocol(Sep24Transaction.Protocol.SEP24.toString())
-            .memoType(memoTypeString(MEMO_ID))
-            .domainClient(token.getClientDomain())
-            .build();
+            .domainClient(token.getClientDomain());
```
marcelosalloum, line 125: domainClient should actually be clientDomain, right?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -305,7 +316,12 @@ public GetTransactionResponse findTransaction(JwtToken token, GetTransactionRequ
       throw new SepNotFoundException("transaction is not found");
     }
 
-    return GetTransactionResponse.of(fromTxn(txn));
+    // If the token has a memo, make sure the transaction belongs to the account with the same memo.
+    if (token.getAccountMemo() != null && !token.getAccountMemo().equals(txn.getAccountMemo())) {
+      throw new SepNotFoundException("transaction is not found");
```
marcelosalloum, line 321: nit:
```suggestion
      throw new SepNotFoundException("transaction not found");
```

core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionBuilder.java
```diff
@@ -48,6 +48,11 @@ public Sep24TransactionBuilder stellarAccount(String stellarAccount) {
     return this;
   }
 
+  public Sep24TransactionBuilder stellarAccountMemo(String accountMemo) {
+    txn.setAccountMemo(accountMemo);
+    return this;
+  }
```
marcelosalloum, line 54: 💡  These names are inconsistent. Throughout other parts of the project, we use just `memo` and `memoType`.

core/src/main/java/org/stellar/anchor/util/SepHelper.java
```diff
@@ -47,6 +48,19 @@ public static String memoTypeString(MemoType memoType) {
...
+    } else if (memo instanceof MemoHash) {
+      return MEMO_HASH;
+    } else if (memo instanceof MemoText) {
+      return MEMO_TEXT;
+    } else if (memo instanceof MemoNone) {
+      return MEMO_NONE;
+    }
+    throw new SepException("Unsupported memo type: " + memo.getClass());
+  }
+
```
marcelosalloum, line 63: ©️  We already have a very similar function in MemoHelper. Could you reuse that? And feel free to change it to your needs.

https://github.com/stellar/java-stellar-anchor-sdk/blob/29c7ef66e94c0b9503ca68e15b07da064a76ee2d/core/src/main/java/org/stellar/anchor/util/MemoHelper.java#L35-L47

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
```diff
@@ -0,0 +1,105 @@
...
+      return fallback.toString();
+    } catch (SepValidationException e) {
+      return LANGUAGE_DEFAULT;
+    }
+  }
+}
+
+@Getter
+@AllArgsConstructor
+class Iso3316Language {
```
marcelosalloum, line 76: Can you leave a link to this ISO? I didn't find it in my search...

The representation I found (similar to what you showed here) was a combination of [ISO-639-1](https://en.wikipedia.org/wiki/ISO_639-1) and  [ISO-3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2): https://www.fincher.org/Utilities/CountryLanguageList.shtml

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
```diff
@@ -0,0 +1,105 @@
...
+                fallbacks.putIfAbsent(isoLang.getLangKey(), isoLang);
+              } catch (SepValidationException ex) {
+                errorEx(ex);
+              }
+            });
+    return fallbacks;
+  }
+
+  static final String LANGUAGE_ONLY_DEFAULT = "en";
+  static final String LANGUAGE_DEFAULT = "en-US";
```
marcelosalloum, line 55: 💡 The word `only` is a bit ambiguous to me. What about renaming these two to:
LANGUAGE_WITHOUT_LOCALE_DEFAULT
LANGUAGE_WITH_LOCALE_DEFAULT

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
```diff
@@ -0,0 +1,105 @@
...
+            lang -> {
+              try {
+                Iso3316Language isoLang = Iso3316Language.of(lang);
+                fallbacks.putIfAbsent(isoLang.getLangKey(), isoLang);
+              } catch (SepValidationException ex) {
+                errorEx(ex);
+              }
+            });
+    return fallbacks;
+  }
```
marcelosalloum, line 52: Hey, I'm a bit confused by your usage of the word fallback. What you are calling fallback languages are actually the configured languages found in the config files. The word fallback refers to "an alternative plan". So the `LANGUAGE_ONLY_DEFAULT`, for instance is what we use as a fallback, rather than the languages from AppConfig.

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
```diff
@@ -0,0 +1,105 @@
...
+
+  static final String LANGUAGE_ONLY_DEFAULT = "en";
+  static final String LANGUAGE_DEFAULT = "en-US";
+
+  static String getFallbackLanguage(String lang) {
+    try {
+      Iso3316Language language = Iso3316Language.of(lang);
+      Iso3316Language fallback = fallbackLangs.get(language.getLangKey());
```
marcelosalloum, line 60: 💡  Here you're comparing both the language and locale. Maybe we could fallback to the language only. 

For instance, `pt` and `pt_br` would be considered different in your logic, while in practice they're about the same thing.

Maybe a change for another PR though.

core/src/test/kotlin/org/stellar/anchor/sep10/JwtTokenTest.kt
```diff
@@ -0,0 +1,34 @@
...
+    val accountMemo = "135689"
+    val token =
+      JwtToken.of("iss", "$TEST_ACCOUNT:$accountMemo", issuedAt, expiresAt, "", TEST_CLIENT_DOMAIN)
+    assertEquals(TEST_ACCOUNT, token.account)
+    assertEquals(accountMemo, token.accountMemo)
+    assertEquals(TEST_CLIENT_DOMAIN, token.getClientDomain())
+    assertEquals(issuedAt, token.getIat())
+    assertEquals(expiresAt, token.getExp())
+    assertEquals(TEST_CLIENT_DOMAIN, token.getClientDomain())
+  }
```
marcelosalloum, line 33: Nice tests 🎉

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
```diff
@@ -0,0 +1,105 @@
...
+
+  static Map<String, Iso3316Language> prepareFallbackLanguages(AppConfig appConfig) {
+    Map<String, Iso3316Language> fallbacks = new HashMap<>();
+    appConfig
+        .getLanguages()
+        .forEach(
+            lang -> {
+              try {
+                Iso3316Language isoLang = Iso3316Language.of(lang);
+                fallbacks.putIfAbsent(isoLang.getLangKey(), isoLang);
```
marcelosalloum, line 46: ❓  Won't languages with the same prefix be overwritten? Like, if you had both `"zh-TW"` and `"zh-CN"`, wouldn't one of them be left out if you use `getLangKey`? Maybe you should use `toString()` instead.

marcelosalloum (commented):
The updates look great! I just added a few suggestions to make this code a little more polished.

core/src/main/java/org/stellar/anchor/sep10/JwtToken.java
```diff
@@ -51,6 +53,14 @@ public static JwtToken of(
...
     token.jti = jti;
     token.clientDomain = clientDomain;
+    String[] subs = sub.split(":", 2);
+    if (subs.length == 2) {
+      token.account = subs[0];
+      token.accountMemo = subs[1];
+    } else {
+      token.account = sub;
+      token.accountMemo = null;
+    }
```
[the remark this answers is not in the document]
lijamie98: I was actually debating about this. Instead of throwing an error, the strategy is fall-back to non-splitting and let the downstream processing to decide whether we should throw an error. 

I suggest we can leave the fall-back strategy for the JWT token and let the downstream to throw "Invalid account" exception.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
lijamie98: Thanks a lot. Don't know why I wrote `domainClient`.

Fixed.

core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionBuilder.java
lijamie98: This is actually the SEP-10 account memo which is different from the `memo` (transaction memo).

core/src/main/java/org/stellar/anchor/util/SepHelper.java
lijamie98: Cool. I had merged the two functions.

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
lijamie98: Thanks for bringing this up. I checked SEP-24 again. The `lang` field is RFC4646. 
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request

However, SEP-6 uses ISO-369-1. 
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#request

I think we should make the SEPs consistent. 

I had renamed the class to RFC4646Language and left a link.

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
lijamie98: Changed.

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
lijamie98: The word `fallback` is used in the[ slack discussion:](https://stellarfoundation.slack.com/archives/CHVFR99UJ/p1653323769451659) 

If no exact language is found, it will first fall back to a alternative language that is pre-defined. If no alternative language is found, it will fall back to a default. (We should make this configurable in the following PR).

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
lijamie98: There was a discussion in slack between Cassio, Niranjan and me. Here is the test cases that may better illustrate the strategy. We may of course make changes in the future PR to make strategy change or even make them configurable. 

https://github.com/stellar/java-stellar-anchor-sdk/blob/c5a3f2b398c30fb998216508f655fd6eef54a26d/core/src/test/kotlin/org/stellar/anchor/util/SepLanguageHelperTest.kt#L37-L58

core/src/main/java/org/stellar/anchor/util/SepLanguageHelper.java
lijamie98: The lang key defines how we should find the alternative language. Here is the language finding strategy:

1. Find the exact match. In this case, `zh-TW` will only match `zh-TW`
2. Find the alternative match. In this case, `zh-TW` will find some language that starts with `zh`. And that alternative language is the first occurrence in the `languages` list. (this is why the code is written this way)
3. Fallback to DEFAULT_LANGUAGE_WITHOUT/WITH_LOCALE_DEFAULT

marcelosalloum (approved):

core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionBuilder.java
lijamie98: This is not from SEP. It is from Polaris when I ported from Python to Java. 

Changing the name will break the MGI's implementation. I will add comments to the fields to avoid confusion.

PR #313, opened 2022-06-09
docs: Add more details of the `client_id` field of the fee callback API.

lijamie98:
The `client_id` field description is good only for SEP-31. This PR adds more descriptions for SEP-6 and SEP-24 and custodial account. 

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add more details of the client_id field of the fee callback API

### Why

adds more descriptions for SEP-6 and SEP-24 and custodial account. 

### Known limitations
The code should be modified to reflect the change.

### DIFF
--- a/docs/03 - Implementing the Anchor Server/Callbacks API.yml
+++ b/docs/03 - Implementing the Anchor Server/Callbacks API.yml
@@ -69,9 +69,12 @@ paths:
           description: |
             An identifier for the client application making the request. This ID can be used to offer different fees to different clients.
 
-            Client IDs will be Stellar public keys when the client is initiating a SEP-31 transaction. In this case the client should be a 
-            known SEP-31 sending anchor and the public key will be the Stellar address used to authenticate via SEP-10. Anchors must ensure
-            that the public key specified in the request matches a public key known to belong to the sending anchor.
+            A client ID can be one of the following:
+            1. the `client_domain` if the transaction is initiated from a non-custodial wallet. 
+            2. the Stellar custodial/omnibus account if the transaction is initiated from a custodian, such as Wyre.
+            3. The Stellar account of a SEP-31 sending anchor. In this case the client should be a known SEP-31 sending 
+               anchor and the public key will be the Stellar address used to authenticate via SEP-10. Anchors must ensure
+               that the public key specified in the request matches a public key known to belong to the sending anchor.
           schema:
             type: string
           required: true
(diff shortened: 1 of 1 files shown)


docs/Anchor Integration/Callbacks API.yml
```diff
@@ -69,9 +69,12 @@ paths:
           description: |
             An identifier for the client application making the request. This ID can be used to offer different fees to different clients.
 
-            Client IDs will be Stellar public keys when the client is initiating a SEP-31 transaction. In this case the client should be a 
-            known SEP-31 sending anchor and the public key will be the Stellar address used to authenticate via SEP-10. Anchors must ensure
-            that the public key specified in the request matches a public key known to belong to the sending anchor.
+            A client ID can be one of the following:
+            1. the `client_domain` if the transaction is initiated from a non-custodial wallet. 
+            2. the Stellar custodial/omnibus account if the transaction is initiated from a custodian, such as Wyre.
+            3. The Stellar account of a SEP-31 sending anchor. In this case the client should be a known SEP-31 sending
```
JakeUrban, line 75: In terms of phrasing, I think we should make it clear that the public key used in both option 2 & 3 should be known to the anchor. I think right now, it is not clear that the anchor needs to match the public key used in 2.

In 2, the public key belongs to a custodial wallet. In 3, it belongs to a SEP-31 sending anchor. Either way, it is a public key that the anchor should include in a map or list they have.

JakeUrban (commented):
Thanks for adding, but I think we can make it a bit more clear that option 2 and 3 are effectively the same from the perspective of the anchor.

docs/Anchor Integration/Callbacks API.yml
lijamie98: @JakeUrban I think (2) and (3) are actually in distinguishable from the receiving anchor's point of view. Is that correct?

lijamie98: @marcelosalloum Sure, I will review the API docs for `client_id`.

lijamie98: @JakeUrban I think (2) is used in SEP-6 and SEP-24 and (3) is used in SEP-31. Can you elaborate the reason they should match?

JakeUrban: Thats right. Maybe we can use `client_id` but keep the docs simple by only describing the SEP-31 case right now? Then we can update the docs when we add SEP-24/6. wdyt?

lijamie98: I think that's good point that we only describe for the SEPs that we support. Currently, SEP-24 is implemented although not officially supported. I will the description of SEP-24 and leave out SEP-6.

In the near future, I think we may start to separate the `develop` and `main` branch in the post-mvp.

JakeUrban (approved):

PR #319, opened 2022-06-10
core: Ensure `quotes_required` and `quotes_supported` for SEP-31 are handled

erika-sdf:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add support for SEP-31 `quotes_required` and `quotes_supported`.

### Why

The anchor platform should reject invalid configuration requests  for these values should.


erika-sdf:
Depends on https://github.com/stellar/stellar-anchor-tests/pull/90/files

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -326,9 +326,13 @@ void validatePatchTransactionFields(Sep31Transaction txn, Sep31PatchTransactionR
...
-  void preValidateQuote() throws AnchorException {
+  void preValidateQuote(boolean isQuotesRequired) throws AnchorException {
     Sep31PostTransactionRequest request = Context.get().getRequest();
 
+    if (isQuotesRequired && request.getQuoteId() == null) {
+      throw new BadRequestException("quotes_required is set to true; quote id cannot be empty");
+    }
+
     // Check if quote is provided.
     if (request.getQuoteId() == null) {
```
marcelosalloum, line 337: 👀  Can you also add a condition where `quoteSupported == false` also returns early?

```suggestion
    if (!quoteSupported || request.getQuoteId() == null) {
```

core/src/test/resources/test_assets.json.quotes_not_supported
```diff
@@ -0,0 +1,208 @@
+{
+  "assets": [
+    {
+      "schema": "stellar",
+      "code": "USDC",
```
marcelosalloum, line 5: 💡  It seems like you're only using this file in one place, and it's only used to test if the validation fails when quotes are marked as required but not supported. With that in mind, you can probably reduce this file a lot, by leaving just one asset and only the configurations you need for the test. WDYT?

We already have a few similar files in the project and having all configs in all files is a bit hard to maintain, so I'd go leaner with just what you need for this test.

Also, the file name could be more accurate, with something like `test_assets.json.quotes_required_but_not_supported`.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
marcelosalloum: Also, could you plz add a test for this new condition?

marcelosalloum (commented):
Congrats, the code and tests are pretty thorough 🎉.

I've added a couple of comments below, one with a missing validation and another with a suggestion 💡.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -326,9 +326,13 @@ void validatePatchTransactionFields(Sep31Transaction txn, Sep31PatchTransactionR
     }
   }
 
-  void preValidateQuote() throws AnchorException {
+  void preValidateQuote(boolean isQuotesRequired) throws AnchorException {
```
lijamie98, line 329: `assetInfo` is available by calling `Context.get().getAsset();` WDYT if we remove `boolean isAuotesRequired`? 

`
void preValidateQuote() throws AnchorException {
    AssetInfo assetInfo = Context.get().getAsset();
    boolean isQuotesRequired = assetInfo.getSep31().isQuotesRequired();
....
}

`

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -99,7 +99,7 @@ public Sep31PostTransactionResponse postTransaction(
     }
     validateRequiredFields(assetInfo.getCode(), request.getFields().getTransaction());
```
lijamie98, line 100: WDYT if we also remove the arguments passing to `validateRequiredFields` if possible to make the style more consistent.  But feel free not to take care of this at this moment.

erika-sdf:
> Congrats, the code and tests are pretty thorough 🎉.
> 
> I've added a couple of comments below, one with a missing validation and another with a suggestion 💡.

Thanks for the review. I've addressed the changes below.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
erika-sdf: Thanks for the suggestion, I've made this change

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
erika-sdf: This has been addressed, as well as adding a new call to `AssetService` to get `AssetInfo` with only the asset code.

core/src/main/java/org/stellar/anchor/asset/JsonAssetService.java
```diff
@@ -19,12 +19,22 @@ public List<AssetInfo> listAllAssets() {
...
+    }
+    return null;
+  }
+
+  public AssetInfo getAsset(String code, String issuer) {
+    if (issuer == null) {
+      return getAsset(code);
+    }
+    for (AssetInfo asset : assets.assets) {
+      if (asset.getCode().equals(code) && issuer.equals(asset.getIssuer())) {
```
lijamie98, line 36: 💡 nit: Can we use the safer version of comparison in case `asset.code()` is null? 
```java
if (code.equals(asset.getCode()) && issuer.equals(asset.getIssuer())) {
```

lijamie98 (approved):
LGTM.

core/src/main/java/org/stellar/anchor/asset/JsonAssetService.java
```diff
@@ -19,12 +19,22 @@ public List<AssetInfo> listAllAssets() {
...
-        }
+      }
+    }
+    return null;
+  }
+
+  public AssetInfo getAsset(String code, String issuer) {
+    if (issuer == null) {
+      return getAsset(code);
+    }
```
marcelosalloum, line 34: nit: the design I see more often in libraries and SDKs is for the method with fewer parameters to call the one with the most parameters, for instance:

```java
public AssetInfo getAsset(String code) {
  return getAsset(code, null);
}
```

Nothing wrong with your design though, we can move on with that.

PR #321, opened 2022-06-11
core: Refactor/clean up SEP24 fields and add field comments

lijamie98:
1. Renamed `stellarAccount` to `sep10Account`
2. Renamed `stellarAccountMemo` to `sep10AccountMemo`
3. Renamed `receivingAnchorAccount` to `withdrawAnchorAccount`
4. Removed un-related fields from MGI implementation.
5. Added field comments

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Refactor and clean up SEP-24 fields

### Why

Some of the fields are confusing. Rename, remove, and comment the fields.

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24RefundPayment.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24RefundPayment.java
@@ -1,11 +1,33 @@
 package org.stellar.anchor.sep24;
 
 public interface Sep24RefundPayment {
+  /**
+   * The payment ID that can be used to identify the refund payment. This is either a Stellar
+   * transaction hash or an off-chain payment identifier, such as a reference number provided to the
+   * user when the refund was initiated. This id is not guaranteed to be unique.
+   *
+   * @return the ID field.
+   */
   String getId();
 
+  /**
+   * <code>stellar</code> or <code>external</code>.
+   *
+   * @return the type.
+   */
   String getIdType();
 
+  /**
+   * The amount sent back to the user for the payment identified by id, in units of amount_in_asset.
+   *
+   * @return the amount.
+   */
   String getAmount();
 
+  /**
+   * The amount charged as a fee for processing the refund, in units of amount_in_asset.
+   *
+   * @return the fee.
+   */
   String getFee();
 }
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -123,7 +123,6 @@ public class Sep24Service {
             .stellarAccount(token.getAccount())
             .stellarAccountMemo(token.getAccountMemo())
             .fromAccount(sourceAccount)
-            .protocol(Sep24Transaction.Protocol.SEP24.toString())
             .clientDomain(token.getClientDomain());
 
     if (memo != null) {
@@ -232,7 +231,6 @@ public class Sep24Service {
             .stellarAccount(token.getAccount())
             .stellarAccountMemo(token.getAccountMemo())
             .toAccount(destinationAccount)
-            .protocol(Sep24Transaction.Protocol.SEP24.toString())
             .clientDomain(token.getClientDomain())
             .claimableBalanceSupported(claimableSupported);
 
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/stellar/anchor/sep24/Sep24RefundPayment.java
```diff
@@ -1,11 +1,33 @@
 package org.stellar.anchor.sep24;
 
 public interface Sep24RefundPayment {
+  /**
+   * The payment ID that can be used to identify the refund payment. This is either a Stellar
+   * transaction hash or an off-chain payment identifier, such as a reference number provided to the
+   * user when the refund was initiated. This id is not guaranteed to be unique.
```
marcelosalloum, line 7: ❓   Why it's not guaranteed to be unique? Is it because we could have multiple payments in the same tx?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -315,12 +312,13 @@ public GetTransactionResponse findTransaction(JwtToken token, GetTransactionRequ
     }
 
     // We should not return the transaction that belongs to other accounts.
-    if (txn == null || !txn.getStellarAccount().equals(token.getAccount())) {
+    if (txn == null || !txn.getSep10Account().equals(token.getAccount())) {
```
marcelosalloum, line 315: ❓ Which accounts could be used for the SEP-10? I'm assuming it could be a user wallet or a custodial wallet, is that correct?

core/src/main/java/org/stellar/anchor/sep24/Sep24Transaction.java
```diff
@@ -4,118 +4,259 @@
...
-  String getDocumentType();
-
-  void setDocumentType(String documentType);
-
+  /**
+   * Unique, anchor-generated id for the transaction.
+   *
+   * @return The <code>id</code> field of the SEP-24 transaction history.
+   */
   String getTransactionId();
```
marcelosalloum, line 22: It doesn't make much sense to me why we have both an ID and a transactionId. Do we really need both?

core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionBuilder.java
```diff
@@ -44,17 +44,17 @@ public Sep24TransactionBuilder completedAt(long time) {
...
-    txn.setStellarAccount(stellarAccount);
+    txn.setSep10Account(stellarAccount);
     return this;
   }
 
   public Sep24TransactionBuilder stellarAccountMemo(String accountMemo) {
-    txn.setAccountMemo(accountMemo);
+    txn.setSep10AccountMemo(accountMemo);
     return this;
   }
```
marcelosalloum, line 54: 👀  I think you forgot to update the name of these builder methods.

marcelosalloum (commented):
Looking good, I've added a few comments below.

Also, since you're changing the JdbcSep24Transaction, I think you may need to update the database migration files. Please, refer to [this doc](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/docs/02%20-%20Contributing/D%20-%20Database%20Migration.md) for instructions.

core/src/main/java/org/stellar/anchor/sep24/Sep24RefundPayment.java
lijamie98: I cannot figure out either. I posted a message in Slack.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
lijamie98: That's correct.

core/src/main/java/org/stellar/anchor/sep24/Sep24Transaction.java
lijamie98: This was to separate the database ID from the SEP24 transaction ID. In some cases, the SEP24 transaction ID cannot be used as a database ID.

core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionBuilder.java
lijamie98: Good catch. 👍 

Fixed.

core/src/main/java/org/stellar/anchor/sep24/Sep24Transaction.java
marcelosalloum: In the code do we use different ones or are we defaulting to use the same in both fields?

Or is this defined in the Anchor side?

marcelosalloum (approved):

lijamie98:
i will update the liquid base file in the following PR.

PR #330, opened 2022-06-14
hotfix: fix `gson` reference cycle triggered by a print

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix reference cycle caused by print.

### Why

For some reason, when we printed the request, it caused a reference cycle that went on until the computer memory is over, to finally throw a 500.

The reference cycle happened in:
https://github.com/stellar/java-stellar-anchor-sdk/blob/c6bfc92f9499f9efec4ec10cca6933a6c3088ac9/core/src/main/java/org/stellar/anchor/filter/BaseTokenFilter.java#L84

The error:

Screen Shot 2022-06-13 at 6 57 51 PM

> Note: these lines repeat in a loop, the full stack trace has hundreds of lines.


lijamie98 (approved):
LGTM.

PR #334, opened 2022-06-15
docs: document the step-by-step of how partners should implement their anchor servers

### LINKED ISSUE #233: Update testing documentation with information on how to run manual tests

### LINKED ISSUE #272: Document the step-by-step of how partners should implement their Anchor servers

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Document the step-by-step of how partners should implement their anchor servers.

### Why

To make implementation faster and easier.

Close #272
Close #233


docs/00 - Stellar Anchor Platform.md
```diff
@@ -66,6 +80,91 @@ As you can see, the Anchor Platform receives interactions from ecosystem partici
 
 This drastically reduces the amount of code that needs to be written by the Anchor, and allows them to focus on the business logic that's specific to their businesses and use cases.
 
+## SEP-31 Flow
```
lijamie98, line 83: This is awesome. 

Should we merge the [one in SEP-31 ](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#completing-a-transaction-with-a-firm-quote)with this or we would want to continue maintaining both?

docs/01 - Running & Configuring the Application/A - Running & Configuring the Application.md
```diff
@@ -80,6 +79,10 @@ If all of the above fail, the server fails with an error.
 
 Secrets are passed to the Anchor Platform (and Anchor Reference Server) via environment variables, which can be set either through command line or using a `.env` file. A list of supported environment variables and their descriptions can be found at [`example.env`].
 
+### Supported Assets
+
+The Anchor Platform reads the list of supported assets from a json file whose address is configured in the config file under `app-config.app-assets` and defaults to [`assets-test.json`] ([ref](https://github.com/stellar/java-stellar-anchor-sdk/blob/1f84429f0c5d35cee75445686242643fbd8cffa5/platform/src/main/resources/anchor-config-defaults.yaml#L74)).
```
lijamie98, line 84: Hi Marcelo, this link points to a specific commit instead of the HEAD of main. But I also think pointing to main would cause the line number being inaccurate. 

I don't have much experience in this. Is this the common practice?

platform/src/main/java/org/stellar/anchor/platform/EventsConfig.java
```diff
@@ -0,0 +1,26 @@
+package org.stellar.anchor.platform;
```
lijamie98, line 1: nit: This is a refactoring in a docs PR. 

But this is ok with me.

lijamie98 (approved):

docs/00 - Stellar Anchor Platform.md
marcelosalloum: That's a great point! I also thought of using that one, but the difference between both is that this one covers the interaction between `Platform<>Anchor`, which is a concept that's not part of the SEP itself.

With that in mind, I think we should maintain both. I recognize this means more work, but it is unavoidable as far as I can see.

docs/01 - Running & Configuring the Application/A - Running & Configuring the Application.md
marcelosalloum: Yeah, if we want to point to a specific line, we need to use a permalink as I did here... I'd love to reference `main`, but that would become inaccurate 😕 and harder to maintain 🔨 .

I think this approach is easier to maintain, but I'm open to change if there's something better.

Issue #297, opened 2022-06-03
Spring configs in yaml file not being honored by Anchor Platform


stfung77:
Spring configs (defined in anchor-config-defuaults.yaml) are not being honored by the Anchor Platform application. Current work around is to set the config in AnchorPlatformServer.java 
ex: line 44 & 45
<img width="778" alt="image" src="https://user-images.githubusercontent.com/11511032/171948528-d022be61-8a4c-47e5-9ca7-929b5d671c68.png">

cc: @lijamie98

marcelosalloum:
I think these configs are not being honored because we're bypassing the default configuration flow from spring boot.

I believe the fix for that is to manually inject the yaml properties as a spring boot source, like I did in this commit (that was not merged):
https://github.com/stellar/java-stellar-anchor-sdk/pull/334/commits/24d5a2cb8e03dc1d2a582f52432317c5e61ac29d#diff-c32d6fef3a0ecfcf799c1efbee4bad5b0f73c1cb82684e4c8adb7f4a6d9796f5R91-R94

```java
    ConfigurableEnvironment environment =
        (ConfigurableEnvironment) applicationContext.getEnvironment();
    for (PropertySource<?> source : sources) {
      environment.getPropertySources().addFirst(source);  // injecting the (yaml) source into the spring boot environments
```

marcelosalloum:
@lijamie98: any thoughts?

nirojapanda:
Can be tested along with #479

PR #335, opened 2022-06-15
Add Prometheus metrics 

stfung77:
Add Prometheus Metrics
- using Spring Actuator and Micrometer
- Actuator provides many metrics out of the box (http request, system metrics, etc...)
- Micrometer allows for a generic metrics interface and supports: Atlas, Datadog, Graphite, Ganglia, Influx, JMX, and Prometheus
- This PR adds Actuator metrics and custom Anchor Platform metrics via Prometheus support


build.gradle.kts
```diff
@@ -58,7 +58,9 @@ subprojects {
     implementation("org.springframework.cloud:spring-cloud-aws-messaging:2.2.6.RELEASE")
     implementation("org.postgresql:postgresql:42.3.5")
     implementation("org.liquibase:liquibase-core:4.10.0")
-
+    implementation("org.springframework.boot:spring-boot-starter-actuator:2.7.0")
```
lijamie98, line 61: 🤔 These dependencies will be applied to all sub-projects. If some of them are only used in the `platform` subproject, can we limit the scope?

platform/src/main/java/org/stellar/anchor/platform/AnchorPlatformServer.java
```diff
@@ -87,4 +90,9 @@ public EventPublishService eventService(
             String.format("Invalid event publisher: %s", eventConfig.getPublisherType()));
     }
   }
+
+  @Bean
+  public MetricEmitterService metricService(JdbcSep31TransactionStore sep31TransactionStore) {
```
lijamie98, line 95: This ties the metric emitter service to the specific JdbcSep31TransactionStore. If we switch the transaction store to database other than JDBC, such as Mongo, this will fail. 

IMO, the metrics should be considered separately from the analytics. The number of transactions matching certain status should belong to the analytics while the number of the transactions should belong to metrics. 

I think querying the database with status filter for Prometheus scraping may hinder the scalability. Can we not query the database at this moment? We may discuss this if you have other thoughts.

core/src/main/java/org/stellar/anchor/event/KafkaEventService.java
```diff
@@ -43,6 +44,9 @@ public void publish(AnchorEvent event) {
       ProducerRecord<String, AnchorEvent> record = new ProducerRecord<>(topic, event);
       record.headers().add(new RecordHeader("type", event.getType().getBytes()));
       producer.send(record);
+      Metrics.counter(
```
lijamie98, line 47: 👍 It is awesome to use a vendor neutral interface! (`io.micrometer.core.instrument.Metrics`)

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
```diff
@@ -0,0 +1,63 @@
...
+    Metrics.gauge("sep31.transaction", Tags.of("status", "pendingReceiver"), pendingReceiverTxns);
+    Metrics.gauge("sep31.transaction", Tags.of("status", "pendingExternal"), pendingExternalTxns);
+    Metrics.gauge("sep31.transaction", Tags.of("status", "completed"), completedTxns);
+
+    // TODO add gauges for SEP-24 Transactions
+
+  }
+
+  class MetricEmitter implements Runnable {
+    public void run() {
```
lijamie98, line 40: These queries are quite expensive in JDBC. These are usually taken care of in analytics rather than metrics. 

Here is a short introduction to the difference between metrics and analytics. https://infogram.com/blog/metrics-vs-analytics/

People usually send logs to search engine such as ElasticSearch to perform transaction analytics.

stfung77: as discussed last week with @lijamie98 and @reecexlm, the metrics that require database queries will be configurable and will be disabled by default

core/src/main/java/org/stellar/anchor/util/Log.java
```diff
@@ -69,6 +72,7 @@ public static void debugF(final String format, final Object... args) {
   public static void error(final String msg) {
     Logger logger = getLogger();
     logger.error(msg);
+    Metrics.counter("logger", "type", "error").increment();
```
marcelosalloum, line 75: What is the value of counting the amount of error/warning logs? I'm not opposed to this, just trying to understand the reason behind it.

docs/deployment-examples/prometheus/prometheus.yml
```diff
@@ -0,0 +1,14 @@
+# Sample Prometheus config
+# This assumes that your Prometheus instance can access this application on localhost:8080
```
marcelosalloum, line 2: When you say `this application`, do you mean the Platform? Can you update the wording to be a little bit clearer?

docs/deployment-examples/prometheus/prometheus.yml
```diff
@@ -0,0 +1,14 @@
...
+  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
+  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
+  # scrape_timeout is set to the global default (10s).
+
+scrape_configs:
+  - job_name: 'spring boot scrape'
+    metrics_path: '/actuator/prometheus'
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:8080']
```
marcelosalloum, line 14: Do you need to set the global scrape_interval if you will override it in the scrape configs?

build.gradle.kts
marcelosalloum: After looking into the code, seems like all you need in `core` is `io.micrometer:micrometer-registry-prometheus`.

platform/src/main/java/org/stellar/anchor/platform/ConfigManagementConfig.java
```diff
@@ -91,4 +91,10 @@ KafkaConfig kafkaConfig() {
   SqsConfig sqsConfig() {
     return new PropertySqsConfig();
   }
+
+  @Bean
+  @ConfigurationProperties(prefix = "metric-emitter-service")
```
marcelosalloum, line 96: please remember to update stellar/kube when you merge this code.

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep31TransactionRepo.java
```diff
@@ -18,4 +18,7 @@ public interface JdbcSep31TransactionRepo extends CrudRepository<JdbcSep31Transa
   Optional<JdbcSep31Transaction> findByStellarAccountId(@NonNull String stellarAccountId);
 
   Optional<Sep31Transaction> findByStellarMemo(String stellarMemo);
+
+  @Query(value = "SELECT COUNT(t) FROM JdbcSep31Transaction t WHERE t.status = :status")
+  Integer findByStatusCount(@Param("status") String status);
```
marcelosalloum, line 23: Nice, this is new to me! TIL!

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -91,6 +93,8 @@ public PatchTransactionsResponse patchTransactions(PatchTransactionsRequest requ
     for (JdbcSep31Transaction txn : txnsToSave) {
       // TODO: consider 2-phase commit DB transaction management.
       txnStore.save(txn);
+      Metrics.counter("sep31.transaction",
+              "status",txn.getStatus()).increment();
```
marcelosalloum, line 97: ‼️ This is not right. We cannot assume that all calls to `patchTransactions` are meant to update a specific status.

Firstly, you need to check if we're really updating a status here.

Also, to be accurate, you'll need to gather both the old and new statuses, so you can decrement the old status count and increment the new one.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -113,12 +116,17 @@ public void onReceived(ObservedPayment payment) {
...
+                "status", SepTransactionStatus.PENDING_RECEIVER.toString()).increment();
       } catch (SepException ex) {
         Log.errorEx(ex);
       }
     }
     // send to the event queue
     sendToQueue(event);
+    Metrics.counter(
+            "payment.received", "toAddress", payment.getTo(), "asset", payment.getAssetName())
+        .increment();
```
marcelosalloum, line 129: 👀  Getting one counter per receiver can be a hassle to visualize.

‼️  Also, some metrics tools charge per different metric, so a client with high throughput could end up spending a ton on metrics services if we ship this kind of metric. This is something to be logged rather than a metric.

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
```diff
@@ -0,0 +1,64 @@
...
+  AtomicInteger pendingStellarTxns = new AtomicInteger(0);
+  AtomicInteger pendingCustomerInfoUpdateTxns = new AtomicInteger(0);
+  AtomicInteger pendingSenderTxns = new AtomicInteger(0);
+  AtomicInteger pendingReceiverTxns = new AtomicInteger(0);
+  AtomicInteger pendingExternalTxns = new AtomicInteger(0);
+  AtomicInteger completedTxns = new AtomicInteger(0);
+
+  public MetricEmitterService(JdbcSep31TransactionRepo sep31TransactionRepo) {
+    this.sep31TransactionStore = sep31TransactionRepo;
+    this.executor.scheduleAtFixedRate(new MetricEmitter(), 0, 30, TimeUnit.SECONDS);
```
marcelosalloum, line 27: I don't think we need this level of real-time (every 30 seconds). With a huge database, this could become detrimental to performance. 

💡 Can we have the number of seconds configurable through MetricConfig?

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
marcelosalloum: 💡 Since this is going to be a recurring query, we should index the `status` column. Here's some background on that: https://www.baeldung.com/jpa-indexes

core/src/main/java/org/stellar/anchor/util/Log.java
stfung77: seems like something an operator would want to keep tabs on and maybe even alert on and look into if there are a high number of error/warnings.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
marcelosalloum: Not necessarily, because we may use custodial solutions like Circle that can provide a new receiving address per Transaction.

docs/deployment-examples/prometheus/prometheus.yml
stfung77: its just something i've seen in multiple other docker files (including this one that was copied). useful if you have multiple jobs and dont want to specify the scrape_interval for each one.  
I can get rid of it if you'd like

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
stfung77: we cant decrement counters. @lijamie98 @reecexlm  and i had a call on this the other day and decided to just increment a counter every time a transaction transitions into a different state. this can be useful on a time series graph if we see the number of transactions in `complete` vs 'pending_*' starting to diverge too much. other alerts can also be build arount this. 

i'll add a check for status changes

stfung77:
will add indexing in a separate PR after flyway changes are merged

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
```diff
@@ -0,0 +1,87 @@
...
+        pendingCustomerInfoUpdateTxns);
+    Metrics.gauge("sep31.transaction.db", Tags.of("status", "pending_sender"), pendingSenderTxns);
+    Metrics.gauge(
+        "sep31.transaction.db", Tags.of("status", "pending_receiver"), pendingReceiverTxns);
+    Metrics.gauge(
+        "sep31.transaction.db", Tags.of("status", "pending_external"), pendingExternalTxns);
+    Metrics.gauge("sep31.transaction.db", Tags.of("status", "completed"), completedTxns);
+    Metrics.gauge("sep31.transaction.db", Tags.of("status", "error"), errorTxns);
+
+    // TODO add gauges for SEP-24 Transactions
```
marcelosalloum, line 54: Can you create an issue for this?

marcelosalloum (commented):
Almost there! 

I've left a few suggestions below. In special, adding string literals is something that I'd like to avoid as it's hard to undo (or maintain) that in the future.

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
stfung77: https://github.com/stellar/java-stellar-anchor-sdk/issues/393

platform/src/main/java/org/stellar/anchor/platform/service/MetricEmitterService.java
```diff
@@ -0,0 +1,87 @@
...
+        pendingCustomerInfoUpdateTxns.set(
+            sep31TransactionStore.findByStatusCount("pending_customer_info_update"));
+        pendingSenderTxns.set(sep31TransactionStore.findByStatusCount("pending_sender"));
+        pendingReceiverTxns.set(sep31TransactionStore.findByStatusCount("pending_receiver"));
+        pendingExternalTxns.set(sep31TransactionStore.findByStatusCount("pending_external"));
+        completedTxns.set(sep31TransactionStore.findByStatusCount("completed"));
+        errorTxns.set(sep31TransactionStore.findByStatusCount("error"));
+      } catch (Exception ex) {
+        Log.errorEx(ex);
+      }
```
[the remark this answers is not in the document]
stfung77: https://github.com/stellar/java-stellar-anchor-sdk/issues/394

docs/01 - Running & Configuring the Application/A - Running & Configuring the Application.md
```diff
@@ -161,6 +162,20 @@ You can test against this setup by running the end-to-end tests ([end_to_end_tes
 
 The default configuration of the project uses a Stellar network observer to identify incoming Stellar payments. In case the Anchor relies on Circle, it should configure the project to use the Circle Payment Observer. For more information on how to do that, please refer to the [01.B - Circle Payment Observer](/docs/01%20-%20Running%20%26%20Configuring%20the%20Application/B%20-%20Circle%20Payment%20Observer.md) section.
 
+
+## Metrics
+The Anchor Platform exposes a Prometheus metrics endpoint at `<host>:8082/actuator/prometheus`. All standard Spring
```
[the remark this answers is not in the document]
stfung77: still dependent on https://github.com/stellar/java-stellar-anchor-sdk/issues/297

marcelosalloum (approved):
LGTM!

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -113,12 +115,17 @@ public void onReceived(ObservedPayment payment) {
       txn.setStellarTransactionId(payment.getTransactionHash());
       try {
         transactionStore.save(txn);
+        Metrics.counter(
+                "sep31.transaction", "status", SepTransactionStatus.PENDING_RECEIVER.toString())
+            .increment();
```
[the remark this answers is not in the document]
marcelosalloum: Hum, I don't think this solves the issue of avoiding string literals.

The problem with string literals is that we're humans and can't be sure if they're all correct. For instance, even though we've both reviewed your code we can't be sure there are no typos causing events to be generated into the wrong metric groups. You could easily have two calls like the following and not notice the issue:
* `Metrics.counter("logger", "type", "error").increment();`
* `Metrics.counter("logger", "typo", "error").increment();  // used "typo" instead of "type"`

A more appropriate approach here would be to have this enum in a package that's common to both core and platform. Why not define this enum inside `core` itself? This should make sense since you are already calling `Metrics.counter` from core, right?

docs/01 - Running & Configuring the Application/A - Running & Configuring the Application.md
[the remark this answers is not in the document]
marcelosalloum: The issue with hardcoding is that some partners could have another service in this same port. If we could have this config in a yaml file, this would be great.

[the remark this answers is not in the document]
stfung77: @marcelosalloum 
due to #297, the port config has to be specified in the code - ex: 
      builder.properties(
          "management.endpoints.web.exposure.include=health,info,prometheus",
          "management.server.port=8082");

this would be configurable when #297 is resolved and we are able to specify this config in the config file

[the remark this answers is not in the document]
marcelosalloum: Alright. I noticed that issue was not on the board yet, so we were not considering it in the sprint planning. Perhaps it can be tackled as part of the config refactor.

PR #345, opened 2022-06-20
sep service helm-chart initial

reecexlm:
This PR adds a helm chart for SEP service.    This helm chart has been tested by publishing the chart to a helm chart repository and then deploying it to a aws EKS, MSK test environment.  The Stellar validator service has run after deploying this chart successfully.    This has not yet been tested in SDF infrastructure.  

next step for this PR is to test helm chart (currently [published/hosted by s3 bucket](http://anchorplatformhelmchart.s3-website.us-east-2.amazonaws.com/) ) using [these instruction](https://github.com/stellar/ops/issues/1624#issuecomment-1096320478) and enable cicd for updates (helm package, helm repo index and publish)


stfung77:
this is awesome! question regarding the values.yaml file, looks like ~80% of the configmap values (and looks like some settings are not made configurable, ex: data-access) are coming from this values.yaml file, why not just have a separate configmap for each deployment?

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.fullName }}-configmap-{{ .Values.service.name }}
+  namespace: {{ .Values.namespace }}
+data:
+  # TODO: remove all secrets (ex: move to env vars, k8s secrets, vault, etc...)
```
jacekn, line 7: +1!

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
...
+          settings: app-config                        # The location of the configuration data
+        data-access:
+          type: data-spring-jdbc                      # Activate [config-spring-jdbc] module.
+          settings: data-spring-jdbc-sqlite           # The location of the configuration data in this file.
+        logging:
+          type: logging-logback
+          settings: logging-logback-settings
+    app-config:
+      app:
+        stellarNetworkPassphrase: Test SDF Network ; September 2015
```
jacekn, line 23: Should we template this? We could have TESTNET and PUBNET in the values file and derive passphrase from there.

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
...
+        circleUrl: {{ .Values.stellar.app_config.circle.circleUrl }}
+        apiKey: {{ .Values.stellar.app_config.circle.apiKey }}
+      payment-gateway:
+        circle:
+          name: "circle"
+          enabled: true
+        stellar:
+          enabled: false
+          name: "stellar"
+          horizonUrl: https://horizon-testnet.stellar.org
```
jacekn, line 69: We probably want to support pubnet as well?

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
...
+          enabled: true
+        stellar:
+          enabled: false
+          name: "stellar"
+          horizonUrl: https://horizon-testnet.stellar.org 
+          secretKey: secret # stellar account secret key
+      circle-payment-observer:
+        enabled: true
+        horizonUrl: https://horizon-testnet.stellar.org
+        stellarNetwork: TESTNET # TESTNET or PUBLIC
```
jacekn, line 74: TESTNET or PUBNET could probably be moved to the values file? And we could derive horizon URL from it.

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
...
+    data-spring-jdbc-h2:
+      spring.datasource.url: {{ .Values.stellar.app_config.data_spring_jdbc_h2.spring_datasource_url }}
+      spring.jpa.properties.hibernate.dialect: {{ .Values.stellar.app_config.data_spring_jdbc_h2.spring_jpa_properties_hibernate_dialect }}
+    spring:
+      logging:
+        level:
+          root: DEBUG
+          org.springframework: DEBUG
+          org.springframework.web.filter: DEBUG
+          org.stellar: DEBUG
```
jacekn, line 109: I'm not sure how verbose it is but maybe this should also be configurable via the values file?

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,135 @@
...
+    status = "test"
+    is_asset_anchored = true
+    anchor_asset_type = "fiat"
+    desc = "A test USDC issued by Circle."
+    [DOCUMENTATION]
+    ORG_NAME = "Stellar Development Foundation"
+    ORG_URL = "https://www.stellar.org"
+    ORG_DESCRIPTION = "Stellar is an open network for storing and moving money."
+    ORG_LOGO="https://assets-global.website-files.com/5deac75ecad2173c2ccccbc7/5dec8960504967fd31147f62_Stellar_lockup_black_RGB.svg"
+    ORG_SUPPORT_EMAIL="jamie@stellar.org"
```
jacekn, line 135: Maybe a generic email would be better?

helm-charts/sep-service/templates/ingress.yaml
```diff
@@ -0,0 +1,45 @@
...
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.fullName }}-ing-{{ .Values.service.name }}
+    helm.sh/chart: {{ $.Chart.Name }}-{{ $.Chart.Version }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}  
+spec:
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
```
jacekn, line 22: I believe `ingressClassName` is optional so maybe you can avoid setting it unless it's provided in the values file?

helm-charts/sep-service/templates/service.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - protocol: TCP
+    port: {{ .Values.service.servicePort }}       # port number the service will listen on
+    targetPort: {{ .Values.service.targetPort }}  # port number pods listen on
+  selector:
+    app.kubernetes.io/name: {{ .Values.service.name }}
```
jacekn, line 28: I think we also need to use `app.kubernetes.io/instance` to ensure multiple deployments are supported.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
+fullName: anchor-platform
+namespace: sandbox
```
jacekn, line 2: I wouldn't provide default here. Many charts use `namespaceOverride` and default to empty one. That may be the safest way to do it. WDYT?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+service:
+   containerPort: 8080
+   servicePort: 8080
+   replicas: 1
+   type: NodePort
+   name: sep
+image:
+   repo: stellar
+   name: anchor-platform
+   tag: 022afab
```
jacekn, line 12: How will consumers of this chart be handling software upgrades?

helm-charts/sep-service/templates/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ .Values.service.name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "common.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
```
jacekn, line 20: It would be good to handle custom annotations. For example `prometheus.io/scrape` is a common one that people will want to set to scrape metrics.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+   pullPolicy: Always
+ingress:
+   labels:
+     app: appy
+     test: label_test
+   annotations:
+     kubernetes.io/ingress.class: "public"
+     cert-manager.io/cluster-issuer: "default"
```
jacekn, line 20: Those can be deployment specific. I think it would be best to avoid setting them here as ingress class names might differ between k8s clusters. Also not every cluster will have cert-manager installed.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+      data-access:
+         type: data-spring-jdbc                      # Activate [config-spring-jdbc] module.
+         settings: data-spring-jdbc-sqlite           # The location of the configuration data in this file.
+      logging:
+         type: logging-logback
+         settings: logging-logback-settings
+   app_config:
+      app:
+         stellarNetworkPassphrase: Test SDF Network ; September 2015
+         hostUrl: https://www.stellaranchordemo.com
```
jacekn, line 53: This is the same host as in `ingress.tls.host`. Can we reuse it?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+         type: config-spring-property                # Activate [config-spring-property] module
+         settings: "app-config"                        # The location of the configuration data
+      data-access:
+         type: data-spring-jdbc                      # Activate [config-spring-jdbc] module.
+         settings: data-spring-jdbc-sqlite           # The location of the configuration data in this file.
+      logging:
+         type: logging-logback
+         settings: logging-logback-settings
+   app_config:
+      app:
```
jacekn, line 51: Structure here is quite complex and multi-level. Maybe it would make sense to remove the `stellar` level completely to reduce amount of indentation needed? We could go even further and just have things like `sep1` and `sep12` at the top level. WDYT?

Also I think some names might be confusing to some users. For example we have `stellar.anchor.app-config` but also `stellar.app_config`. Maybe it's possible to rename one of them to clarify what they mean?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+         stellarNetworkPassphrase: Test SDF Network ; September 2015
+         hostUrl: https://www.stellaranchordemo.com
+         languages: en
+         horizonUrl: https://horizon-testnet.stellar.org
+         assets: assets-test.json
+         jwtSecretKey: secret
+      sep1:
+         enabled: true
+         stellarFile: file:/config/stellar-wks.toml
```
jacekn, line 60: Are users mean to be configuring this parameter at all? I suspect not and if so then maybe remove it completely from the values file to make it shorter?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+      circle-payment-observer:
+         enabled: true
+         horizonUrl: https://horizon-testnet.stellar.org
+         stellarNetwork: TESTNET # TESTNET or PUBLIC
+         trackedWallet: all
+      event:
+         enabled: true
+         publisherType: kafka
+      kafka_publisher:
+         bootstrapServer: anchor-platform-kafka:9092
```
jacekn, line 105: How will kafka be deployed?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+      data_spring_jdbc_h2:
+         sprint_datasource_url: jdbc:h2:mem:test
+         spring_jpa_properties_hibernate_dialect: org.hibernate.dialect.H2Dialect
+      spring:
+      logging:
+         level:
+            root: DEBUG
+            org.springframework: DEBUG
+            org.springframework.web.filter: DEBUG
+            org.stellar: DEBUG
```
jacekn, line 132: You don't actually consume those values in the config map.

jacekn (commented):
Great work! I added a few comments and questions inline.

One high level think that I'd like to mention is amount of settings in the values file. It's not a problem at all if they are needed but maybe not all of them are? Are there any settings that will never be changed? If so we could remove them from the values file to make it simpler.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,134 @@
...
+            enabled: true
+         stellar:
+            enabled: false
+            name: "stellar"
+            horizonUrl: https://horizon-testnet.stellar.org
+            secretKey: secret # stellar account secret key
+      circle-payment-observer:
+         enabled: true
+         horizonUrl: https://horizon-testnet.stellar.org
+         stellarNetwork: TESTNET # TESTNET or PUBLIC
```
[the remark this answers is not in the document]
reecexlm: Yes, thanks for that feedback. Going to make a pass and leverage default values across most of the configuration and check for redundant references to the same values.  Also will incorporate your suggestion about some logic about default settings based on stellarNetwork (testnet or public) working on this now.

helm-charts/sep-service/values.yaml
reecexlm: it expects kafka to be deployed and provide the bootstrap broker (will soon be a list) as config parameter.   that being said, we will be providing some terraform example code that sets up msk along with anchor platform.

helm-charts/sep-service/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Sep Helm chart 
+name: sep
+version: 0.3.81
```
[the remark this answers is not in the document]
reecexlm: i will move it to 0.1.0 once we publish to our official repository if that works?

helm-charts/sep-service/values.yaml
reecexlm: I think I will do this but will delay this refactor after initial merge if thats ok and once we have it into our test pipeline.

helm-charts/sep-service/values.yaml
reecexlm: it is configurable now but with default value.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+        signingSeed: ${SEP10_SIGNING_SEED}
+      sep12:
+        enabled: true
+        customerIntegrationEndpoint: https://anchor-reference-server-dev.stellar.org
+      sep31:
+        enabled: true
+        feeIntegrationEndPoint: https://anchor-reference-server-dev.stellar.org
+      sep38:
+        enabled: true
+        quoteIntegrationEndPoint: https://anchor-reference-server-dev.stellar.org
```
jacekn, line 89: Those 3 can probably come from a single variable with option to override per sep. WDYT?

helm-charts/sep-service/templates/ingress.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - host: {{ tpl .host $ }}
+    http:
+      paths:
+        - path: {{ .path | quote }}
+          pathType: {{ .pathType| quote }}
+          backend:
+            service:
+              name: {{ tpl .backend.serviceName $ }}
+              port:
+                number: {{ tpl .backend.servicePort $ }}
```
jacekn, line 43: Do we allow port to be configurable? Is there any need for that?

helm-charts/sep-service/templates/service.yaml
```diff
@@ -0,0 +1,28 @@
...
```
jacekn, line 28: This might cause clashes with other deployments of the same chart. I think you need to use more labels to ensure you only match pods from a single deployment. `app.kubernetes.io/instance:` might be enough but it's worth checking

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,94 @@
+fullName: anchor-platform
+namespace: sandbox
+service:
+   containerPort: 8080
+   servicePort: 8080
```
jacekn, line 5: Maybe those don't have to be configurable? Those port numbers are more of an internal implementation detail. I think it might be OK to just use 8080 without even adding option to override in the values file.
But of course if there is a use case for changing ports we should keep those settings

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+   replicas: 1
```
jacekn, line 6: Will this setting apply to service? Or deployment?

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+   type: NodePort
```
jacekn, line 7: NodePort might not work in all deployments. In fact if you use ingress ClusterIP is better.

helm-charts/sep-service/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+     kubernetes.io/ingress.class: "public"
+     cert-manager.io/cluster-issuer: "default"
+     ingress.kubernetes.io/ssl-redirect: "true"
+     ingress.kubernetes.io/browser-xss-filter: "true"
+     ingress.kubernetes.io/frame-deny: "true"
+     ingress.kubernetes.io/content-type-nosniff: "true"
+     ingress.kubernetes.io/hsts-max-age: "31536000"
+     ingress.kubernetes.io/hsts-include-subdomains: "true"
+   tls:
+     host: "reece.sandbox.kube001.services.stellar-ops.com"
```
jacekn, line 32: This exposes some SDF internal sauce. It would be good to change this to more generic default and it make it clear that it needs to be configured

jacekn (commented):
Thank you, it looks much much cleaner now!
I dropped a few more comments inline

helm-charts/sep-service/readme.md
```diff
@@ -0,0 +1,105 @@
...
+## Helm Chart Stellar Anchor Platform Configuration
+The following table lists the additional configurable parameters of the Anchor Platform chart and their default values.
+|  Parameter | Description | Required?  | Default Value | 
+|---|---|---|---|
+| stellar.anchor.data_access.type  | database access type  | yes  | data-spring-jdbc  |
+| stellar.anchor.data_access.setttings  | values config root-level key for jdbc config  | yes | data-spring-jdbc-sqlite  |
+| stellar.app_config.app.hostUrl  | URL of the Anchor Platform SEP Service   | y  | n/a |
+| stellar.app_config.app.jwtSecretKey | web encryption key | ${JWT_SECRET_KEY} |
+| sep1.enabled | sep1 true if service enabled | yes  | true  |
+| sep1.stellarFile | location/name of stellar.toml file on service pod  | n | file:/config/stellar-wks.toml  |
```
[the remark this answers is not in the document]
reecexlm: ack, i will remove it for now

helm-charts/sep-service/templates/configmap.yaml
```diff
@@ -0,0 +1,241 @@
...
+          settings: logging-logback-settings
+        {{- end }}
+    app-config:
+      app:
+        stellarNetwork: {{ .Values.stellar.app_config.app.stellarNetwork | default "TESTNET" }}
+      {{- if eq .Values.stellar.app_config.app.stellarNetwork "PUBNET" }} 
+        stellarNetworkPassphrase: Public Global Stellar Network ; September 2015 
+        horizonUrl: https://horizon.stellar.org
+      {{- else }} 
+        stellarNetworkPassphrase: {{ .Values.stellar.app_config.app.stellarNetworkPassphrase | default "Test SDF Network ; September 2015" }}
```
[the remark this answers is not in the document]
reecexlm: oh yea.  thank you. actually meant to do that. cheers for catching this.

helm-charts/sep-service/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+    {{- range $key, $value := .Values.deployment.annotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
+  labels:
+    app.kubernetes.io/name: {{ include "common.name" . }}
+    helm.sh/chart: {{ include "common.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
```
javijvj, line 17: can you please allow set the `serviceAccountName` and `automountServiceAccountToken` to be able to manage the pod permissions in a simpler way. [doc](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)

helm-charts/sep-service/templates/rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+      name: {{ .Release.Name }}-role
+      namespace: {{ .Release.Namespace }}
+      {{- if (.Values.rbac).annotations }}
+      annotations:
+        {{- with .Values.rbac.annotations }}
+{{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- end }}
+    rules:
+{{ toYaml .Values.rbac.rules | indent 6 }}
```
jacekn, line 34: I don't see any defaults in the values file. Could you document what RBAC/permissions does the anchor SDK require to function properly?

This change also has potentially significant security impact. Access to create role bindings is effectively privilege escalation and in secure environments and it would be better to avoid this.

Maybe the same functionality can be achieved without granting extra access?

jacekn (approved):
As per slack feel free to merge. There are some open comments but due to the size of this PR it might be nicer to work on improvements under separate PRs.

PR #362, opened 2022-06-29
core: Add customer type to sender/receiver validation methods and delete calls.

erika-sdf:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add customer type to sender/receiver validation methods and delete calls.

### Why

Customer type should be specified for these calls, as specified in SEP12.

### Known limitations

This is meant to be a short term-solution to resolve integration issues. TODOs have been added to address long term concerns (eg, choosing the first customer type in the sender/receiver map)

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
@@ -68,23 +68,27 @@ public class Sep12Service {
           String.format("Not authorized to delete account [%s]", account));
     }
 
-    String senderTypes[] = {"sep31-sender", "sep31-receiver"};
-    for (String senderType: senderTypes) {
+    // TODO: Move this into configuration instead of hardcoding customer type values.
+    boolean existingCustomerMatch = false;
+    String customerTypes[] = {"sep31-sender", "sep31-receiver"};
+    for (String customerType: customerTypes) {
       Sep12GetCustomerResponse existingCustomer =
           customerIntegration.getCustomer(
               Sep12GetCustomerRequest.builder()
                   .account(account)
                   .memo(memo)
                   .memoType(memoType)
-                  .type(senderType)
+                  .type(customerType)
                   .build());
-      if (existingCustomer.getId() == null) {
-        infoF(
-            "No existing customer found for account={} memo={} memoType={}", account, memo, memoType);
-        throw new SepNotFoundException("User not found.");
+      if (existingCustomer.getId() != null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -68,20 +68,24 @@ public void deleteCustomer(JwtToken jwtToken, String account, String memo, Strin
...
-                .account(account)
-                .memo(memo)
-                .memoType(memoType)
-                .build());
-    if (existingCustomer.getId() == null) {
-      infoF(
-          "No existing customer found for account={} memo={} memoType={}", account, memo, memoType);
-      throw new SepNotFoundException("User not found.");
-    }
+    String senderTypes[] = {"sep31-sender", "sep31-receiver"};
```
marcelosalloum, line 71:
```suggestion
    String customerTypes[] = {"sep31-sender", "sep31-receiver"};
```

marcelosalloum: Maybe for another PR, move it to a config file.

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -68,20 +68,24 @@ public void deleteCustomer(JwtToken jwtToken, String account, String memo, Strin
...
+    for (String senderType: senderTypes) {
+      Sep12GetCustomerResponse existingCustomer =
+          customerIntegration.getCustomer(
+              Sep12GetCustomerRequest.builder()
+                  .account(account)
+                  .memo(memo)
+                  .memoType(memoType)
+                  .type(senderType)
+                  .build());
+      if (existingCustomer.getId() == null) {
```
marcelosalloum, line 81: !!  You should only throw a 404 not found if none of the requests in the for loop succeed.

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
erika-sdf: Acknowledged, added a TODO for this

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
erika-sdf: Good catch, I'll alter the logic to only error if no match is found.

marcelosalloum (approved):
LGTM!

PR #367, opened 2022-07-01
Let /customer DELETE calls expect 204 on success instead of 200

erika-sdf:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Let '/customer' DELETE calls expect 204 on success, instead of 200.

### Why

This is in line with the OpenAPI spec for anchor platform

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java
@@ -111,7 +111,8 @@ public class RestCustomerIntegration implements CustomerIntegration {
     Response response = call(httpClient, callbackRequest);
     String responseContent = getContent(response);
 
-    if (response.code() != HttpStatus.NO_CONTENT.value()) {
+    if (response.code() != HttpStatus.NO_CONTENT.value()
+        && response.code() != HttpStatus.OK.value()) {
       throw httpError(responseContent, response.code(), gson);
     }
   }
--- a/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java
@@ -111,7 +111,7 @@ public class RestCustomerIntegration implements CustomerIntegration {
     Response response = call(httpClient, callbackRequest);
     String responseContent = getContent(response);
 
-    if (response.code() != HttpStatus.OK.value()) {
+    if (response.code() != HttpStatus.NO_CONTENT.value()) {
       throw httpError(responseContent, response.code(), gson);
     }
   }


JakeUrban:
Can we remove the node modules directory from the PR? 3k files is a lot to review :-)

erika-sdf:
Oops, definitely didn't mean to include those. Removed them

erika-sdf:
Depends on a change to the anchor tests: https://github.com/stellar/stellar-anchor-tests/pull/92/files

marcelosalloum (approved):

marcelosalloum (changes requested):
I'm dismissing the approval because I don't think this change was made in the correct place.

In the [slack message](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#delete-responses), our partners says the problem is the message coming from Anchor->Platform, but you changed the SEP-12 response, wich is a different thing.

marcelosalloum:
The SEP-12 response should be 200, instead of 204 ([ref](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#delete-responses))

Our partner's Anchor is returning 204 from the callbackAPI, which is the value expected in the [documentation](https://github.com/stellar/java-stellar-anchor-sdk/blob/0cbf8cd20f8e17b8f7dd97c1831765845a6f4942/docs/03%20-%20Implementing%20the%20Anchor%20Server/Communication/Callbacks%20API.yml#L419-L420), but the Platform seems to be mistakenly waiting for a 200.

marcelosalloum:
The mistake seems to be here:
https://github.com/stellar/java-stellar-anchor-sdk/blob/0cbf8cd20f8e17b8f7dd97c1831765845a6f4942/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java#L114-L116

According with the documentation, both a 204 or a 404 could be returned from the Anchor.

erika-sdf:
> The mistake seems to be here:
> 
> https://github.com/stellar/java-stellar-anchor-sdk/blob/0cbf8cd20f8e17b8f7dd97c1831765845a6f4942/platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java#L114-L116
> 
> According with the documentation, both a 204 or a 404 could be returned from the Anchor.

That makes sense, thanks for catching this.

In that case, I will revert the anchor test change as well: https://github.com/stellar/stellar-anchor-tests/pull/93

platform/src/main/java/org/stellar/anchor/platform/callback/RestCustomerIntegration.java
```diff
@@ -111,7 +111,7 @@ public void deleteCustomer(String id) {
     Response response = call(httpClient, callbackRequest);
     String responseContent = getContent(response);
 
-    if (response.code() != HttpStatus.OK.value()) {
+    if (response.code() != HttpStatus.NO_CONTENT.value()) {
```
marcelosalloum, line 114: 💡  Maybe support both, just in case? I mean, if we get a 2xx, that's usually interpreted as a successful response.

marcelosalloum (approved):

Issue #370, opened 2022-07-05
Kafka Event Serialization


reecexlm:
during  end 2 end tests in our side between the Stellar Platform and our Stellar Anchor Service. We found an issue between the integration of Kafka events. We're expecting the events coming with this format https://github.com/stellar/stellar-anchor-platform/blob/main/design/eventSchema/event.yml (snake_case for the fields right?). But we're seeing in the logs of our service, events coming in incorrectly.

PR #374, opened 2022-07-06
use snake case for kafka events

stfung77:
Tested both Kafka and SQS events

originally fields only had the @SerializedName annotation which wasnt used by the kafka serializer. added @JsonProperty to event fields to fix this. 

@JsonProperty is for Kafka since it uses a JsonSerializer
@SerializedName is for SQS (and potentially jdbc in the future if we decided to save events to the database if Event sending isnt possible) since it uses gson for serialization

this addresses - https://github.com/stellar/java-stellar-anchor-sdk/issues/370


JakeUrban (approved):

spolaore:
@stfung77, another interesting option is that in spring boot apps, it is also possible to change jackson serialization behavior at app level by setting the property `spring.jackson.property-naming-strategy=SNAKE_CASE`.

Issue #365, opened 2022-07-01
SEP-31: add sending anchor allowlist


JakeUrban:
### What problem does your feature solve?

The current Java Anchor SDK / Anchor Platform code does not have any logic to deny an unapproved SEP-31 sending anchor.

[`app-config.sep10.omnibusAccountList`](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/platform/src/main/resources/anchor-config-defaults.yaml#L129) is currently only used in relation to SEP-10 client attribution, and this SEP-10 feature is only utilized when Stellar accounts being authenticated are owned by individual users, not the applications / businesses themselves.

### What would you like to see?

This is up for debate. 

#### Option 1

One option is to add a configuration option that allows anchors to specify whether or not an allowlist should be checked whenever a Stellar account is authenticating via SEP-10. For anchors that only interact with applications that use omnibus accounts, this would be sufficient. 

`omnibusAccountList` could be checked if `requireKnownOmnibusAccount` (name TBD) is true, and clients would be rejected or accepted based on their public key's presence in the list. If `requireKnownOmnibusAccount` is false, the existing behavior would be used.

#### Option 2

The problem with Option 1 is that is doesn't account for the case where an anchor wants to restrict access to known SEP-31 senders but also allow access to noncustodial SEP-24 & 6 wallets. Clients do not specify the SEPs they intend to engage with when authenticating via SEP-10, and it would be a breaking change to start requiring that information.

Polaris deals with this by only restricting access to SEP-31 endpoints (excluding `GET /info`). Unknown accounts are still able to register customers via SEP-12 & create SEP-38 quotes, which are both actions necessary in SEP-6 flows.

We don't know if this would acceptable to anchors who only implement SEP-31, and the safe bet is to assume it would not be.

#### Option 3

We do both.

Add `requireKnownOmnibusAccount` to determine if `omnibusAccountList` should be used to gate access to all services. If `requireKnownOmnibusAccount` is false, the existing SEP-10 functionality implemented should be used. However, we can have another config option `knownSendingAnchorPublicKeyList` (name TBD), that is used to protect SEP-31 endpoints. 

### What alternatives are there?

I'm open to suggestions.

marcelosalloum:
[...]

PR #378, opened 2022-07-08
Add Sep10 Allowlist

stfung77:
added config `requireKnownOmnibusAccount` to sep10 config. if enabled, only allow accounts in the `omnibusAccountList` to authenticate with SEP-10. 

addresses: https://github.com/stellar/java-stellar-anchor-sdk/issues/365


core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
```diff
@@ -562,6 +562,33 @@ internal class Sep10ServiceTest {
     assertThrows<SepException> { sep10Service.createChallenge(cr) }
   }
 
+  @Test
+  fun testRequireKnownObnibusAccount() {
```
marcelosalloum, line 566: The typo I mentioned before is also in all these tests names. I'm very guilty of falling into this same copy/paste error 😅.

core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
```diff
@@ -562,6 +562,33 @@ internal class Sep10ServiceTest {
...
+    every { sep10Config.isRequireKnownOmnibusAccount } returns true
+    every { sep10Config.omnibusAccountList } returns listOf(TEST_ACCOUNT)
+    val cr = ChallengeRequest.of(TEST_ACCOUNT, TEST_MEMO, TEST_HOME_DOMAIN, null)
+
+    assertDoesNotThrow{ sep10Service.createChallenge(cr) }
```
marcelosalloum, line 571: 💡  The `Mockk` library has the `verify` function that can be used to verify that your mocks were called. We use that in some parts of our project, like in:
https://github.com/stellar/java-stellar-anchor-sdk/blob/ff95819513c0d381dc3322709ab5e4abcc316790/core/src/test/kotlin/org/stellar/anchor/sep31/Sep31ServiceTest.kt#L568-L571

This helps, among other things, to make sure you're not mocking anything unnecessarily, or that your mocked functions are being called with the correct parameters.

core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
```diff
@@ -562,6 +562,33 @@ internal class Sep10ServiceTest {
...
+  }
+
+  @Test
+  fun testRequireKnownObnibusAccountUnknownAccount() {
+    every { sep10Config.isRequireKnownOmnibusAccount } returns true
+    val cr = ChallengeRequest.of(TEST_ACCOUNT, TEST_MEMO, TEST_HOME_DOMAIN, null)
+    every { sep10Config.omnibusAccountList } returns listOf("G321E23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP")
+
+    assertThrows<SepException> { sep10Service.createChallenge(cr) }
```
marcelosalloum, line 589: 💡  Can you improve on this test to verify if the error thrown is what we expect? We do that in other parts of the project, like in:
https://github.com/stellar/java-stellar-anchor-sdk/blob/ff95819513c0d381dc3322709ab5e4abcc316790/platform/src/test/kotlin/org/stellar/anchor/platform/PaymentConfigTest.kt#L21-L26

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -65,6 +65,12 @@ public ChallengeResponse createChallenge(ChallengeRequest challengeRequest) thro
           sep10Config.getOmnibusAccountList().contains(challengeRequest.getAccount().trim());
     }
 
+    if (sep10Config.isRequireKnownOmnibusAccount() && !omnibusWallet){
+      // validate that requesting account is allowed access
+      infoF("requesting account: {} is not in allow list", challengeRequest.getAccount().trim());
```
marcelosalloum, line 70: 🤔 Since we're returning an error here, what about logging with `errorF` instead of `infoF`? This could make it easier to debug when the authentication fails?

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -15,4 +15,5 @@ public class PropertySep10Config implements Sep10Config {
   private List<String> clientAttributionDenyList;
   private List<String> clientAttributionAllowList;
   private List<String> omnibusAccountList;
+  private boolean requireKnownOmnibusAccount = true;
```
marcelosalloum, line 18: Can you add validation for this new parameter here at `PropertySep10Config`? If requireKnownOmnibusAccount is true and omnibusAccountList is empty, we should throw an error to stop the boot because that configuration would prevent any SEP-10 authentication from passing.

We're already doing this kind of validation in [PropertyIntegrationAuthConfig.java#L12](https://github.com/stellar/java-stellar-anchor-sdk/blob/ff95819513c0d381dc3322709ab5e4abcc316790/platform/src/main/java/org/stellar/anchor/platform/config/PropertyIntegrationAuthConfig.java#L12). You basically need to implement the `Validator` interface and its methods.

marcelosalloum (commented):
Cool, thanks for preparing that! I've added a few suggestions below for the sake of completeness.

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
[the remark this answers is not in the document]
stfung77: idea behind this is `secure by default`. also, i believe omnibusAccountList is empty on our deployments so this shouldnt break anything. please let me know if you feel differently.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
stfung77: i think we should stick with infoF here since the `error` thrown isn't is system error but rather invalid user input. this also messes with metrics and makes it seem like theres an issue with the system rather than user error

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
stfung77: yeah, i'll add the validation. 
as for an empty omnibusAccountList, i think that should be valid, eg: an anchor wants to block access to the Anchor Platform. thoughts on this use case?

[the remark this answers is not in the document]
reecexlm: Yea, I would communicate with Bitso. Also they use helm chart so we should make sure the pipeline is updated with helm before we land this in main.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -65,6 +65,12 @@ public ChallengeResponse createChallenge(ChallengeRequest challengeRequest) thro
           sep10Config.getOmnibusAccountList().contains(challengeRequest.getAccount().trim());
     }
 
+    if (sep10Config.isRequireKnownOmnibusAccount() && !omnibusWallet) {
+      // validate that requesting account is allowed access
+      infoF("requesting account: {} is not in allow list", challengeRequest.getAccount().trim());
+      throw new SepValidationException("unable to process");
```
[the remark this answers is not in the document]
stfung77: agreed. added this for client attribution allow/deny list as well

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
stfung77: yup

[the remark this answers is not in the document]
lijamie98: AFAIU, `requireKnownOmnibusAccount` would not make it secure :-). It is an option to enforce `omnibusAccountList`. This will make non-custodial wallet to fail. 

IMO, if we want to make Anchor Platform to work with non-custodial wallet, we should make the default to `false`. But if we want to make Anchor Platform to work with omnibus wallet only, then we set default to 'true'.

lijamie98:
@stfung77 I think the `requireKnownOmnibusAccount` should be set to `false` to avoid breaking non-custodial wallet SEP-6 and SEP-24 implementations.

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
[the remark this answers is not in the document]
stfung77: anchor-config-defaults.yaml has the value set to false and that will work as the default value now. removed the java object default value

lijamie98 (approved):

platform/src/main/java/org/stellar/anchor/platform/controller/Sep10Controller.java
```diff
@@ -88,6 +90,15 @@ public SepExceptionResponse handleSepValidationException(Exception ex) {
...
   }
 
+  @ExceptionHandler({
+    SepNotAuthorizedException.class,
+  })
+  @ResponseStatus(value = HttpStatus.FORBIDDEN)
+  public SepAuthorizationExceptionResponse handleSepAuthorizationException(Exception ex) {
+    errorEx(ex);
+    return new SepAuthorizationExceptionResponse(ex.getMessage());
+  }
```
marcelosalloum, line 100: This piece of code seems redundant with what we already had in the Global error handler:

https://github.com/stellar/java-stellar-anchor-sdk/blob/8cd8d07a4ff8ce202158c4c064cde542bbc4acc2/platform/src/main/java/org/stellar/anchor/platform/controller/GlobalControllerExceptionHandler.java#L33-L38

api-schema/src/main/java/org/stellar/anchor/api/sep/SepAuthorizationExceptionResponse.java
```diff
@@ -0,0 +1,12 @@
...
+import lombok.Data;
+
+@Data
+public class SepAuthorizationExceptionResponse {
+  String error;
+
+  public SepAuthorizationExceptionResponse(String error) {
+    this.error = error;
+  }
+}
```
marcelosalloum, line 12: I don't see any reason for adding this class, since it's the same as `SepExceptionResponse`:
https://github.com/stellar/java-stellar-anchor-sdk/blob/8cd8d07a4ff8ce202158c4c064cde542bbc4acc2/api-schema/src/main/java/org/stellar/anchor/api/sep/SepExceptionResponse.java#L5-L12

marcelosalloum (commented):
Very cool!

@stfung77: I've found a couple of pieces of code that are not necessary. Do you mind opening a new PR to do the clean up?

platform/src/main/java/org/stellar/anchor/platform/controller/Sep10Controller.java
stfung77: i thought we agreed to return 403 errors for those instead of a generic 400?

marcelosalloum: Oh, my mistake. I highlighted the wrong piece of code in my comment. Here is what I should have highlighted: 

https://github.com/stellar/java-stellar-anchor-sdk/blob/8cd8d07a4ff8ce202158c4c064cde542bbc4acc2/platform/src/main/java/org/stellar/anchor/platform/controller/GlobalControllerExceptionHandler.java#L40-L45

If you look at the Global exception handler ☝️ , it does the same thing as the structure you added:
1. Intercept `SepNotAuthorizedException`.
2. Use `HttpStatus.FORBIDDEN` in the response status code.
3. The response body will be the same: 
  ```json
  {
    "error": "{originalExceptionMessage}"
  }
  ```

marcelosalloum: So my point is that you added a new exception handler for `SepNotAuthorizedException` in Sep10Controller that overrides the default exception handler in GlobalControllerExceptionHandler and does the exact same thing as the default exception handler, so you're basically adding some code that's not necessary.

PR #381, opened 2022-07-11
platform: Add log4j2 json appender as an appender option for anchor platform server

erika-sdf:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add Splunk-friendly log4j2 json appender as an option for anchor platform server

### Why

Json logs are query-able, and ingest-able by Splunk.

### Limitations

The format specified here follows the [Splunk documentation](https://docs.splunk.com/Documentation/Splunk/9.0.0/Data/FormateventsforHTTPEventCollector).


marcelosalloum:
Is there a way to test this configuration? Did you test it manually or do you intend to test it after merged?

marcelosalloum:
> I tested it manually, but I will add a unit test

Cool, thank you!

Also, considering this is related to a config file that is not mentioned in the documentation yet, could you plz add a section to [Running & Configuring the Application](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/docs/01%20-%20Running%20%26%20Configuring%20the%20Application/A%20-%20Running%20%26%20Configuring%20the%20Application.md) explaining how to configure the logs through this new `LOG_APPENDER` env var?

core/src/main/java/org/stellar/anchor/util/JsonLog.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class JsonLog {
+  @Getter @Setter Integer time;
+
+  @Getter @Setter String source;
+
+  @Getter @Setter String index;
+
+  @Getter @Setter EventObjectJson event;
+}
```
marcelosalloum, line 14: Since you're only using `JsonLog` (and `EventObjectJson`) inside the tests, maybe it's better to declare it in the test itself?

Actually, you could delete these classes and use a Map to parse the json and test its values, similar to what we do here:
https://github.com/stellar/java-stellar-anchor-sdk/blob/f82985c0ce81df8aeee6fd3c95db30a224a7e69e/integration-tests/src/main/kotlin/org/stellar/anchor/platform/Sep31Client.kt#L19-L20

In general, I don't think we should add new models to the project `src` if they are only meant to be used for tests.

platform/src/test/kotlin/org/stellar/anchor/platform/LogAppenderTest.kt
```diff
@@ -0,0 +1,24 @@
...
+  fun test_logJsonAppenderFormat() {
+    val outputStreamCaptor = ByteArrayOutputStream()
+    System.setOut(PrintStream(outputStreamCaptor))
+    Log.info("hello world")
+
+    val json = outputStreamCaptor.toString().trim()
+    val gson = Gson()
+
+    assertDoesNotThrow { gson.fromJson(json, JsonLog::class.java) }
+    val actualLog = gson.fromJson(json, JsonLog::class.java)
```
marcelosalloum, line 20: You can use optionals to avoid running the same command twice, as we do here:
https://github.com/stellar/java-stellar-anchor-sdk/blob/f82985c0ce81df8aeee6fd3c95db30a224a7e69e/core/src/test/kotlin/org/stellar/anchor/sep38/Sep38ServiceTest.kt#L304-L307

marcelosalloum (approved):
LGTM! I've left a few suggestions below to make the PR a leaner.

In particular, I don't think we should add the new models JsonLog and EventObjectJson to the source, considering they are only being used for tests. I left a suggestion in one of the comments below.

Feel free to merge it after addressing the comments 🙏.

marcelosalloum (approved):

erika-sdf:
Thanks for the tips. Added the changes, and will merge.

PR #412, opened 2022-07-21
SEP-31: improve documentation on the TransactionEvent fields and fix events values in the code

### LINKED ISSUE #387: transaction-status-changed event doesn't include customers.receiver.memo

### LINKED ISSUE #388: transaction-status-changed event sourceAccount and destioationAccount filled incorrectly
On `transaction-created` event, under `sourceAccount` and `destinationAccount` , the customers Anchor's UUID from `POST /transaction` are being sent:
https://github.com/stellar/java-stellar-anchor-sdk/blob/137b200c886345a0289b7eb6fe43e78e10452a66/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java#L176-L177

However, during `transaction-status-changed`, those fields are being populated with stellar accounts from the observed blockchain payment:
https://github.com/stellar/java-stellar-anchor-sdk/blob/0ab3db136ccf256294b05082271b23559844be33/platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java#L193-L194

--

> This is a blocker,. for now a temporary workaround this issue by ignoring the source and destination account coming in transaction-status-changed and use the UUIDs we got from the first transaction-created.

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Improve documentation on the TransactionEvent fields and fix events values in the code. The following properties were off:
* `TransactionEvent.source_account`
* `TransactionEvent.destination_account`
* `TransactionEvent.customers.sender`
* `TransactionEvent.customers.receiver`
* `TransactionEvent.creator`

### Why

Close #387
Close #388

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
@@ -113,6 +113,14 @@ public class Sep31Service {
     // Query the fee
     updateFee();
 
+    // Get the creator's stellarId
+    StellarId creatorStellarId =
+        StellarId.builder()
+            .account(Objects.requireNonNullElse(jwtToken.getMuxedAccount(), jwtToken.getAccount()))
+            .memo(jwtToken.getAccountMemo())
+            .memoType(jwtToken.getAccountMemo() == null ? null : "id")
+            .build();
+
     AssetInfo asset = Context.get().getAsset();
     Amount fee = Context.get().getFee();
     Sep31Transaction txn =
@@ -135,6 +143,7 @@ public class Sep31Service {
             .refunds(null)
             .senderId(Context.get().getRequest().getSenderId())
             .receiverId(Context.get().getRequest().getReceiverId())
+            .creator(creatorStellarId)
             // updateAmounts will update these ⬇️
             .amountIn(request.getAmount())
             .amountInAsset(assetInfo.getAssetName())
@@ -177,12 +186,7 @@ public class Sep31Service {
             .custodialTransactionId(null)
             .sourceAccount(request.getSenderId())
             .destinationAccount(request.getReceiverId())
-            .creator(
-                StellarId.builder()
-                    .account(txn.getStellarAccountId())
-                    .memo(txn.getStellarMemo())
-                    .memoType(txn.getStellarMemoType())
-                    .build())
+            .creator(creatorStellarId)
             .build();
     eventService.publish(event);
 
--- a/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
@@ -192,7 +192,7 @@ public class Sep31Service {
 
     return Sep31PostTransactionResponse.builder()
         .id(txn.getId())
-        .stellarAccountId(Context.get().getAsset().getDistributionAccount())
+        .stellarAccountId(txn.getStellarAccountId())
         .stellarMemo(txn.getStellarMemo())
         .stellarMemoType(txn.getStellarMemoType())
         .build();
--- a/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
@@ -32,8 +32,10 @@ import org.stellar.anchor.auth.JwtToken;
 import org.stellar.anchor.config.AppConfig;
 import org.stellar.anchor.config.Sep31Config;
 import org.stellar.anchor.event.EventPublishService;
+import org.stellar.anchor.event.models.Customers;
 import org.stellar.anchor.event.models.StellarId;
 import org.stellar.anchor.event.models.TransactionEvent;
+import org.stellar.anchor.sep12.Sep12CustomerStore;
 import org.stellar.anchor.sep38.Sep38Quote;
 import org.stellar.anchor.sep38.Sep38QuoteStore;
 import org.stellar.anchor.util.Log;
@@ -41,6 +43,7 @@ import org.stellar.anchor.util.Log;
 public class Sep31Service {
   private final AppConfig appConfig;
   private final Sep31Config sep31Config;
+  private final Sep12CustomerStore sep12CustomerStore;
   private final Sep31TransactionStore sep31TransactionStore;
   private final Sep31DepositInfoGenerator sep31DepositInfoGenerator;
   private final Sep38QuoteStore sep38QuoteStore;
@@ -53,6 +56,7 @@ public class Sep31Service {
   public Sep31Service(
       AppConfig appConfig,
       Sep31Config sep31Config,
+      Sep12CustomerStore sep12CustomerStore,
       Sep31TransactionStore sep31TransactionStore,
       Sep31DepositInfoGenerator sep31DepositInfoGenerator,
       Sep38QuoteStore sep38QuoteStore,
@@ -64,6 +68,7 @@ public class Sep31Service {
     debug("sep31Config:", sep31Config);
     this.appConfig = appConfig;
     this.sep31Config = sep31Config;
+    this.sep12CustomerStore = sep12CustomerStore;
     this.sep31TransactionStore = sep31TransactionStore;
     this.sep31DepositInfoGenerator = sep31DepositInfoGenerator;
     this.sep38QuoteStore = sep38QuoteStore;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -151,6 +165,8 @@ public Sep31PostTransactionResponse postTransaction(
     updateDepositInfo(txn);
     sep31TransactionStore.save(txn);
 
+    StellarId senderStellarId = sep12CustomerStore.findById(txn.getSenderId()).toStellarId();
```
lijamie98, line 168: This can be done by calling [`CustomerIntegration.getCustomer()`](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/api-schema/src/main/java/org/stellar/anchor/api/callback/CustomerIntegration.java#L17)

lijamie98: 🤔 Can we consider not to store the customer information in the platform database to avoid multiple sources of truth?

marcelosalloum: After updating the description of the fields in [Events Schema.yml](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/docs/03%20-%20Implementing%20the%20Anchor%20Server/Communication/Events%20Schema.yml), seems like the database call wasn't needed.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -97,29 +97,36 @@ components:
...
           enum: ["receive"]
         status:
           type: string
+          description: |
+            The status of the transaction according with the latest Platform update. Currently, we're supporting all [SEP-31 states](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#get-transaction).
           enum: [
             "pending_sender",
             "pending_stellar",
             "pending_customer_info_update",
+            "pending_transaction_info_update",
```
JakeUrban, line 125: I didn't add this because the platform doesn't support custom per-transaction fields, and the only time `pending_transaction_info_update` is used is when those fields need updating.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -156,16 +167,18 @@ components:
             $ref: '#/components/schemas/StellarTransaction'
         external_transaction_id:
           type: string
+          description: "The id of an external transaction to be handled by the Anchor Backend."
         custodial_transaction_id:
           type: string
         source_account:
           type: string
-          description: The SEP-12 user ID of the source account.
+          description: "The Stellar account who sent the payment that originated this event. Should be present if this event was originated from a Stellar payment."
```
JakeUrban, line 175: This description could be confused with a transaction's source account, which could be different than the source of funds. Can we make it clear that this is the account whose balance was debited?

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -156,16 +167,18 @@ components:
...
         destination_account:
           type: string
-          description: The SEP-12 user ID of the destination account.
+          description: "The Stellar account who received the payment that originated this event. Should be present if this event was originated from a Stellar payment."
         customers:
           type: object
+          description: "This field should contain the `id` or the **deprecated** `account+memo` identification pair of the sending and receiving customers created through [SEP-12](https://stellar.org/protocol/sep-12)."
```
JakeUrban, line 181: Why is the account+memo deprecated? These fields are sent in the SEP-12 requests which get forwarded to the anchor, so why are we deprecating them here?

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -219,9 +232,13 @@ components:
     StellarId:
       type: object
       description:
-        StellarId's are objects that identify an on-chain entity. Both end users and SEP-31 senders can be identified
-        with StellarId's. Id's for SEP-31 senders will never include a `memo` or `memo_type`.
+        StellarId's are objects that identify a SEP entity. Both end-users and SEP-31 senders/receivers can be identified
```
JakeUrban, line 235: I'm not sure "SEP Entity" is any more clear than "on-chain entity". Maybe we should just be explicit and say "Stellar ID's are objects that identify end-users and SEP-31 senders (they won't identify receivers).

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -219,9 +232,13 @@ components:
...
+        with a StellarId. Id's for SEP-12 customers, and consequently SEP-31 senders/receivers, will never include a `memo`
+        or `memo_type`.
```
JakeUrban, line 237: This isn't true. SEP-12 customers, when registered by a SEP-31 sender, will always have memos. This is because the public key is owned by the SEP-31 sender, so we need memos to differentiate the customers.

JakeUrban (commented):
I think there may be some misunderstanding around SEP-12 memos and StellarId's. Happy to talk more about it.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
marcelosalloum: > I'm not sure "SEP Entity" is any more clear than "on-chain entity"

To me, `on-chain entity` is the same as `stellar account id`, which is not exactly what this is for, right?

> "Stellar ID's are objects that identify end-users and SEP-31 senders(...)

Ok

> (they won't identify receivers)

You mean they don't identify the `Receiving Anchors`, right? The receiving end-user can be identified with it.

--
I've updated the description in https://github.com/stellar/java-stellar-anchor-sdk/pull/412/commits/df6d62803935be72225784b835eabd1797eacfab

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
marcelosalloum: Removed in https://github.com/stellar/java-stellar-anchor-sdk/pull/412/commits/df6d62803935be72225784b835eabd1797eacfab.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
marcelosalloum: Updated in https://github.com/stellar/java-stellar-anchor-sdk/pull/412/commits/df6d62803935be72225784b835eabd1797eacfab.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
marcelosalloum: I updated the description in https://github.com/stellar/java-stellar-anchor-sdk/pull/412/commits/a08c05507a781745f8931e393cc6e972bb825972.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
JakeUrban: IDs would be enough, but IDs aren't generated until the anchor sees the account + memo used to determine whether the a customer is new. So in the platform, we could send the IDs for customers and just the Stellar account for SEP-31 senders. Wdyt?

Or, we could just send all the info we have. So ID, account, & memo for customers, and just Stellar account for SEP-31 senders.

JakeUrban: There other issue is that in the future, SEP-24 or SEP-6 non-custodial wallets will need to be identified using their home domain (since they use client attribution). So it is even more convoluted.

marcelosalloum: > IDs would be enough, but IDs aren't generated until the anchor sees the account + memo used to determine whether the a customer is new. So in the platform, we could send the IDs for customers and just the Stellar account for SEP-31 senders. Wdyt?

That's a good idea. We should go with that ✅ 

> Or, we could just send all the info we have. So ID, account, & memo for customers, and just Stellar account for SEP-31 senders.

I think this is the most reliable option and I'm up for going with that, but it will require a more thorough change:
* We purposely decided not to store the customer data in the Platform, so the Anchor Backend can be the single source of truth for customer info.
* Nevertheless, if we want to link the customer account+memo to the SEP-31 transaction, we will need to store it in our database as soon as we create the SEP-31 tx. In order to access that information, we will need to update the `GET {callbackApi}/customer` response body to include not only the id, but also the account and memo.

--
> There other issue is that in the future, SEP-24 or SEP-6 non-custodial wallets will need to be identified using their home domain (since they use client attribution). So it is even more convoluted.

Gotcha. That can be handled when we start supporting those SEPs.

marcelosalloum: Updated in https://github.com/stellar/java-stellar-anchor-sdk/pull/412/commits/4cc4d677e9541afc453f7fd39c6ea3c5fa702591.

JakeUrban (approved):
Looks good, but we need to remove the memo & memo_type from the StellarId object right?

PR #420, opened 2022-07-21
simplify Dockerfile

JakeUrban:
Makes some minor changes to the Dockerfile to make it a bit simpler:

- Uses `WORKDIR` to create the `/code` directory, allowing the `COPY` command to use `. .`
- Removed unnecessary installs (these can always be installed when `docker exec`'ed in the container)
- Removes `ENV` commands.
    - I'm not sure if this would break anything, but we should be able to specify these environment variables at runtime instead of having defaults built into the image
- Removes the `EXPOSE` command, since it actually has no affect

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,13 @@
 FROM openjdk:11-jdk AS build
-# build jar
-ADD . /code
 WORKDIR /code
+COPY . .
 RUN ./gradlew clean bootJar --stacktrace
 
 FROM ubuntu:20.04
 
 RUN apt-get update && \
-apt-get install -y --no-install-recommends \
-        openjdk-11-jre
+    apt-get install -y --no-install-recommends openjdk-11-jre
 
-#temporary for testing reece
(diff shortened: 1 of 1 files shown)


reecexlm:
we need to have SEPs pass before this can be merged. partners are testing with these dockers.

JakeUrban:
Yes, I think there needs to be a pipeline PR, since it looks like our CI processes relied on the default values being set.

The other option is to keep the defaults. Its not a big deal to keep them, but my thought was that we shouldn't set defaults in the image everyone uses just for our own pipelines.

JakeUrban:
Actually, it looks like this may require a kube PR, since the pipelines repo simply deploys the k8 spec files: https://github.com/stellar/pipelines/blob/master/anchor-platform/Jenkinsfile-preview

marcelosalloum (approved):
LGTM!

Just make sure to update the deploys for `dev` and `prd` in stellar/kube.

PR #437, opened 2022-07-28
platform: Set the integration-auth as NONE and disable the metrics in default

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The default value of integration-auth configuration is JWT_TOKEN. We should set it to none.

### Why

N/A

### Known limitations

N/A

### DIFF
--- a/anchor-reference-server/src/main/java/org/stellar/anchor/reference/config/IntegrationAuthSettings.java
+++ b/anchor-reference-server/src/main/java/org/stellar/anchor/reference/config/IntegrationAuthSettings.java
@@ -12,5 +12,5 @@ public class IntegrationAuthSettings {
   AuthType authType = AuthType.NONE;
   String platformToAnchorSecret;
   String anchorToPlatformSecret;
-  long expirationMilliseconds = 30000;
+  Long expirationMilliseconds;
 }
--- a/anchor-reference-server/src/main/java/org/stellar/anchor/reference/config/IntegrationAuthSettings.java
+++ b/anchor-reference-server/src/main/java/org/stellar/anchor/reference/config/IntegrationAuthSettings.java
@@ -1,11 +1,13 @@
 package org.stellar.anchor.reference.config;
 
 import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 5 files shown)


reecexlm:
can you also make default for metrics service disabled in this PR as well.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/config/IntegrationAuthSettings.java
```diff
@@ -1,13 +1,11 @@
 package org.stellar.anchor.reference.config;
 
 import lombok.Data;
-import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 import org.stellar.anchor.config.IntegrationAuthConfig.AuthType;
 
 @Data
 @Configuration
-@ConfigurationProperties(prefix = "integration-auth")
```
marcelosalloum, line 10: Why are you removing this? Won't it be impossible for the reference server to pick up the configuration?

lijamie98: That's right. I have added this back.

platform/src/main/resources/anchor-config-defaults.yaml
```diff
@@ -86,16 +86,16 @@ app-config:
     #     NONE: no authentication is used
     #     API_KEY: the authentication is done using an API key added to the `X-Api-Key` header.
     #     JWT_TOKEN: the authentication is done using a JWT token added to the `Authorization` header. this token is generated from the secret key.
-    authType: JWT_TOKEN
+    authType: NONE
```
marcelosalloum, line 89: The default value for the reference server is still JWT_TOKEN though:
https://github.com/stellar/java-stellar-anchor-sdk/blob/702bbb66cf44491cc0af39223abb108a8e98e44f/anchor-reference-server/src/main/resources/anchor-reference-server.yaml#L26

marcelosalloum (approved):

platform/src/main/resources/anchor-config-defaults.yaml
lijamie98: Fixed!

PR #430, opened 2022-07-26
all: Issue 416 Implement GET `/unique_address` of the callback API

lijamie98:
Closes #416

<details>
  <summary>PR Checklist</summary>

### PR Structure

* [ ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with name of package that is most changed in the PR, ex.
  `paymentservice.stellar`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.
</details>

### What

1. implement GET /unique_address of the callback API
2. Implemented Sep31DepositInfoGeneratorApi
3. Test, test, test.

### Why
N/A

### Known limitations 
N/A

PR #441, opened 2022-07-28
Revert: all: Issue 416 Implement GET `/unique_address` of the callback API

marcelosalloum:
### What

Reverts stellar/java-stellar-anchor-sdk#430

### Why

It broke the integration tests.


marcelosalloum:
Fixed in #444

PR #454, opened 2022-08-02
platform: Dynamic account list handling in StellarPaymentObserver

### LINKED ISSUE #182: Stream all payments endpoint instead of payments for account with multiple accounts
### What

Instead of streaming "payments for account" for multiple accounts, stream all payments and filter the results. This would allow us to monitor payments from multiple/several accounts without reaching a 429 rate limit.

### Why

Address https://github.com/stellar/java-stellar-anchor-sdk/pull/179#discussion_r847740497.

### Further discussion

This change will only be required once the anchor starts monitoring several accounts.

lijamie98:
Closes #182 

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Dynamic account list handling in StellarPaymentObserver.

The observing accounts are also persisted to datastore. 

### Why

1. Increase the efficiency of Stellar network payment operation stream.
2. Dynamically listening to the accounts that the AP is interested in. 

### Known limitations

evictMaxAge and evictCheckPeriod should be made configurable.
https://github.com/stellar/java-stellar-anchor-sdk/issues/492


reecexlm:
is there integration test coverage in our CI for this?

platform/src/main/java/org/stellar/anchor/platform/data/PaymentStreamerCursor.java
```diff
@@ -7,11 +7,13 @@
...
-@Table(name = "stellar_account_page_token")
+@Table(name = "stellar_payment_observer_page_token")
 public class PaymentStreamerCursor {
+  public static final String SINGLETON_ID = "SINGLETON_ID";
+
   @Id
-  @SerializedName("account_id")
-  String accountId;
+  @SerializedName("id")
+  String id = SINGLETON_ID;
```
erika-sdf, line 16: Why switch over to using a single cursor instead of an account cursor?

platform/src/main/java/org/stellar/anchor/platform/data/PaymentObservingAccountRepo.java
```diff
@@ -0,0 +1,8 @@
+package org.stellar.anchor.platform.data;
+
+import org.springframework.data.repository.CrudRepository;
+
+public interface PaymentObservingAccountRepo
```
erika-sdf, line 5: Is this interface actually implemented anywhere?

platform/src/main/java/org/stellar/anchor/platform/PaymentConfig.java
```diff
@@ -58,16 +63,30 @@ public StellarPaymentObserver stellarPaymentObserverService(
...
   }
 
+  @Bean
+  public PaymentObservingAccountsManager observingAccounts(
+      PaymentObservingAccountStore paymentObservingAccountStore) {
+    return new PaymentObservingAccountsManager(paymentObservingAccountStore);
+  }
+
+  @Bean
+  public PaymentObservingAccountStore observingAccountStore(PaymentObservingAccountRepo repo) {
```
erika-sdf, line 86: When a bean is initialized and is passed an interface for an arg, how is this interface initialized?

erika-sdf (commented):
Looks good to me, I just have a few questions

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+    for (PaymentObservingAccount account : accounts) {
+      ObservingAccount oa =
+          new ObservingAccount(account.getAccount(), account.getLastObserved(), true);
+      upsert(oa);
+    }
+
+    // Start the eviction task
+    ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    scheduler.scheduleAtFixedRate(
+        this::evictAndPersist, 60, getEvictPeriod().getSeconds(), TimeUnit.SECONDS);
```
marcelosalloum, line 42: I think we do need to make this configurable and use bigger default expiration times. Maybe 1 MONTH instead of 1 DAY?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+        store.delete(acct.account);
+      }
+    }
+  }
+
+  @AllArgsConstructor
+  public static class ObservingAccount {
+    String account;
+    Instant lastObserved;
+    Boolean expiring;
```
marcelosalloum, line 147: 💡  `expiring` sends the idea it's about to expire and something must be done, which is not a very precise naming here IMO. What do you think about `canExpire`?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+      if (existingAccount == null) {
+        allAccounts.put(observingAccount.account, observingAccount);
+        // update the database
+        store.upsert(observingAccount.account, observingAccount.lastObserved);
+      } else {
+        if (!observingAccount.expiring.equals(existingAccount.expiring))
+          throw new ValueValidationException(
+              String.format(
+                  "The expiring flag cannot be modified. Account=[%s]", observingAccount.account));
+        existingAccount.account = observingAccount.account;
```
marcelosalloum, line 83: 👀  Is this line needed?
```suggestion
```

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @param account The account to be checked.
+   * @return true if the account is being observed. false, otherwise.
+   */
+  public boolean observe(String account) {
+    ObservingAccount acct = allAccounts.get(account);
+    if (acct == null) return false;
+    acct.lastObserved = Instant.now();
+    return true;
+  }
```
marcelosalloum, line 124: 🤔  A method named `observe` would make more sense if we had an **observer** receiving/observing an **event**, but what you're doing here is updating a field rather than observing an event.

What about `isObserving`?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * Evict expired accounts
+   *
+   * @param maxAge evict all accounts that are older than maxAge
+   */
+  public void evict(Duration maxAge) {
+    for (ObservingAccount acct : getAccounts()) {
+      if (!acct.expiring) continue;
+
+      Duration age = Duration.between(Instant.now(), acct.lastObserved).abs();
```
marcelosalloum, line 135: 💡  nit: what about renaming `age` to something like `idleTime` or `inactivityTime`?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+   * @param account The account being removed.
+   */
+  public void remove(String account) {
+    if (account != null) {
+      ObservingAccount targetAccount = allAccounts.get(account);
+      if (targetAccount != null) {
+        allAccounts.remove(account);
+      }
+    }
+  }
```
marcelosalloum, line 101: ❓  Does it make sense to remove the account from the store too?

platform/src/test/kotlin/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManagerTest.kt
```diff
@@ -0,0 +1,181 @@
...
+    obs.upsert("GCK5ECMM67ZN7RWGUSDKQAW6CAF6Q5WYK2VQJ276SJMINU6WVCIQW6BL", true)
+    obs.upsert("GAPBFA5ZYG5VVKN7WPMH6K5CBXGU2AM5ED7S54VX27J7S222NKMTWKR6", true)
+
+    assertEquals(3, obs.accounts.size)
+
+    obs.remove(null)
+    assertEquals(3, obs.accounts.size)
+
+    obs.remove("GCIWQDKACLW26UJXY5CTLULVYUOYROZPAPDDYEQKNGIERVOAXSPLABMB")
+    obs.remove("GCIWQDKACLW26UJXY5CTLULVYUOYROZPAPDDYEQKNGIERVOAXSPLABMB")
```
marcelosalloum, line 59: Did you intend to remove the same account twice here?

platform/src/test/kotlin/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManagerTest.kt
```diff
@@ -0,0 +1,181 @@
...
+import java.time.Instant
+import java.time.temporal.ChronoUnit.DAYS
+import java.time.temporal.ChronoUnit.HOURS
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.stellar.anchor.platform.data.PaymentObservingAccount
+
+class PaymentObservingAccountsManagerTest {
```
marcelosalloum, line 15: Nice tests! 🎉

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -108,10 +97,10 @@ public void onEvent(OperationResponse operationResponse) {
 
             if (observedPayment != null) {
               try {
-                if (observedPayment.getTo().equals(account)) {
+                if (paymentObservingAccountsManager.observe(observedPayment.getTo())) {
                   final ObservedPayment finalObservedPayment = observedPayment;
                   observers.forEach(observer -> observer.onReceived(finalObservedPayment));
-                } else if (observedPayment.getFrom().equals(account)) {
+                } else if (paymentObservingAccountsManager.observe(observedPayment.getFrom())) {
```
marcelosalloum, line 103: Nice 👏

marcelosalloum (commented):
Very cool to see this improvement landing!

I left a few questions and suggestions below, but overall the approach looks good.

There are two important changes that I'd like to see here:
1. We should be updating the observed accounts based on the result from `Sep31DepositInfoGeneratorApi.uniqueAddressIntegration.getUniqueAddress`, right?
2. Making the expiration time configurable seems like a high-priority task for your design, could you incorporate it in tour PR? If it's too much work, we should increase the default expiration times and open a new issue for that.

platform/src/main/java/org/stellar/anchor/platform/data/PaymentStreamerCursor.java
lijamie98: We are switching from streaming multiple accounts to all. That means, we had multiple streams and now only one stream. Therefore, we only need to persist one cursor instead of many.

platform/src/main/java/org/stellar/anchor/platform/data/PaymentObservingAccountRepo.java
lijamie98: The implementation is generated by Spring Data module. That means, only the interfaces are defined. Spring will generate database ORM code according to the names of the interface functions. 

Here is a good example of declaring JDBC repositories.
https://www.baeldung.com/spring-data-jdbc-intro#declaring-jdbc-repositories

platform/src/main/java/org/stellar/anchor/platform/PaymentConfig.java
lijamie98: The concrete class of `interface PaymentObservingAccountRepo` is generated by Spring and added to the Spring IoC container. This line requests an object of type `interface PaymentObservingAccountRepo`. Spring then searches the container, finds the class that implements the interface, and pass the object to the `@Bean` annotated function.

erika-sdf: Thanks for explaining!

platform/src/main/java/org/stellar/anchor/platform/data/PaymentObservingAccountRepo.java
erika-sdf: I missed the nuance that some of these classes are tables. That makes more sense!

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/MemoryStellarPaymentStreamerCursorStore.java
```diff
@@ -5,14 +5,15 @@
 
 public class MemoryStellarPaymentStreamerCursorStore implements StellarPaymentStreamerCursorStore {
   Map<String, String> mapTokens = new HashMap<>();
```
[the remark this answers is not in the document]
lijamie98: removed!

platform/src/main/java/org/stellar/anchor/platform/data/PaymentStreamerCursor.java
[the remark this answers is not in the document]
lijamie98: `id` cannot be final because no-arg constructor is needed for JPA.

Java complaints if no-arg constructor is used and a field is set to `final`.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountStore.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public PaymentObservingAccountStore(PaymentObservingAccountRepo repo) {
+    this.repo = repo;
+  }
+
+  List<PaymentObservingAccount> list() {
+    Log.debug("Retrieving the list of observing account from the store.");
+    List<PaymentObservingAccount> result = new ArrayList<>();
+    repo.findAll().forEach(result::add);
+    return result;
```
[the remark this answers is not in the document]
lijamie98: changed to `paymentObservingAccounts`.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountStore.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  void delete(String account) {
+    PaymentObservingAccount poa = repo.findByAccount(account);
+    if (poa != null) {
+      Log.infoF("Delete account[{}]", account);
+      repo.delete(poa);
+    }
+  }
```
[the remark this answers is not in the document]
lijamie98: Added.

platform/src/main/java/org/stellar/anchor/platform/data/PaymentObservingAccount.java
```diff
@@ -0,0 +1,27 @@
...
+    this.account = account;
+    this.lastObserved = lastObserved;
+  }
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.AUTO)
+  @Column(name = "id", nullable = false)
+  private Long id;
+
+  String account;
```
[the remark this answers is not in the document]
lijamie98: I had made `account` the `@Id` and removed `id` field.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountStore.java
```diff
@@ -0,0 +1,48 @@
...
+
+  void upsert(String account, Instant lastObserved) {
+    Log.infoF("Upserting account[{}, {}]", account, lastObserved);
+    PaymentObservingAccount poa = repo.findByAccount(account);
+    if (poa == null) {
+      poa = new PaymentObservingAccount(account, lastObserved);
+      Log.infoF("Save account[{}, {}] to data store", account, lastObserved);
+      repo.save(poa);
+    } else {
+      if (lastObserved.isAfter(poa.getLastObserved())) {
```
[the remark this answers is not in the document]
lijamie98: Changed.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -108,10 +97,10 @@ public void onEvent(OperationResponse operationResponse) {
...
                   final ObservedPayment finalObservedPayment = observedPayment;
                   observers.forEach(observer -> observer.onSent(finalObservedPayment));
                 }
```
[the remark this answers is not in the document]
lijamie98: Good catch. When we have one stream per account, this use case won't happen. But with one stream, many accounts, this scenario can break our code. 

👍 Nice catch.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @AllArgsConstructor
+  public static class ObservingAccount {
+    String account;
+    Instant lastObserved;
+    Boolean expiring;
+  }
+
```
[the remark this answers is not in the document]
lijamie98: There are primarily two reasons.

`PaymentObservingAccount` is a ORM (ie: Hibernate) managed object. And `ObservingAccount` is stored and handled by the manager. If the ORM changes `account` field, this will break the integrity of the ConcurrentHashMap.

To avoid potential problems caused by ORM, I feel more comfortable to separate these two classes.

In addition, the PaymentObservingAccount table only stores expiring accounts. The manager has to handle all account.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Good name! Changed to `canExpire`.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Yes, the function `getEvictPeriod()` is the place-holder for external configuration. [I plan to use ISO-8601 duration representation string notation used by `Duration.parse()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)). 

Does that make sense to you?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Probably not. But makes me worry if we do not have this line. 😓

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: I had been struggling with this name. `isObserving` implies this function is read-only. But we are actualy updating the `lastObserved` field. 

This is the payment observer to `observe` if the account is being followed, monitored, or observed.

I agree `observe` is not perfect. But choose between `isObserving` and `observe`, I would prefer `observe`.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Changed to idleTime.

platform/src/test/kotlin/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManagerTest.kt
lijamie98: Yes. I want to make sure the second time does not cause failure.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: There is no rea; use case for now. It does not hurt to keep it here but it has no immediate use either. Do you prefer to remove?

platform/src/main/java/org/stellar/anchor/platform/data/PaymentStreamerCursor.java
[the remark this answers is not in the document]
marcelosalloum: Alright, thanks for explaining that!

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
marcelosalloum: Yes, it does!

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
marcelosalloum: That's ok to leave it there then if we want to make things more explicit 👍

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
marcelosalloum: I think we should be consistent with the `evict` method where we're removing it from the store.

https://github.com/stellar/java-stellar-anchor-sdk/blob/eb7fa5b76d213c2c430f79b563c19c12b69139b9/platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java#L126-L141

platform/src/test/kotlin/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManagerTest.kt
marcelosalloum: Cool, thanks for clarifying that 😄

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
marcelosalloum: If this was a private method I'd roll with `observe` but since it's used in other classes (and it's part of the SDK) I still think we should use different wording. Do any of these options appeal to you?
* canObserve
* refreshObservedAccountTimestamp
* tryRefreshingObservedAccountTimestamp
* tryUpdatingLastObservedField
* tryUpdatingLastObservedTimestamp
* elicitObservedAccountTimestamp
* elicitObservedAccount
* pingObservedAccount
* ping

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Good point. I have removed the `remove()` function.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: Renamed to `lookupAndUpdate`

lijamie98:
Adding configurations break the current helm config map. I am delaying the configuration after the new git workflow is in place.

platform/src/test/kotlin/org/stellar/anchor/platform/service/Sep31DepositInfoGeneratorApiTest.kt
```diff
@@ -49,14 +58,21 @@ class Sep31DepositInfoGeneratorApiTest {
...
-    val generator = Sep31DepositInfoGeneratorApi(uniqueAddressIntegration)
+    val generator =
+      Sep31DepositInfoGeneratorApi(uniqueAddressIntegration, paymentObservingAccountsManager)
     val depositInfo = generator.generate(txn)
     assertEquals(stellarAddress, depositInfo.stellarAddress)
     assertEquals(memo, depositInfo.memo)
     assertEquals(memoType, depositInfo.memoType)
+
+    // Check if the manager is observing the account
+    assertTrue(paymentObservingAccountsManager.lookupAndUpdate(uniqueAddress.stellarAddress))
```
marcelosalloum, line 75: Cool, thanks for adding this test 🎉

platform/src/main/java/org/stellar/anchor/platform/service/Sep31DepositInfoGeneratorApi.java
```diff
@@ -4,29 +4,43 @@
...
+
     if (!Objects.toString(uniqueAddress.getMemo(), "").isEmpty()
         && !Objects.toString(uniqueAddress.getMemoType(), "").isEmpty()) {
       // Check the validity of the returned memo and memo type
       MemoHelper.makeMemo(uniqueAddress.getMemo(), uniqueAddress.getMemoType());
     }
 
+    // Add to payment observer manager so that we get events of the stellar address.
+    paymentObservingAccountsManager.upsert(uniqueAddress.getStellarAddress(), true);
+
```
marcelosalloum, line 43: Nice!

marcelosalloum:
> Adding configurations break the current helm config map. I am delaying the configuration after the new git workflow is in place.

Does it break helm even if the configurations are optional and have a default value?

If you prefer to address it later, can you at least add an issue so we can track this piece of work?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,142 @@
...
+    Boolean canExpire;
+  }
+
+  Duration getEvictPeriod() {
+    return Duration.of(1, HOURS);
+  }
+
+  Duration getEvictMaxIdleTime() {
+    return Duration.of(7, DAYS);
+  }
```
marcelosalloum, line 141: 👀  I know this is still arbitrary, but I'm more comfortable having a longer period than 1 WEEK. Can you at least bump it to 2 weeks, since this is not configurable for now?

I'd appreciate if you could either make this as an optional configuration in this PR or create an issue to address it later.

lijamie98:
> > Adding configurations break the current helm config map. I am delaying the configuration after the new git workflow is in place.
> 
> Does it break helm even if the configurations are optional and have a default value?
> 
> If you prefer to address it later, can you at least add an issue so we can track this piece of work?

Yes, the issue had been added to backlog.
https://github.com/stellar/java-stellar-anchor-sdk/issues/492

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param account The account to be checked.
+   * @return true if the account is being observed. false, otherwise.
+   */
+  public boolean lookupAndUpdate(String account) {
+    ObservingAccount acct = allAccounts.get(account);
+    if (acct == null) return false;
+    acct.lastObserved = Instant.now();
+    return true;
+  }
```
marcelosalloum, line 109: ❗ hey, I think you should be saving the updated value of `lastObserved` in the database. 

If you don't, we could incur in problems if we're expecting multiple Sep31Transactions to go to the same account. Imagine this scenario for instance:
1. GET callbackApi/unique_address returns `Account:memo1`
2. After 6 days, GET callbackApi/unique_address returns `Account:memo2` (same account but different memo)
3. The system is re-deployed right after that.
4. The `evict` function will evict the account in one day, but it should actually hold on to it for 7 extra days.
5. If it was persisted in the database, the re-deploy would contain the updated value and would know that the account needs to be observed for 7 extra days.

lijamie98:
> is there integration test coverage in our CI for this? also does this require config changes?

@reecexlm This does not require configuration change, It is covered by the integration test in our CI.

marcelosalloum (commented):
Looks great!

I found an issue ❗ and a concern 👀 below. After we're on the same page there, we should be good to go.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: The issue is created. https://github.com/stellar/java-stellar-anchor-sdk/issues/492

The max idle time is increased to 1 MONTH and flush period is set to 5 minutes.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
lijamie98: `shutdown()` function and @PreDestroy was added to flush to database. This prevents the inconsistency when the service is shutdown for re-deployment.

Since the service become stateful because of the `Map`, we should consider memory grid implementation. I created an issue to track this problem. https://github.com/stellar/java-stellar-anchor-sdk/issues/493

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
marcelosalloum: Cool, thank you!

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * The shutdown hook that is run when Spring terminates. The allAccounts map will be evicted and
+   * flushed.
+   */
+  @PreDestroy
+  public void shutdown() {
+    evictAndPersist();
+  }
```
marcelosalloum, line 54: Cool!

marcelosalloum (approved):
LGTM! Thanks for addressing the comments and opening the side-quests (issues)

PR #464, opened 2022-08-04
Add timeout to end to end test/ fix helm prometheus annotations

stfung77:
What?

Update end-to-end tests and add a timeout for polling a transaction's status
fix prometheus annotations in helm chart


.github/workflows/build_and_test.yml
```diff
@@ -89,9 +89,39 @@ jobs:
...
+
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Run End to End Tests
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f end-to-end-tests/requirements.txt ]; then pip install -r end-to-end-tests/requirements.txt; fi
+          python end-to-end-tests/end_to_end_tests.py --domain ${{ steps.endpoint-finder.outputs.HOME_DOMAIN }} --tests all --secret SDLE32FFTGZT4O6MAUHUUEUNFK3MTH4Q2VJR2TQFYCECJBGDSE2LJILV
```
stfung77, line 121: created a new stellar test account to be used for this. i dont think theres a way for forks to access repository git secrets so just hardcoding this for now. please let me know if you are aware of another way to handle this

this is another account to maintain so i've created this issue to auto recreate assets/accounts after testnet resets
https://github.com/stellar/java-stellar-anchor-sdk/issues/471

marcelosalloum: Having this secret key exposed could cause us problems because the repo is open source and if someone starts messing around with this account, they could easily break our tests. Some things that could break our tests are:
* Close/merge this account into another one.
* Decrease the account balance of our desired assets to zero or a value below the minimum we need for the tests.
* Decrease the XLM balance to zero, making it impossible to pay for fees.
* Close a trustline.

I think the robust approach here could be either:
1. having the tests create and setup a brand new account, so the account only has the balance needed.
2. Or store this secret safely – ops can help with that – and still have a step to re-create and configure the account if needed.

.github/workflows/build_and_test.yml
```diff
@@ -89,9 +89,39 @@ jobs:
...
   complete:
-    needs: [build_and_test, sep_validation_suite]
+    needs: [build_and_test, sep_validation_suite, end_to_end_test]
     runs-on: ubuntu-latest
     steps:
-      - if: contains(needs.*.result, 'failure')
+      - if: contains(needs.*.result, 'failure') || failure()
```
marcelosalloum, line 126: Is this new conditional `|| failure()` needed?

end-to-end-tests/end_to_end_tests.py
```diff
@@ -82,7 +84,10 @@ def get_transaction(endpoints, headers, transaction_id):
     "address": "123 Washington Street",
     "city": "San Francisco",
     "state_or_province": "CA",
-    "address_country_code": "US"
+    "address_country_code": "US",
+    "clabe_number": "1234",
+    "bank_number": "abcd",
+    "bank_account_number": "1234"
```
marcelosalloum, line 90: Thanks for fixing that.

marcelosalloum (commented):
I've left a comment about exposing the secret key. Other than that, it looks good.

.github/workflows/build_and_test.yml
stfung77: >>>
1. having the tests create and setup a brand new account, so the account only has the balance needed.
2. Or store this secret safely – ops can help with that – and still have a step to re-create and configure the account if needed.
>>>
1 - i considered this but the account would still need to be funded with USDC and JPYC. we could have the account try to fund itself with the USDC/XLM and JPYC/XLM offers but those arent always guaranteed to be available since someone could just deplete those offers. 
2 - ops added the secrets to the repo but secrets are only accessible by the repo (when code is merged) and not on forks (PR previews)

.github/workflows/build_and_test.yml
stfung77: having the tests create and setup a brand new account, so the account only has the balance needed.
Or store this secret safely – ops can help with that – and still have a step to re-create and configure the account if needed.
1 - i considered this but the account would still need to be funded with USDC and JPYC. we could have the account try to fund itself with the USDC/XLM and JPYC/XLM offers but those arent always guaranteed to be available since someone could just deplete those offers.
2 - ops added the secrets to the repo but secrets are only accessible by the repo (when code is merged) and not on forks (PR previews)

stfung77: > 
> I think the robust approach here could be either:
> 
> 1. having the tests create and setup a brand new account, so the account only has the balance needed.
> 2. Or store this secret safely – ops can help with that – and still have a step to re-create and configure the account if needed.

1 - i considered this but the account would still need to be funded with USDC and JPYC. we could have the account try to fund itself with the USDC/XLM and JPYC/XLM offers but those arent always guaranteed to be available since someone could just deplete those offers.
2 - ops added the secrets to the repo but secrets are only accessible by the repo (when code is merged) and not on forks (PR previews)

marcelosalloum: > 2 - ops added the secrets to the repo but secrets are only accessible by the repo (when code is merged) and not on forks (PR previews)

Oh, I didn't know that.

I still think this code has a moderate chance of causing problems for us in the future though so we should store the secrets safely. If the CI fails because of a leaked secret we may get blocked while figuring this out, or we can just lose faith in the CI and force-merge things thinking the error is a false negative.

I've found [this text](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) suggesting the usage of a `workflow_run` for these cases, where the script would be executed from the main branch and have access to the secret.

Or you can open your PRs from a repository branch rather than a fork branch. This is simpler but it would change your workflow and would not solve this problem for external contributors.

marcelosalloum: @stfung77: what do you think of the python code grabbing a secret from the environmental variables (if none is provided by the CLI)? This is a pattern often used in Go.

It should be possible to do it similarly to how the java code grabs secrets from k8s (or helm).

stfung77: yeah this is a tricky one. thanks for the link, i'll check it out

reecexlm: agree with marcello. we should not checkin any secrets.  if this is run from github actions, why can't we use github actions secrets? https://docs.github.com/en/rest/actions/secrets

.github/workflows/build_and_test.yml
stfung77: just tested without and it looks like the contains() statement still catches the failure. i'll remove it.

.github/workflows/build_and_test.yml
stfung77: @reecexlm github actions cant access repo secrets from a fork.

i'm looking into running the e2e tests in SDF environment where we can control the secret

marcelosalloum (approved):
LGTM!

PR #474, opened 2022-08-08
fixing paths for buildspec after documentation refactor

reecexlm:
after deploying to partner, some paths which existed prior to documentation refactor remained.  
specifically,  docs/resources/deployment-examples/example-fargate


lijamie98 (approved):
LGTM

PR #475, opened 2022-08-09
bugfix: fix issue where the incoming payment was not populating the event correctly

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix issue where the incoming payment was not populating the event correctly.

### Why

The field `Transaction.stellarTransactions[0].createdAt` was empty because it relied on the transaction object update and it was happening later in time due to a mistake in the code.

I also added tests to make sure this kind of problem would be catch.


platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -97,6 +95,8 @@ public void onReceived(ObservedPayment payment) {
     }
 
     // Check if the payment contains the expected amount (or greater)
+    Instant paymentTime =
```
lijamie98, line 98: 💡 `FormatException` can be thrown. Since this is in a listener, we may do a try/catch block to protect against potential problems.

lijamie98 (approved):

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
marcelosalloum: Added in https://github.com/stellar/java-stellar-anchor-sdk/pull/475/commits/8d3bf604909523080260dd9e65ab5ff713179fde.

PR #494, opened 2022-08-12
AP config refactor

erika-sdf:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

A number of changes that were discussed for the config:
* Refactor and document default config values. These should become the new default config values used by the AP server.
* Remove secrets (which should now be configured through env vars)
  *  Label example.env as an example secrets file
*  Add `anchor-config-dev.yaml` and `anchor-config-integration.yaml` file. This is what we expect a developer would do to spin up their own AP. In these cases, we define the values we use for our own APs during dev and integration testing.
* Convert field names to camel case
* Add version field to the config

### Why
This is in response to the config being large and difficult for a new adopter to use and understand.


reecexlm:
thanks for getting something up for review so quickly!! awsome to get this started. I think what we decided (correct me if I'm mistaken) is to just have a single file.  not named basic or advanced/defaults.
simply, anchor_platform_config.yaml.  This will have a "basic" section at the top, and include all advanced/defaults below which are just commented out. much like the datadog_agent config example
https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml

platform/src/main/resources/anchor-config-defaults.yaml
```diff
@@ -1,305 +1,501 @@
-# This file configures the anchor platform server.
+###############################################################################
+## Anchor Platform - Full Configuration
+##
+## Values not defined in the basic config use the default values defined here.
+###############################################################################
 
 stellar:
   anchor:
```
reecexlm, line 8: do we need this nesting?

platform/src/main/resources/anchor-config-basic.yaml
```diff
@@ -0,0 +1,38 @@
...
+  sep31:
+    enabled: true
+    feeIntegrationEndPoint: https://localhost:8081
+
+  ## Enable events, and point AP at a locally running Kafka queue.
+  ## <all> events are published to queue named <all-events>
+  event:
+    enabled: true
+    publisherType: kafka
+  kafka.publisher:
```
reecexlm, line 34: should be kafkaPublisher ?

platform/src/main/resources/anchor-config-defaults.yaml
```diff
@@ -329,20 +657,24 @@ data-spring-jdbc-local-postgres:
...
-    level:
-      root: INFO
-      org.springframework: INFO
-      org.springframework.web.filter: INFO
-      org.stellar: INFO
-  mvc:
-    async.request-timeout: 6000
\ No newline at end of file
+
+###################################
```
erika-sdf, line 669: @lijamie98 Is this section needed?

erika-sdf:
> thanks for getting something up for review so quickly!! awsome to get this started. I think what we decided (correct me if I'm mistaken) is to just have a single file. not named basic or advanced/defaults. simply, anchor_platform_config.yaml. This will have a "basic" section at the top, and include all advanced/defaults below which are just commented out. much like the datadog_agent config example https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml

@reecexlm -- thanks for pointing that out. I do recall us coming to the conclusion of keeping everything in a single configuration file.

After that discussion, @lijamie98 followed up with a document discussing the approach in some more detail: https://docs.google.com/document/d/1xueA-lCornppEv_tItF5J6oftmAzuUfrLY24NdVh5Rs/edit#

The doc suggests doing both -- maintaining 2 configs (basic and advanced, with the basic config overriding the advanced config), in addition to adding a "summary" section at the top of the advanced config that points at the relevant sections of the config ("events", "app", "metrics", etc). I've included both of these suggestions in this PR so we can take a look and see what we think.

It seems like the discrepancy in our understanding here is over whether we decide to keep the basic config for a less intimidating ramp up for users, or just eliminate it entirely.


It does seem nice to be able to get a basic server up and running with a a config file that is less than 40 lines. If we do plan to keep the basic config, we should definitely label it clearly as something that overrides the "default" configs that are fully specified in the complex config. 


On the other hand, I think adding a basic config does add a layer of complexity to the system. Additionally, after refactoring the "complex" file, it's still quite large, but I think it is conceptually more organized, so the basic config might not be as necessary. 

I lean towards removing the basic config since I think a user will eventually have to make use of the complex config to get their AP instance to do what they want anyway, but I can see the the merits of leaving it as well.

@JakeUrban in case you also want to weigh in here.

JakeUrban:
I really think that the developer shouldn't even know these files exist. The developer should be able to run the platform out of the box with the defaults we've defined internally (using whatever mechanism or file structure we want), and when the developer wants to change the configuration, they need to provide an environment variable that specifies the path to their own config file.

The developers config file would then override the defaults we've defined.

If there are config values we shouldn't provide defaults for, like a SEP-10 signing seed, then the platform should fail until the developer specifies the path to their environment variable and the required variables are present.

For example, in Polaris, if SEP-10 is enabled using the `ACTIVE_SEPS` environment variable, and the `SIGNING_SEED` environment variable is not provided, launching the server fails.

```sh
$ poetry run python manage.py runserver
...
django.core.exceptions.ImproperlyConfigured: Set the SIGNING_SEED environment variable
```

We should have similar behavior here: providing sensible defaults when possible, and requiring values that don't have sensible defaults via environment variables or config files.

JakeUrban:
The developer shouldn't need to go to our github and find our default config file embedded in the source code to learn about each config attribute. They should be able to go to wherever our documentation is and fine the schema & definitions there.

erika-sdf:
Updated this PR to be in line with our discussion this morning.

platform/src/main/resources/anchor-config-default-values.yaml
```diff
@@ -0,0 +1,456 @@
+####################################################################################################
+## Anchor Platform - Default Configuration
+##
+## Values not defined in the user config use the default values defined here.
+####################################################################################################
+
+version: 0.2
```
lijamie98, line 7: Can we use integer number? 
```suggestion
version: 1
```

integration-tests/src/test/resources/anchor-config-integration.yaml
```diff
@@ -0,0 +1,249 @@
...
+metrics:
+  enabled: true
+
+language: en
+
+#######################
+## SEP Configuration
+#######################
+sep1:
+  enabled: true
```
lijamie98, line 51:
```suggestion
  enabled: false
```

platform/src/main/resources/anchor-config-default-values.yaml
```diff
@@ -0,0 +1,456 @@
...
+  # stellarFile: 'file:sep1/stellar-wks.toml'
+
+sep10:
+
+  ## @param: enabled
+  ## @type: bool
+  ## Whether to enable SEP-10 for this instance of AP server.
+  ## If true, @required_secrets: SEP10_SIGNING_SEED
+  #
+  enabled: false
```
lijamie98, line 133:
```suggestion: 
  enabled: true
```

platform/src/main/resources/anchor-config-default-values.yaml
```diff
@@ -0,0 +1,456 @@
...
+  ##     api: the memo and memo type are generated through calling the anchor's GET /unique_address endpoint.
+  #
+  depositInfoGeneratorType: self
+
+sep38:
+  ## @param: enabled
+  ## @type: bool
+  ## Whether to enable SEP-38 for this instance of AP server.
+  #
+  enabled: true
```
lijamie98, line 249:
```suggestion: 
  enabled: false
```

integration-tests/src/test/resources/anchor-config-integration.yaml
erika-sdf: This is the config file used for integration tests. Do we want to have these disabled in that context?

platform/src/main/resources/anchor-config-default-values.yaml
erika-sdf: Sounds good

platform/src/main/resources/anchor-config-basic.yaml
```diff
@@ -0,0 +1,38 @@
...
+##
+## A basic, minimum config to get anchor platform up and running.
+## For a list of all config options, see anchor-config-defaults.yaml.
+#######################################################################
+
+anchorPlatformConfig:
+
+  ## Point anchor platform (AP) at the assets definition
+  app:
+    assets: config/assets-test.json
```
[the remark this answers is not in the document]
erika-sdf: Are we adding an assets endpoint to AP?

erika-sdf:
@lijamie98 I've consolidated the `events` and `data` sections here as we discussed. PTAL

PR #507, opened 2022-08-18
core: Use Kotlin backticks to rename tests to descriptive names

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Use Kotlin backticks to rename tests to descriptive names

The test result will look like:
image

### Why

Kotlin use backtick `` ` ``  as the keyword escape to convert a string into identifier. We can use that to make test function names descriptive. 

### Known limitations

This is POC. Let me know how you feel about making test descriptive to increase the readability.


core/src/test/kotlin/org/stellar/anchor/api/sep/sep10/ChallengeRequestTest.kt
```diff
@@ -12,7 +12,7 @@ internal class ChallengeRequestTest {
   }
 
   @Test
-  fun of() {
+  fun `test ChallanegeRequest creation`() {
```
marcelosalloum, line 15: What do you think of being more explicit about the class and method signature? Something like this:

```suggestion
  fun `test ChallanegeRequest.of`() {
```

erika-sdf (approved):
Cool, looks easier to read

core/src/test/kotlin/org/stellar/anchor/api/sep/sep10/ChallengeRequestTest.kt
lijamie98: It is not proper to use `of`. It's now changed to builder pattern.

PR #512, opened 2022-08-19
platform: Fix the contention problem that causes locked database.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Start the metrics emitter in @PostConstruct. 

### Why

If MetricsEmitter calls Hibernate before it is fully initialized, it causes database locking error in SQLite. We should delay the scheduling of the thread to avoid the contention.

### Known limitations

This issue can be further avoided when we implement the start-up sequence dependency.


marcelosalloum (approved):
LGTM!

PR #521, opened 2022-08-23
core: Make testing connection of UrlValidateionUtil.validateUrl() optional

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The `UrlValidateionUtil.validateUrl()` tries to connect to a URL defined in a config file. However, the URL may not be reachable at the time of construction. 

Added `testConnection` argument to make the connection test optional.

### Why

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/util/UrlValidationUtil.java
+++ b/core/src/main/java/org/stellar/anchor/util/UrlValidationUtil.java
@@ -8,10 +8,16 @@ import org.springframework.validation.Errors;
 
 public class UrlValidationUtil {
   static UrlConnectionStatus validateUrl(String urlString) {
+    return validateUrl(urlString, false);
+  }
+
+  static UrlConnectionStatus validateUrl(String urlString, boolean testConnection) {
     try {
       URL url = new URL(urlString);
-      URLConnection conn = url.openConnection();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/stellar/anchor/util/UrlValidationUtil.java
```diff
@@ -8,10 +8,16 @@
 
 public class UrlValidationUtil {
   static UrlConnectionStatus validateUrl(String urlString) {
+    return validateUrl(urlString, false);
```
erika-sdf, line 11: Do we always want to set this to false? Can we make this specified in configuration or env var maybe?

marcelosalloum (approved):

core/src/main/java/org/stellar/anchor/util/UrlValidationUtil.java
lijamie98: Is it kind of recursion if we make this configurable?

lijamie98: I think we can make the choice to the decision of the validator of each config.

PR #529, opened 2022-08-26
Feature: update how the platform deals with incoming payments with insufficient amounts

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Update how the platform deals with incoming payments with insufficient amounts. The changes include:
* Update the Sep31Transaction schema to hold the list of StellarTransactions. This way, we make sure to keep track of all incoming payments even when they don't fully match the criteria.
* Update JdbcSep31Transaction to store the stellar transactions as a json, instead of a join table
* Update the Platform API design schema.
* Unify similar objects
* Delete unused objects

### Why

We were not handling the use case where the incoming payment amount is smaller than expected.


platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -97,29 +98,59 @@ public void onReceived(ObservedPayment payment) {
...
+              "The incoming payment amount was insufficient! Expected: \"%s\", Received: \"%s\"",
+              formatAmount(expectedAmount), formatAmount(gotAmount));
+      Log.warn(message);
+    } else {
+      updateTransactionStatusTo(PENDING_RECEIVER, txn, payment, stellarTransaction);
+
+      newStatus = TransactionEvent.Status.PENDING_RECEIVER;
+      message = "Incoming payment for SEP-31 transaction";
+      Log.info(message);
     }
```
JakeUrban, line 145: So we're saying that if the amount sent is less than expected, the transaction's status will remain in `pending_sender`? Polaris updates the transaction status to `pending_receiver` regardless of the amount sent, so we should probably discuss what makes sense here so we can be consistent.

My thought is that the anchor needs to decide how to proceed if the amount sent isn't what was expected, so we should change the status to `pending_receiver`, but I could be convinced otherwise.

marcelosalloum: > So we're saying that if the amount sent is less than expected, the transaction's status will remain in `pending_sender`?

No, if the amount is smaller than expected, the status changes to `ERROR` and an event is triggered. This indicates that the receiving anchor should handle this error on their side and decide what to do later.

I'm happy to discuss this further, I don't think the SEP enforces a specific behavior here but as you said, it's better to be consistent.

JakeUrban: Hmm, I'm not sure `error` makes sense here. If the anchor wants to continue facilitating the transaction, they'd have to update the status back to `pending_receiver` or `pending_sender`, and the Sending Anchor may stop watching for updates after they see the `error` status.

Thats why I think using `pending_receiver` makes sense. If the anchor wants to change it to `error` it can, but they could also continue facilitating the transaction without changing the status.

marcelosalloum: So your rationale is that any received transaction (with the expected memo) would trigger a pending_receiver, and the receiver can handle that however they want. And they should validate the amounts are correct on their side as well and maybe mark things as `error` or `refunded` if they want to?

I think that kinda works, because the [statuses section from SEP-31](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#get-transaction) states:
> `pending_receiver` -- payment is being processed by the Receiving Anchor.

This means if the payment amount is smaller than expected, the payment will be processed and marked as error or refunded, depending on the Anchor policy. 

Alright, I'll proceed with that.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -120,38 +120,51 @@ public void onReceived(ObservedPayment payment) {
...
     BigDecimal gotAmount = decimal(payment.getAmount());
-    if (gotAmount.compareTo(expectedAmount) < 0) {
-      updateTransactionStatusTo(ERROR, txn, payment, stellarTransaction);
-
-      newStatus = TransactionEvent.Status.ERROR;
+    String message = "Incoming payment for SEP-31 transaction";
+    ;
+    if (gotAmount.compareTo(expectedAmount) >= 0) {
+      Log.info(message);
+      txn.setTransferReceivedAt(paymentTime);
```
JakeUrban, line 134: So we're not setting the transfer receive time if the payment was insufficient?

JakeUrban (approved):

core/src/main/java/org/stellar/anchor/event/models/TransactionEvent.java
```diff
@@ -81,7 +83,7 @@ public String getType() {
 
   @JsonProperty("stellar_transactions")
   @SerializedName("stellar_transactions")
-  StellarTransaction[] stellarTransactions;
+  List<StellarTransaction> stellarTransactions;
```
lijamie98, line 86: 👍 Nice.

core/src/main/java/org/stellar/anchor/event/models/TransactionEvent.java
```diff
@@ -119,6 +121,11 @@ public enum Status {
       this.status = status;
     }
 
+    @Override
```
lijamie98, line 124: 💡 Can we add detailed status string? It would be somewhat confusing if we see the event printed to debug messages.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -52,7 +50,7 @@ public void onReceived(ObservedPayment payment) {
     // Check if the payment contains the expected asset type
     if (!List.of("credit_alphanum4", "credit_alphanum12").contains(payment.getAssetType())) {
       // Asset type does not match
-      Log.warn("Not an issued asset");
+      Log.info("Not an issued asset");
```
lijamie98, line 53: 💡 nit: Can we add more event information here or at the beginning of the function to provide context of these messages?

lijamie98 (approved):
LGTM with minor comments. Please feel free to merge after taking care of the comments.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
marcelosalloum: Yeah, my rationale is the `transferReceivedAt` would indicate the expected transfer was received, but it actually was not received with the expected specs.

I'm happy to discuss it further.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -120,38 +120,51 @@ public void onReceived(ObservedPayment payment) {
...
 
     // Check if the payment contains the expected amount (or greater)
     BigDecimal expectedAmount = decimal(txn.getAmountIn());
     BigDecimal gotAmount = decimal(payment.getAmount());
-    if (gotAmount.compareTo(expectedAmount) < 0) {
-      updateTransactionStatusTo(ERROR, txn, payment, stellarTransaction);
-
-      newStatus = TransactionEvent.Status.ERROR;
+    String message = "Incoming payment for SEP-31 transaction";
+    ;
```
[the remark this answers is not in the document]
marcelosalloum: Fixed in f3831bc1c15e78a4e0bf8e4bdb388025c3d5f345.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
marcelosalloum: Sure, done in 8e0f8e30fa979abb4f1b2a4c1c1f45422a3c96a1.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
JakeUrban: My thought is that there is no harm in saving the time the payment was received even if it was insufficient, then we can overwrite the value if another transaction with the same memo sends the remaining amount.

But I don't feel strongly.

Issue #528, opened 2022-08-25
Investigate how to turn off flyway auto-apply features


marcelosalloum:
### What the problem is

Flyway currently auto-migrates by default. I've tested it with a local postgres database and noticed that if I set the configuration `spring.flyway.enabled: false`, the database still gets auto-updated.

The weirdest thing is that the database gets updated beyond the migration files. Meaning that if we have new JPA Entities in the code and we don't have migrations to add them, they get auto-created in the Postgres database.

### What's the expected behavior

This behavior should be disabled by default.

In reality, we just want the software to go as far as applying the migration but never applying changes that are not even in the migrations. Even migrations should only be applied if explicitly configured by the user.

stfung77:
close with: https://github.com/stellar/java-stellar-anchor-sdk/pull/532

PR #532, opened 2022-08-29
remove jpa.generate-ddl from postgres config

stfung77:
What:
- remove jpa.generate-ddl from the postgres database configs
- update docker-compose integration tests to use the same image (dont need to build both) for anchor platform and reference server

Why:
postgres database schema is being managed by flyway and there is not need for jpa.generate-ddl to initialize the database

jpa.generate-ddl reference - https://docs.spring.io/spring-boot/docs/1.0.2.RELEASE/reference/html/howto-database-initialization.html

addresses: https://github.com/stellar/java-stellar-anchor-sdk/issues/528


lijamie98 (approved):
LGTM.

platform/src/main/resources/anchor-docker-compose-config.yaml
```diff
@@ -284,10 +284,9 @@ data-spring-jdbc-sqlite:
   spring.datasource.username: admin
   spring.datasource.password: admin
   spring.mvc.converters.preferred-json-mapper: gson
-  spring.liquibase.enabled: false    # certain features that liquibase uses (ex: addForeignKeyConstraint) are not supported by sqlite
+  spring.flyway.enabled: false    # certain features that liquibase uses (ex: addForeignKeyConstraint) are not supported by sqlite
```
lijamie98, line 287: Do we no longer support liquid base?

stfung77: yeah, liquibase was removed a couple months ago

PR #536, opened 2022-08-30
Feature: add integration tests to CI

marcelosalloum:
<details>
  <summary>PR Checklist</summary>

### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with name of package that is most changed in the PR, ex.
  `paymentservice.stellar`, or `all` or `doc` if the changes are broad or impact many
  packages.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.
</details>

### What

Add a new CI step that executes end-to-end tests for the project on the branches `develop` and `release/**`, `hotfix/**`, and whenever we launch a new GH release.

### Why

So we have a safer workflow, where we make sure that our changes are not breaking any of the use cases we support.

PR #541, opened 2022-09-01
remove escape character

stfung77:
remove escape character when checking if escape character exists


marcelosalloum (approved):
LGTM 🚀

stfung77:
to be included in #536

PR #577, opened 2022-09-20
Feature: address snyk vulnerabilities

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix Snyk vulnerabilities by upgrading versions and doing some tweaks to the code.

### Why

To make the project safer.


platform/src/main/java/org/stellar/anchor/platform/payment/observer/circle/CirclePaymentObserverService.java
```diff
@@ -91,11 +92,16 @@ public void handleCircleNotification(CircleNotification circleNotification)
   public void handleSubscriptionConfirmationNotification(CircleNotification circleNotification)
       throws BadRequestException {
     String subscribeUrl = circleNotification.getSubscribeUrl();
-    if (subscribeUrl == null) {
+    if (StringHelper.isEmpty(subscribeUrl)) {
```
lijamie98, line 95: 💡 Can we add static import?

```suggestion
    if (isEmpty(subscribeUrl)) {
```

platform/src/main/java/org/stellar/anchor/platform/payment/observer/circle/CirclePaymentObserverService.java
```diff
@@ -91,11 +92,16 @@ public void handleCircleNotification(CircleNotification circleNotification)
...
       throw new BadRequestException(
           "Notification body of type SubscriptionConfirmation is missing subscription URL.");
     }
 
+    // sanitize
+    if (!subscribeUrl.startsWith("https://")) {
```
lijamie98, line 101: Can we also catch this error in the config validator?

lijamie98: To allow "HTTPS://", 

```suggestion
    if (!subscribeUrl.toLowercase().startsWith("https://")) {
```

service-runner/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 }
 
 dependencies {
+  implementation(libs.snakeyaml)
```
lijamie98, line 13: ❓ Do we use `snakeyaml`?

marcelosalloum: It is a dependency from springboot but the version used by springboot has a vulnerability. I'm manually adding it to make sure we can control the version of this dependency and force springboot to use our version.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/circle/CirclePaymentObserverService.java
marcelosalloum: Good catch! Addressed in https://github.com/stellar/java-stellar-anchor-sdk/pull/577/commits/37a3c9d7ff16eb14a08fe2cb98b1e89789bd9daa.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/circle/CirclePaymentObserverService.java
marcelosalloum: I've adopted your suggestion but I'm not a big fan of static imports unless the methods are widely used.

I also think `StringHelper` provides useful context to the function call, while a plain `isEmpty` can be mistaken by a local method.

service-runner/build.gradle.kts
lijamie98: Thanks for the explanation. 

Does it make sense to upgrade the springboot dependency version?

lijamie98: Can you also leave a comments in the snake yaml dependency? 

If you think it is too much to upgrade spring boot, it is also very ok with me. 

Please feel free to merge when you think it is ok.

lijamie98 (approved):

service-runner/build.gradle.kts
marcelosalloum: There was no available version of springboot where they update that vulnerability themselves, that's why I had to force the update.

I'll check if they've released a new update and if not, I'll add comments as you suggested 👍

PR #609, opened 2022-10-03
Fix: workflow secrets were not being passed from parent (caller) to child (callee) workflows

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

When calling another workflow, make sure it will inherit the secrets from its parent.

### Why

Jobs like [this](https://github.com/stellar/java-stellar-anchor-sdk/actions/runs/3137863047/jobs/5100787037) were failing because the secrets were not being passed from caller to callee.

GH covers this in https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#passing-secrets-to-nested-workflows.

### DIFF
--- a/.github/workflows/branch_develop.yml
+++ b/.github/workflows/branch_develop.yml
@@ -15,6 +15,7 @@ on:
 jobs:
   tests:
     uses: ./.github/workflows/basic_tests.yml  # use the callable tests job to run tests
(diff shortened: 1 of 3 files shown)


lijamie98 (approved):
LGTM.

PR #612, opened 2022-10-04
Fix: stop crashing when parsing response content into ErrorResponse

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Stop crashing when parsing response content into ErrorResponse.

### Why

So we can better debug what's causing an issue.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
@@ -50,35 +50,31 @@ public class PlatformIntegrationHelper {
         "Error returned from the Anchor Backend.\nresponseCode={}\nContent={}",
         responseCode,
         responseContent);
-    ErrorResponse errorResponse;
+    ErrorResponse errorResponse = null;
     try {
       errorResponse = gson.fromJson(responseContent, ErrorResponse.class);
-    } catch (Exception e) { // cannot read body from response
-      Log.errorF("Failed to parse responseContent {} to an ErrorResponse object.", responseContent);
-      return new ServerErrorException("internal server error", e);
+    } catch (Exception e) {
+      // cannot read body from response
+      Log.warn("Failed to parse responseContent to an ErrorResponse object.");
     }
 
-    // Handle 422
-    String errorMessage;
-    if (responseCode == HttpStatus.UNPROCESSABLE_ENTITY.value()) {
-      errorMessage =
-          (errorResponse != null)
-              ? errorResponse.getError()
-              : HttpStatus.BAD_REQUEST.getReasonPhrase();
-      return new BadRequestException(errorMessage);
-    }
-
-    errorMessage =
+    String errorMessage =
         (errorResponse != null)
             ? errorResponse.getError()
-            : HttpStatus.valueOf(responseCode).getReasonPhrase();
-    if (responseCode == HttpStatus.BAD_REQUEST.value()) {
-      return new BadRequestException(errorMessage);
-    } else if (responseCode == HttpStatus.NOT_FOUND.value()) {
-      return new NotFoundException(errorMessage);
+            : (responseCode == 422)
+                ? HttpStatus.BAD_REQUEST.getReasonPhrase()
+                : HttpStatus.valueOf(responseCode).getReasonPhrase();
+
+    switch (responseCode) {
+      case 422:
+      case 400:
+        return new BadRequestException(errorMessage);
+      case 404:
+        return new NotFoundException(errorMessage);
+      default:
+        Log.errorF("Unsupported status code {}.", responseCode);
+        return new ServerErrorException("internal server error");
     }
-    Log.errorF("Unsupported status code {} with body {}", responseCode, errorResponse);
-    return new ServerErrorException("internal server error");
   }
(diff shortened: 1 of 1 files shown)


platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
```diff
@@ -76,7 +77,7 @@ public static AnchorException httpError(String responseContent, int responseCode
     } else if (responseCode == HttpStatus.NOT_FOUND.value()) {
       return new NotFoundException(errorMessage);
     }
-    Log.error(errorMessage);
+    Log.errorF("Unsupported status code {} with body {}", responseCode, errorResponse);
```
leighmcculloch, line 80: Could the error response ever contain sensitive information? Sometimes logging data as is can be leak sensitive information like PII, etc.

leighmcculloch (approved):

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
marcelosalloum: Good point! Fixed in https://github.com/stellar/java-stellar-anchor-sdk/pull/612/commits/bc1280e2f7319cb1d2ad3c522b2ce367285b6390

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
```diff
@@ -50,34 +50,31 @@ public static AnchorException httpError(String responseContent, int responseCode
...
+
+    switch (responseCode) {
+      case 422:
+      case 400:
+        return new BadRequestException(errorMessage);
+      case 404:
+        return new NotFoundException(errorMessage);
+      default:
+        Log.errorF("Unsupported status code {}.", responseCode);
+        return new ServerErrorException("internal server error");
```
lijamie98, line 76: Can we use the HttpStatus code?

```suggestion
    switch (responseCode) {
      case UNPROCESSABLE_ENTITY:
      case BAD_REQUEST:
        return new BadRequestException(errorMessage);
      case NOT_FOUND:
        return new NotFoundException(errorMessage);
      default:
        Log.errorF("Unsupported status code {}.", responseCode);
        return new ServerErrorException("internal server error");
```

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
```diff
@@ -50,34 +50,31 @@ public static AnchorException httpError(String responseContent, int responseCode
...
-    errorMessage =
+    String errorMessage =
         (errorResponse != null)
             ? errorResponse.getError()
-            : HttpStatus.valueOf(responseCode).getReasonPhrase();
-    if (responseCode == HttpStatus.BAD_REQUEST.value()) {
-      return new BadRequestException(errorMessage);
-    } else if (responseCode == HttpStatus.NOT_FOUND.value()) {
-      return new NotFoundException(errorMessage);
+            : (responseCode == 422)
```
lijamie98, line 64: Can we use `UNPROCESSABLE_ENTITY`?

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
```diff
@@ -50,34 +50,31 @@ public static AnchorException httpError(String responseContent, int responseCode
         "Error returned from the Anchor Backend.\nresponseCode={}\nContent={}",
         responseCode,
         responseContent);
-    ErrorResponse errorResponse;
+    ErrorResponse errorResponse = null;
```
lijamie98, line 53: This is not necessary.

lijamie98 (approved):
Please feel free merge after taking care of the feedbacks.

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
marcelosalloum: Done in https://github.com/stellar/java-stellar-anchor-sdk/pull/612/commits/e779fbaf4b5fba524a0698a8207a050d885d86e4.

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
marcelosalloum: Done in https://github.com/stellar/java-stellar-anchor-sdk/pull/612/commits/e779fbaf4b5fba524a0698a8207a050d885d86e4.

platform/src/main/java/org/stellar/anchor/platform/callback/PlatformIntegrationHelper.java
marcelosalloum: It's actually necessary to initialize it because it's a local variable. It wouldn't be required for a class/instance variable though. Here's a blog post on the subject: https://www.baeldung.com/java-error-variable-initialized

BTW, I only know it's needed because Intellij told me 😅

PR #621, opened 2022-10-06
Secrets config refactor

stfung77:
### What

- remove aws creds from sqs, we can use the default aws credentials chain thats built into the SDK
- move data.username, data.password to the secrets manager. can be set using env variables SECRET_DATA_USERNAME and SECRET_DATA_PASSWORD
- update reading env variables behavior to conform with POSIX standards (uppercase and underscores chars only)

### Why

- secrets should be handled by the secret manager
- env variables should conform to POSIX standards so they can be set via commands like `export`

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java
@@ -44,7 +44,7 @@ public class ConfigHelper {
     ConfigMap config = new ConfigMap();
 
     // Determine the version.
-    // if system.env['version'] is not defined, use the latest version.
+    // if system.env['VERSION'] is not defined, use the latest version.
     int version;
     String strVersion = ConfigEnvironment.getenv("VERSION");
     if (strVersion == null) {
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
@@ -9,6 +9,7 @@ public class ConfigMap {
   int version;
   Map<String, ConfigEntry> data;
 
+  // ConfigMap keys will be in normalized form (dot separated hierarchy)
   public ConfigMap() {
     data = new HashMap<>();
   }


core/src/main/java/org/stellar/anchor/util/StringHelper.java
```diff
@@ -24,4 +24,13 @@ public static String camelToSnake(String camel) {
...
         .toLowerCase();
   }
+
+  public static String toUpperSnake(String camel) {
+    return camel
+        .replaceAll("([A-Z]+)([A-Z][a-z])", "$1_$2")
+        .replaceAll("([a-z])([A-Z])", "$1_$2")
+        .replaceAll("-", "_")
+        .replaceAll("\\.", "_")
+        .toUpperCase();
```
lijamie98, line 34: Not sure how we would call something like `sep1.toml`?  `camel path`?

Can we reuse the `camelToSnake()`?

```suggestion
  public static String camelPathtoUpperSnake(String camel) {
    return camelToSnake(camel)
        .replaceAll("\\.", "_")
        .toUpperCase();
```

platform/src/main/java/org/stellar/anchor/platform/config/PropertySecretConfig.java
```diff
@@ -17,4 +17,10 @@ public class PropertySecretConfig implements SecretConfig {
 
   @Value("${secret.platform_api.auth_secret:#{null}}")
   private String platformApiSecret = null;
+
+  @Value("${secret.data.username:#{null}}")
+  private String dataUsername = null;
+
+  @Value("${secret.data.password:#{null}}")
+  private String dataPassword = null;
```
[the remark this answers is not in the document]
lijamie98: This is going through the Spring environment and that is not necessary.

We can do somthing like `SecretManger.getInstance().get("...")`.

platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigEnvironment.java
```diff
@@ -16,13 +15,11 @@ public class ConfigEnvironment {
...
     env = new HashMap<>();
-    for (Map.Entry<String, String> entry : System.getenv().entrySet()) {
-      env.put(normalize(entry.getKey()), entry.getValue());
-    }
+    env.putAll(System.getenv());
 
     Properties sysProps = System.getProperties();
     for (Map.Entry<Object, Object> entry : sysProps.entrySet()) {
-      env.put(normalize(String.valueOf(entry.getKey())), String.valueOf(entry.getValue()));
+      env.put(String.valueOf(entry.getKey()), String.valueOf(entry.getValue()));
```
lijamie98, line 22: I think we should `normalize` otherwise, we are breaking the relaxed binding.

platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java
```diff
@@ -44,7 +46,7 @@ public static ConfigMap loadConfigFromEnv(int latestVersion) throws InvalidConfi
     // Determine the version.
     // if system.env['version'] is not defined, use the latest version.
     int version;
-    String strVersion = ConfigEnvironment.getenv("version");
+    String strVersion = ConfigEnvironment.getenv("VERSION");
```
lijamie98, line 49: We should use the normalized form for stability.

platform/src/main/resources/config/anchor-config-schema-v1.yaml
```diff
@@ -44,6 +42,8 @@ payment_observer.tracked_wallet:
 platform_api.auth.expiration_milliseconds:
 platform_api.auth.type:
 secret.callback_api.secret:
+secret.data.password:
```
lijamie98, line 45: These should not be in the schema.

platform/src/main/java/org/stellar/anchor/platform/configurator/DataConfigAdapter.java
```diff
@@ -92,24 +92,25 @@ void updateSpringEnv(ConfigMap config) throws InvalidConfigException {
...
         set("spring.jpa.database-platform", "org.hibernate.dialect.H2Dialect");
         set("spring.jpa.properties.hibernate.dialect", "org.hibernate.dialect.H2Dialect");
-        copy(config, "data.username", "spring.datasource.username");
-        copy(config, "data.password", "spring.datasource.password");
+        set(
+            "spring.datasource.username",
+            SecretManager.secretManager.props.get("secret.data.username").toString());
+        set(
+            "spring.datasource.password",
+            SecretManager.secretManager.props.get("secret.data.password").toString());
```
lijamie98, line 100:
```suggestion
        copy(config, "secret.data.username", "spring.datasource.username");
        copy(config, "secret.data.password", "spring.datasource.password");
```

lijamie98 (changes requested):

lijamie98:
I suggest that we add uni-tests to show how the env vars are mapped.

platform/src/main/java/org/stellar/anchor/platform/configurator/DataConfigAdapter.java
lijamie98, line 100: SecretManager.getInstance().get("secret.data.password")

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -465,19 +465,6 @@ data:
...
-  #
-  username:
-
-
-  ## @param: username
-  ## @type: string
-  ## The password for database connection
-  #
-  password:
-
```
marcelosalloum, line 480: Can you add comments to this section explaining which secrets will need to be configured in the env vars in order for the DB connection to work?

This file should be self-explanatory in terms of how to configure the application.

marcelosalloum: Is that accurate? I thought you were enforcing it to use the `secret` prefix. So it'd be something like secret.data.password or SECRET_DATA_PASSWORD.

marcelosalloum: Also, could you update example.env with the new secrets?

platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java
lijamie98, line 49: Can we also document what format the ConfigEnvironment is storing (posix format, I guess) and what format the ConfigMap is storing? This will help us to remember in the future.

lijamie98 (approved):
Please feel free to merge the PR after the review feedback is taken care of.

PR #635, opened 2022-10-20
Handle the underlying errors from the Horizon server connection

### LINKED ISSUE #632: StellarPaymentObserver stops processing transactions after the network interruption for extended period of time
BitSo engineering reported the payment observer stops processing transactions after running for a period of time.

**Bug Reproduction**:
1. Start the payment observer with the platform server in a virtual machine with Corretto JDK11. 
2. Stop the network of the VM.
3. Wait for 10-15 minutes
4. Restart the network

The observer will stop processing transactions with the following two scenarios.
1. Silently waiting for transactions.
2. A [SSLProtocolException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/net/ssl/SSLProtocolException.html) is thrown and the observer stops processing the transactions.

**Root Cause Analysis**
In scenario 1, when the observer is silently waiting, `kill -3` was sent to the JVM process and the stack dump shows Horizon SDK and OkHttp library are in RUNNABLE state waiting for the underlying socket to receive data. In this case, we think the socket is in a zombie state and the connection should be remade.

In scenario 2, when the `SSLProtocolException` is thrown, the observer did not handle it correctly to reconnect to the Horizon server. 

**Action Items**
We will do the following three fixes and improvements.
1. Fix the observer's error handling. Upon connection failure, instead of reconnecting to the Horizon server, we will mark the server in un-healthy state to signal Kubernetes to restart the container.
2. Monitor the transaction received. If it takes to long waiting for the transaction, we will mark the observer unhealthy for restart.
3. Add proper HTTP status code to the `/health` endpoint.

lijamie98:
closes #632  

<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

1. Fix the observer's error handling. Upon connection failure, instead of reconnecting to the Horizon server, we will mark the server in un-healthy state to signal Kubernetes to restart the container.
2. Monitor the transaction received. If it takes to long waiting for the transaction, we will mark the observer unhealthy for restart.
3. Add proper HTTP status code to the /health endpoint.

### Why

StellarPaymentObserver stops processing transactions after the network interruption for extended period of time

### Known limitations

N/A


lijamie98:
> not sure if this was considered but this approach of letting kubernetes handle restarting the observer increases downtime and also creates a dependency on kubernetes or some other service that can restart the observer. is there a reason why we dont want to just reconnect on connection failures?

Yes, this is considered. In release 2.0, we may define restart policy. Either to recover from error in-process, or let K8s restart the process.

reecexlm:
i agree with @stfung77 lets restart the stream if it encounters an error.  i see private void restart()  and make some attempts with expential backoff to restart it. it should be the same in 2.0 as well.

docs/resources/docker-examples/kafka/docker-compose.yaml
```diff
@@ -1,14 +1,12 @@
...
   zookeeper:
-    platform: linux/x86_64
     image: confluentinc/cp-zookeeper:latest
     environment:
       ZOOKEEPER_CLIENT_PORT: 2181
       ZOOKEEPER_TICK_TIME: 2000
 
   kafka:
-    platform: linux/x86_64
     image: confluentinc/cp-kafka:latest
```
marcelosalloum, line 10: Why did you remove this? It won't work on a machine that uses a different architecture (it won't work on my Mac M1, for instance).

platform/src/main/java/org/stellar/anchor/platform/controller/HealthController.java
```diff
@@ -16,10 +19,20 @@ public class HealthController {
...
+    HealthCheckResponse healthCheckResponse = healthCheckService.check(checks);
+
+    boolean unhealthy =
+        healthCheckResponse.getChecks().values().stream()
+            .anyMatch(result -> result.getStatus() == HealthCheckStatus.RED);
+    if (unhealthy) {
+      return new ResponseEntity<>(healthCheckResponse, HttpStatus.INTERNAL_SERVER_ERROR);
+    } else {
+      return ResponseEntity.ok(healthCheckResponse);
+    }
```
marcelosalloum, line 36: With this setup you won't be able to identify what exactly is wrong, making most of the health check classes infrastructure useless.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -40,35 +47,69 @@ public class StellarPaymentObserver implements HealthCheckable {
   /** The minimum number of results the Stellar Blockchain can return. */
   private static final int MIN_RESULTS = 1;
 
+  // If the observer had been silent for longer than MAX_SILENT_DURATION, it will be marked
```
marcelosalloum, line 50: What is MAX_SILENT_DURATION?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -40,35 +47,69 @@ public class StellarPaymentObserver implements HealthCheckable {
...
   public void start() {
-    this.stream = watch();
+    this.stream = startSSEStream();
+    this.healthy = true;
+    infoF("Starting the observer watcher");
+    watcher.scheduleAtFixedRate(
+        this::watchTransactionTimeout,
+        10,
+        SILENCE_CHECK_INTERVAL,
+        TimeUnit.SECONDS); // TODO: The period should be made configurable in version 2.x
```
marcelosalloum, line 90: Can you create an issue for that?

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -324,5 +386,10 @@ class StreamHealth {
   boolean threadTerminated;
 
   boolean stopped;
+
+  @SerializedName("last_event_id")
   String lastEventId;
+
+  @SerializedName("silence_duration_seconds")
+  String silenceDurationSeconds;
```
marcelosalloum, line 394: `silenceDurationSeconds` is kinda confusing. What does silence mean?

seconds_since_last_event would be more descriptive.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -86,6 +127,44 @@ private void restart() {
...
+  void handleFailure(Optional<Throwable> exception, Optional<Integer> statusCode) {
+    // The SSEStreamer has internal errors. We will give up and let the container manager to
+    // restart.
+    Log.errorEx("stellar payment observer error: ", exception.get());
+
+    // Stop the SSEStream
+    stream.close();
+
+    // Mark the observer unhealthy
+    healthy = false;
```
marcelosalloum, line 165: We already have a function called `restart()` that pauses the streamer for some (exponential increasing) time and then restarts the streamer after that time has passed. Why don't we just call `restart()`? 

Marking the job as unhealthy in the first failure and not adding any way for the software to recover the job sounds pretty extreme.

Also, your solution relies on the assumption that k8s will detect the unhealthy state and restart the pod, which may not happen depending on the setup.

Also, restarting the pods would make the whole AP unavailable for 2 whole minutes, in case the observer and the sep-server are launched together and there's only a single pod. I feel like this solution will add new problems and would require a very specific deployment setup to work. It can cause more harm than good.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
```diff
@@ -40,35 +47,69 @@ public class StellarPaymentObserver implements HealthCheckable {
...
+    if (lastActivityTime != null) {
+      Duration silenceDuration = Duration.between(lastActivityTime, now);
+      debugF("The observer had been silent for {} seconds", silenceDuration.getSeconds());
+      if (silenceDuration.getSeconds() > MAX_SILENT_TIMEOUT) {
+        infoF(
+            "The observer had been silent for {} seconds. Marking it unhealthy",
+            silenceDuration.getSeconds());
+        this.healthy = false;
+      }
+    }
```
marcelosalloum, line 104: Have you explored the `onClosed` listener? That could give you a simpler and more reliable way to identify that a connection has been closed, so you can restart it.

marcelosalloum (commented):
Killing the observer forever on the first failure sounds like a problematic design that could cause more harm than good. I'd explore ways to restart the service from the software, rather than giving up on the streamer and relying on k8s orchestration.

We already have a `restart()` method and the streamer can make use of `onFailure`, `onClosed`, and `onOpen` to identify if the streamer is active or if it needs to be (re)started.

platform/src/main/java/org/stellar/anchor/platform/controller/HealthController.java
lijamie98: Good to see you here, Marcelo!

This only marks the failure of the health check. The detail/reason of the failure is still in the JSON response.

docs/resources/docker-examples/kafka/docker-compose.yaml
lijamie98: `platform` is only defined after `2.4`. The number is now increased from 2 to 2.4.

platform/src/main/java/org/stellar/anchor/platform/payment/observer/stellar/StellarPaymentObserver.java
lijamie98: renamed

reecexlm (approved):
LGTM

PR #686, opened 2022-12-20
[ANCHOR-103] Add kotlin reference server

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add Kotlin reference server for SEP-24. Currently mirrors minimal functionality from existing reference server. Interactive workflow to be finished in the separate PR.

### Why

It should be more friendly towards developers reading into examples not familiar with Java/Spring

### Known limitations

N/A


.run/1. Platform Server - ServiceRunner.run.xml
```diff
@@ -8,7 +8,7 @@
       <env name="SECRET.DATA.USERNAME" value="admin" />
       <env name="secret.sep10.jwt_secret" value="jwt_secret" />
       <env name="secret.sep10.signing_seed" value="SAKXNWVTRVR4SJSHZUDB2CLJXEQHRT62MYQWA2HBB7YBOTCFJJJ55BZF" />
-      <env name="STELLAR_ANCHOR_CONFIG" value="file:///$PROJECT_DIR$/integration-tests/src/test/resources/integration-test.anchor-config.yaml" />
+      <env name="STELLAR_ANCHOR_CONFIG" value="file:///$PROJECT_DIR$/platform/src/main/resources/example.anchor-config.yaml" />
```
lijamie98, line 11: Please create a new configuration that runs the `example.anchor-config.yaml`. 

We may probably rename the file to `Platform Server (Integration Test Configuration)` and `Platform Server (Example Configuration)

lijamie98 (changes requested):

lijamie98 (approved):
LGTM

docs/resources/deployment-examples/example-fargate/example_config/anchor_config.yaml
```diff
@@ -150,7 +150,7 @@ app-config:
     interactiveJwtExpiration: 3600
 
     # The interactive URL where the platform server will redirect to start the SEP-24 interactive flow.
-    interactiveUrl: http://localhost:8081/sep24/interactive
+    interactive_url: http://localhost:8083/sep24/interactive
```
JakeUrban, line 153: Why did we change the attribute to snake case? The rest of the attributes use camel case. Won't think break things?

platform/src/main/resources/example.anchor-config.yaml
```diff
@@ -83,7 +83,7 @@ sep12:
 
 sep24:
   enabled: true
-  interactiveUrl: http://localhost:8081/sep24/interactive
+  interactive_url: http://localhost:8083/sep24/interactive
```
JakeUrban, line 86: Its hard to understand whats being changed for each configuration file.. why are there so many 😅

kotlin-reference-server/src/main/kotlin/com/example/plugins/Sep24.kt
```diff
@@ -0,0 +1,34 @@
...
+      val operation =
+        call.parameters["operation"]
+          ?: return@get call.respondText(
+            "Missing operation parameter",
+            status = HttpStatusCode.BadRequest
+          )
+
+      when (operation.lowercase()) {
+        "deposit" -> call.respondText("The sep24 interactive DEPOSIT starts here.")
+        "withdraw" -> call.respondText("The sep24 interactive WITHDRAW starts here.")
```
JakeUrban, line 25: So this PR is just to create the kotlin reference server's structure right? We also need to create functionality for providing off-chain status updates back to the anchor platform.

.run/1. Platform Server - ServiceRunner [Example cfg].run.xml
```diff
@@ -0,0 +1,32 @@
+<component name="ProjectRunConfigurationManager">
```
Comment (JakeUrban) [inline .run/1. Platform Server - ServiceRunner [Example cfg].run.xml:1]: Does this have to be XML?

docs/resources/deployment-examples/example-fargate/example_config/anchor_config.yaml
Ifropc: I'm pretty sure it's case insensitive, either way works

platform/src/main/resources/example.anchor-config.yaml
Ifropc: I'm pretty sure most of them are unused, but just in case I changed it everywhere

kotlin-reference-server/src/main/kotlin/com/example/plugins/Sep24.kt
Ifropc: Yes, I indent to do it in different PR to reduce number of changes

.run/1. Platform Server - ServiceRunner [Example cfg].run.xml
Comment (Ifropc) [reply .run/1. Platform Server - ServiceRunner [Example cfg].run.xml:1]: It's just idea config

service-runner/src/main/java/org/stellar/anchor/platform/ServiceRunner.java
```diff
@@ -55,11 +57,20 @@ static ConfigurableApplicationContext startStellarObserver(Map<String, Object> e
...
+
     if (strPort != null) {
       port = Integer.parseInt(strPort);
     }
+    if (kStrPort != null) {
+      kPort = Integer.parseInt(kStrPort);
+    }
+
     AnchorReferenceServer.start(port, "/");
+    ReferenceServerKt.start();
```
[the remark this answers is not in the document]
Ifropc: I don't know yet, I would rather have them both in kotlin. For starers, we can use kotlin server for both test and reference (as it'll be simple and functional enough for both of them). Later, I would separate them

platform/src/main/resources/example.anchor-config.yaml
lijamie98: Gleb has a good suggestion to further simplified the configuration yaml file. I have reduced the number of files once but it kind of growing back as the project progresses. 

I had created a [story ](https://stellarorg.atlassian.net/browse/ANCHOR-120) to track.

PR #698, opened 2022-12-28
docs: [ANCHOR-126] Simplify the component schemas

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The transaction event of the `Event Schema.yml` is SEP-31 centric. When we add SEP-24 and SEP-6, the schema can become unstable and requires changes as the project develops. These fields also duplicate the fields defined in the `Platform API.yml` which increases the difficulties for the community to adopt the event schema.

Similar component schemas are merged and unused schemas were removed.

### Why

For simplification and improvement.

### Known limitations

N/A


JakeUrban (changes requested):
I tried to design the `TransactionEvent` schema to be SEP-generic, can you describe what you think isn't generalized?

The only fields I see that could be confusing are `customers` & `creator`.

We needed `customers` for SEP-31, but in SEP-24 & 6 the customers (`customers.sender` and `customers.receiver`) were intended to be identical.

We needed `creator` for SEP-31 to save the send anchor's public key, but `creator` can also be used for the Stellar wallet. If wallet attribution is used we can put the wallet's home domain in the `creator` field.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -102,92 +103,46 @@ components:
...
-          $ref: '#/components/schemas/Amount'
-        amount_in:
-          $ref: '#/components/schemas/Amount'
-        amount_out:
-          $ref: '#/components/schemas/Amount'
-        amount_fee:
-          $ref: '#/components/schemas/Amount'
-        quote_id:
-          type: string
-          description: "The id of the [SEP-38](https://stellar.org/protocol/sep-38) firm quote used for the transaction."
```
marcelosalloum, line 144: Is this really a good idea to remove this?

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -102,92 +103,46 @@ components:
...
-          type: string
-          format: date-time
-        completed_at:
-          type: string
-          format: date-time
-          description: "When the transaction was marked as `completed` or when it was fully refunded."
-        transfer_received_at:
-          type: string
-          format: date-time
-          description: "When the Stellar transaction related to this object was received."
```
marcelosalloum, line 158: All these timestamps seem important.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -102,92 +103,46 @@ components:
...
         message:
           type: string
           description: "This field may contain any clarifying message related to this TransactionEvent or to the transaction itself."
-        refunds:
-          $ref: '#/components/schemas/Refunds'
-        stellar_transactions:
-          type: array
-          items:
-            $ref: '#/components/schemas/StellarTransaction'
```
marcelosalloum, line 167: It'd be good to return some Ledger identification, like the transaction or operation id, so the Anchor can reach out to the Ledger for more information.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -102,92 +103,46 @@ components:
...
-          description: |
-            The Identification info of the sending and receiving customers. If they were created through [SEP-12](https://stellar.org/protocol/sep-12),
-            this object should contain the SEP-12 customer `id`. Otherwise, the `account` address of the customer.
-          properties:
-            sender:
-              $ref: '#/components/schemas/StellarId'
-            receiver:
-              $ref: '#/components/schemas/StellarId'
-        creator:
-          $ref: '#/components/schemas/StellarId'
```
marcelosalloum, line 190: Are you sure you want to remove any references from the customers related to this event? This information cannot be retrieved any other way in the future by the Anchor, can it?

marcelosalloum (commented):
So you want to force the Anchor to always call `GET /transaction`, right? I think that kind of defeats some of the purpose of the Events approach.

Why don't you unify the payloads of the Event Schema and PlatformAPI? You can update the event so it returns 3 fields: event type, event status, and payload, where the payload is the exact same response you'd get with GET {platform API}/transaction.

This way you can simplify the events, keep parity with the PlatformAPI and keep providing this information to the Anchor.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
lijamie98: We will add the `transaction` as the payload of the transaction event as you have suggested. 

In that case, all these fields will be available in the transaction event schema.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -11,6 +11,7 @@ info:
 tags:
   - name: "Event Schemas"
     description: "Event Schemas for events that are published to a queue for the Anchor to consume."
+paths:
```
marcelosalloum, line 14: To solve the syntax error, I think you need this:
```suggestion
paths: {}
```

marcelosalloum (commented):
About these fields you're removing, are all of them available in the Platform API?

I took a look at the platform API and didn't find a few of them, like:
- status_change
- source_account
- destination_account

I think it's ok to drop the status_change, but the others should probably be added to the Transaction Schema. WDYT?

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
lijamie98: I do not think `status_change` is being used and not sure if it has a meaningful use case. The `status_change` should be derived from the log where we record every state of all fields.

The `source_account` and `destination_account` are available in [StellarTransactions.](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/docs/03%20-%20Implementing%20the%20Anchor%20Server/Communication/Platform%20API.yml#L355)

marcelosalloum: SGTM!

Do you think it's worth writing a small section in the README or CHANGELOG explaining how to proceed with this change? This way you can just ask our partners to look at that section to understand how to proceed with the updated version of the event.

marcelosalloum (approved):
LGTM!

Plz take a look at [this comment](https://github.com/stellar/java-stellar-anchor-sdk/pull/698#discussion_r1062984236) before merging. Just a suggestion to make your life easier in the future 😄

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
lijamie98: Oh yeah... definitely. I will add CHANGELOG which is closer to the change.

marcelosalloum (approved):
LGTM!

lijamie98:
> I tried to design the `TransactionEvent` schema to be SEP-generic, can you describe what you think isn't generalized?
> 
> The only fields I see that could be confusing are `customers` & `creator`.
> 
> We needed `customers` for SEP-31, but in SEP-24 & 6 the customers (`customers.sender` and `customers.receiver`) were intended to be identical.
> 
> We needed `creator` for SEP-31 to save the send anchor's public key, but `creator` can also be used for the Stellar wallet. If wallet attribution is used we can put the wallet's home domain in the `creator` field.

These fields are available in the `transaction` payload.

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.0
 info:
-  version: "0.1"
+  version: "1.0.0"
```
JakeUrban, line 3: How should this version relate to the AP's version? Should it match? Or should it be independent?

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
```diff
@@ -44,7 +45,7 @@ components:
           type: string
           description: price including fees.
         creator:
-          $ref: '#/components/schemas/StellarId'
+          $ref: './Platform API.yml#/components/schemas/StellarId'
```
JakeUrban, line 48: Ah, very cool, I didn't know we could do this.

JakeUrban (commented):
Hey Jamie, thanks for working on this. I think we should make a minor adjustment before merging.

Each event schema could have the same top level keys:
- `id` (this is the `event_id`)
- `type`
- `data`

The structure of `data` is dependent on `type`, and we can put the current `QuoteEvent` & `TransactionEvent` properties inside it. So for example:

For a transaction event:

```js
{
  "id": "1", // or whatever format we use for IDs
  "type": "transaction_created", // or something else transaction-related
  "data": {
    "transaction": {
      // all the other TransactionEvent properties
    }
  }
}
```

For quote events:

```js
{
  "id": "1",
  "type": "quote_created",
  "data": {
    "quote": {
      // all QuoteEvent properties, including the newly added 'transaction' object
    }
  }
}
```

docs/03 - Implementing the Anchor Server/Communication/Events Schema.yml
lijamie98: I would prefer independent. But all APIs should have the same version string.

However, I don't see the way to share the version string in swagger.

docs/03 - Implementing the Anchor Server/Communication/schemas.yml
```diff
@@ -0,0 +1,489 @@
...
+          description: |
+            The transaction event type. Can be one of the following:
+            - `transaction_created` - a transaction was created through the SEP endpoints. The payload is in the `transaction` field.
+            - `transaction_status_changed` - the status of a transaction has changed. The payload is in the `transaction` field.
+            - `transaction_error` - error processing the transaction. The payload is in the `transaction` field.
+            - `quote_created - a quote was created via the SEP38 API. The payload is in the `quote` field.
+        transaction:
+          $ref: '#/components/schemas/Transaction'
+        quote:
+          $ref: '#/components/schemas/Quote'
```
JakeUrban, line 467: Just double checking that you don't want to add a generic container key for the event payload? Something like `data` or `payload`?

lijamie98: I think the generic container key `data` is not necessary. Do you have any good use case for the `data` field?

JakeUrban: I've just seen it used in other event systems, like Circle's. Its not strictly necessary but it does provide a separation between the event scheme (`id`, `type`, & `data`) vs. the schema specific to the event type, which is inside `data`. If we ever had an event with lots of top-level attributes, then it would be nicer to have a container attribute.

I don't feel strongly though, so I'll leave it to you.

lijamie98: Thanks Jake. Appreciated your insight very much. I guess we may always nest extra top-level attributes. My gut feeling is to keep it simple at this moment. 

We still have some time until we release the AP and we may always revisit this issue.

PR #705, opened 2023-01-09
custom labels and remove observer ingress

reecexlm:
This PR enables add custom labels to the deployment and the containers for sep service and payment observer service. it also enables the ability to define java options in values file for the helm chart to be included when the jar is executed.


reecexlm (commented):
asking @stfung77 to approve as I opened the original PR

PR #714, opened 2023-01-23
fix CI e2e tests

stfung77:
- pin ubuntu version to 20.04, there is some issue with latest that causes e2e docker compose tests to file. i did not dive into the root cause

- reverted the change to the master flyway migration file. we can't directly edit this file, we have to generate a new migration file to cover the changes we want to add


.github/workflows/end_to_end_tests.yml
```diff
@@ -4,20 +4,20 @@ on:
...
-#      - 'hotfix/**'
+      - develop
+      - 'release/**'
+      - 'releases/**'
+      - 'hotfix/**'
 
 jobs:
   end_to_end_tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
lijamie98, line 16: 👍 This is good idea!

lijamie98 (approved):

PR #721, opened 2023-01-27
[Anchor-149] Fix customer type not being properly resolved (when multiple types are configured)

### LINKED ISSUE #717: Anchor-platform searches for wrong customertype to validate if the receiver exists.
We configured a new receiving customer type, on top of our pre existing receiving_user .When
 using the id of this new customer type created to call transactions and I am receiving that the customer does not exist. Looking into the logs, I see this:

```Unable to find customer FindCustomer(id=*******, account=null, memo=null, type=receiving_user) due to Failure(code=customer.not.found, reason=, details=[], cause=Optional.empty, i18nData=Failure.I18nData(code=failure.i18.default, defaultMessage=, arguments=[]))

### PR DESCRIPTION
**Pull request by Ifropc**
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Adding iteration over customer types in validateSenderAndReceiver.

### Why

Should fix #717. Previously, this method was searching for the customer with a first type in the map. Now, it searches for all types. 

### Known limitations

Customer types must be mutually exclusive (i.e. same customer can NOT have multiple types at the same time)

### DIFF
--- a/anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
+++ b/anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
@@ -33,6 +33,14 @@ public class CustomerService {
       if (maybeCustomer.isEmpty()) {
         throw new NotFoundException(
             String.format("customer for 'id' '%s' not found", request.getId()));
+      } else if (request.getType() != null) {
+        if (!request.getType().equals("sep31-sender")
+            && !request.getType().equals("sep31-receiver")) {
+          throw new NotFoundException(
+              String.format(
+                  "customer for 'id' '%s' and 'type' '%s' not found",
+                  request.getId(), request.getType()));
+        }
       }
     } else {
       maybeCustomer =
--- a/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
@@ -539,21 +539,49 @@ public class Sep31Service {
       throw new BadRequestException("receiver_id cannot be empty.");
     }
 
-    // TODO: Populate customer type with the first receiver type; this is a temporary fix for cases
-    // where customerType is required
     Sep12Operation sep12Operation = Context.get().getAsset().getSep31().getSep12();
-    String receiverType = null;
+    Sep12GetCustomerResponse receiver;
+
+    // Fix for https://github.com/stellar/java-stellar-anchor-sdk/issues/717 (ANCHOR-149)
+    // TODO: check if there is a better way to do it, i.e. request a type first for a customer OR
+    // stop iteration after first hit.
     if (sep12Operation != null) {
-      Optional<String> receiverTypeOptional =
-          sep12Operation.getReceiver().getTypes().keySet().stream().findFirst();
-      receiverType = receiverTypeOptional.orElse(null);
-    }
-    Sep12GetCustomerRequest request =
-        Sep12GetCustomerRequest.builder().id(receiverId).type(receiverType).build();
-    Sep12GetCustomerResponse receiver = this.customerIntegration.getCustomer(request);
-    if (receiver == null || receiver.getStatus() != Sep12Status.ACCEPTED) {
-      infoF("Customer (receiver) info needed for request ({})", Context.get().getRequest());
-      throw new Sep31CustomerInfoNeededException("sep31-receiver");
+      Set<String> types = sep12Operation.getReceiver().getTypes().keySet();
+      List<Sep12GetCustomerRequest> requests =
+          types.stream()
+              .map(type -> Sep12GetCustomerRequest.builder().id(receiverId).type(type).build())
+              .collect(Collectors.toList());
+      List<Sep12GetCustomerResponse> responses = new ArrayList<>();
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java [hunk 1]
```diff
@@ -33,6 +33,14 @@ public GetCustomerResponse getCustomer(GetCustomerRequest request) throws NotFou
       if (maybeCustomer.isEmpty()) {
         throw new NotFoundException(
             String.format("customer for 'id' '%s' not found", request.getId()));
+      } else if (request.getType() != null) {
+        if (!request.getType().equals("sep31-sender")
+            && !request.getType().equals("sep31-receiver")) {
```
spolaore, line 38: Not sure if we should be hard-coding those types, this will probably break anchos using different types

Ifropc: This is a change in reference (example) implementation. In this example, Business server only accept this 2 types (and reject others)

spolaore: ah this is the reference server, my bad

JakeUrban:
Hi Gleb, can we update the PR title to be more specific to whats being changed? That way someone can read the commit log and get a sense of what has changed.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -539,21 +539,53 @@ private void validateSenderAndReceiver()
...
 
-    // TODO: Populate customer type with the first receiver type; this is a temporary fix for cases
-    // where customerType is required
     Sep12Operation sep12Operation = Context.get().getAsset().getSep31().getSep12();
-    String receiverType = null;
+    Sep12GetCustomerResponse receiver;
+
+    // Fix for https://github.com/stellar/java-stellar-anchor-sdk/issues/717 (ANCHOR-149)
+    // TODO: check if there is a better way to do it, i.e. request a type first for a customer OR
+    // stop iteration after first hit.
```
JakeUrban, line 547: Is there a reason we're not checking the customer status after each request and stopping if we get ACCEPTED?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -539,21 +539,53 @@ private void validateSenderAndReceiver()
...
+        infoF("Customer (receiver) info needed for request ({})", Context.get().getRequest());
+        throw new Sep31CustomerInfoNeededException(types.toString());
+      }
+      if (accepted.size() != 1) {
+        infoF(
+            "Ambiguous customer info for request ({}) having multiple accepted responses ({})",
+            Context.get().getRequest(),
+            accepted);
+        throw new Sep31AmbiguousCustomerInfoException(accepted.toString());
+      }
```
JakeUrban, line 580: I'm struggling with that we should do in this scenario. The Anchor Platform v1 has the assumption that types are mutually exclusive, so if this happens, the AP should probably return 500 to the Sending Anchor, right?

I don't see any changes to the Anchor Platform in this PR, only the underlying Java SDK. Where is the logic for how the platform handles these AnchorExceptions?

Ifropc: @JakeUrban 
org.stellar.anchor.reference.controller.GlobalControllerExceptionHandler

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
Ifropc: I don't think I'm following the question, can you elaborate please?

JakeUrban: Right now the implementation makes all the requests, then checks all the statuses. Why don't we make a request & check the status before we make the next request?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
JakeUrban: Ok, it looks like its throwing 500 for all exceptions that aren't specifically cased in the file. I don't see `AnchorException` listed, so I'm assuming 500 is returned?

Ifropc: Yes. There are handlers for specific AnchorExceptions, but generic one returns 500

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
Ifropc: Do you suggest to request and check them one by one until ACCEPTED found?
We can do that. But if there is a misconfiguration  `Sep31AmbiguousCustomerInfoException` won't be thrown.

JakeUrban: Oh you're right. Ok, lets keep it as-is.

JakeUrban (approved):

lijamie98 (approved):

reecexlm (approved):

PR #722, opened 2023-01-27
[ANCHOR-136] [ANCHOR-137] [ANCHOR-158] Schema fixes

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fixing schema. 
1. Add memo and memo_type
2. Add amount_requested
3. Fix message not updated for SEP-24
4. Fix memo set to NONE instead of null for SEP-24

### Why

Bug fixes

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -31,6 +31,7 @@ import org.stellar.anchor.event.EventService;
 import org.stellar.anchor.util.GsonUtils;
 import org.stellar.sdk.KeyPair;
 import org.stellar.sdk.Memo;
+import org.stellar.sdk.MemoNone;
 
 public class Sep24Service {
   public static final String OPERATION_WITHDRAW = "withdraw";
@@ -136,8 +137,7 @@ public class Sep24Service {
             .transactionId(txnId)
             .status(INCOMPLETE.toString())
             .kind(WITHDRAWAL.toString())
-            .amountIn(strAmount)
-            .amountOut(strAmount)
+            .requestedAmount(strAmount)
             .assetCode(assetCode)
             .assetIssuer(withdrawRequest.get("asset_issuer"))
             .startedAt(Instant.now())
@@ -145,7 +145,7 @@ public class Sep24Service {
             .fromAccount(sourceAccount)
             .clientDomain(token.getClientDomain());
 
-    if (memo != null) {
+    if (!(memo instanceof MemoNone)) {
       debug("transaction memo detected.", memo);
       builder.memo(memo.toString());
       builder.memoType(memoTypeString(memoType(memo)));
@@ -249,8 +249,7 @@ public class Sep24Service {
             .transactionId(txnId)
             .status(INCOMPLETE.toString())
             .kind(Sep24Transaction.Kind.DEPOSIT.toString())
-            .amountIn(strAmount)
-            .amountOut(strAmount)
+            .requestedAmount(strAmount)
             .assetCode(assetCode)
             .assetIssuer(depositRequest.get("asset_issuer"))
             .startedAt(Instant.now())
@@ -259,7 +258,7 @@ public class Sep24Service {
             .clientDomain(token.getClientDomain())
             .claimableBalanceSupported(claimableSupported);
 
-    if (memo != null) {
+    if (!(memo instanceof MemoNone)) {
       debug("transaction memo detected.", memo);
       builder.memo(memo.toString());
       builder.memoType(memoTypeString(memoType(memo)));
--- a/docs/03 - Implementing the Anchor Server/Communication/schemas.yml
+++ b/docs/03 - Implementing the Anchor Server/Communication/schemas.yml
@@ -251,6 +251,8 @@ components:
           $ref: '#/components/schemas/Amount'
         amount_fee:
           $ref: '#/components/schemas/Amount'
+        amount_requested:
+          $ref: '#/components/schemas/Amount'
         quote_id:
           type: string
         started_at:
@@ -278,6 +280,10 @@ components:
         external_transaction_id:
           type: string
         custodial_transaction_id:
+          type: string
+        memo:
+          type: string
+        memo_type:
           type: string
         customers:
           type: object
--- a/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
@@ -119,4 +119,7 @@ public class JdbcSep24Transaction extends JdbcSepTransaction
 
   @SerializedName("claimable_balance_supported")
   Boolean claimableBalanceSupported;
+
+  @SerializedName("amount_requested")
+  String requestedAmount;
 }
--- a/platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
@@ -6,9 +6,7 @@ import org.stellar.anchor.api.exception.SepException;
 import org.stellar.anchor.api.platform.GetTransactionResponse;
 import org.stellar.anchor.api.platform.PlatformTransactionData;
 import org.stellar.anchor.api.sep.SepTransactionStatus;
-import org.stellar.anchor.api.shared.Amount;
-import org.stellar.anchor.api.shared.RefundPayment;
-import org.stellar.anchor.api.shared.Refunds;
+import org.stellar.anchor.api.shared.*;
 import org.stellar.anchor.platform.data.JdbcSep24Transaction;
 import org.stellar.anchor.platform.data.JdbcSep31Transaction;
 import org.stellar.anchor.platform.data.JdbcSepTransaction;
@@ -72,6 +70,14 @@ public class TransactionHelper {
         .amountIn(new Amount(txn.getAmountIn(), txn.getAmountInAsset()))
         .amountOut(new Amount(txn.getAmountOut(), txn.getAmountOutAsset()))
         .amountFee(new Amount(txn.getAmountFee(), txn.getAmountFeeAsset()))
+        .amountRequested(
+            new Amount(
+                txn.getRequestedAmount(),
+                "stellar:" + txn.getRequestAssetCode() + ":" + txn.getRequestAssetIssuer()))
+        .customers(
+            new Customers(
+                StellarId.builder().account(txn.getToAccount()).build(),
+                StellarId.builder().account(txn.getFromAccount()).build()))
         .startedAt(txn.getStartedAt())
         .updatedAt(txn.getUpdatedAt())
         .completedAt(txn.getCompletedAt())
--- a/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
@@ -24,6 +24,9 @@ public class PlatformTransactionData {
   @SerializedName("amount_expected")
   Amount amountExpected;
 
+  @SerializedName("amount_requested")
+  Amount amountRequested;
+
   @SerializedName("amount_in")
   Amount amountIn;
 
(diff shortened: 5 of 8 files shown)


platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -72,6 +70,14 @@ static GetTransactionResponse toGetTransactionResponse(JdbcSep24Transaction txn)
         .amountIn(new Amount(txn.getAmountIn(), txn.getAmountInAsset()))
         .amountOut(new Amount(txn.getAmountOut(), txn.getAmountOutAsset()))
         .amountFee(new Amount(txn.getAmountFee(), txn.getAmountFeeAsset()))
+        .amountRequested(
+            new Amount(
+                txn.getRequestedAmount(),
+                "stellar:" + txn.getRequestAssetCode() + ":" + txn.getRequestAssetIssuer()))
```
JakeUrban, line 76: Using `stellar:` here will only work for SEP-31 transactions and SEP-24 withdrawals. Don't we want something like this? We need to support the deposit case, where the amount requested will be an off-chain asset.
```suggestion
            new Amount(txn.getRequestedAmount(), txn.getAmountRequestedAsset())
```

Ifropc: It's the value passed in the request. Both withdrawal and deposit have asset_code and asset_issuer. Am I missing something? 
Maybe the logic should be
```
if (asset_issuer == null) {
 stellar:code:issuer
} else {
  code
}
```
?

JakeUrban: Maybe I'm misunderstanding the use case for the new attribute. I thought it was the same as Polaris' [amount_expected](https://django-polaris.readthedocs.io/en/stable/api.html#polaris.models.Transaction.amount_expected), which is always in the same units as `amount_in`.

So if its a SEP-24 deposit transaction, the asset being sent to the anchor is an offchain asset. `amount_in` is in units of the off-chain asset, and `amount_out` is in units of the Stellar asset. `amount_expected` should be the amount we expect to receive from the user.

Ifropc: This is the (optional) amount that was passed to the anchor platform in the SEP request and an asset. That's it.

```
asset_code 	string 	Code of the asset the user wants to withdraw. The value passed must match one of the codes listed in the [/info](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#info) response's withdraw object


asset_code 	string 	The code of the stellar asset the user wants to receive for their deposit with the anchor. The value passed must match one of the codes listed in the [/info](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#info) response's deposit object.
```

Ifropc: I think I get what you are saying. Should we just return asset_code and stellar:code:issuer if issuer is non-null?

JakeUrban: Ok, so then this field is different than `amount_expected`. `amount_requested` will be the amount received by the user in the deposit case, and the amount received by the anchor in the withdraw case. Lets just make sure thats documented properly.

JakeUrban: I'm just confused on why we're adding this field. Can you provide some context on why we need it?

Ifropc: I think we can just use amount_in and asset_in instead. I don't remember the details on why I wanted to add it

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -145,7 +144,7 @@ public InteractiveTransactionResponse withdraw(
             .fromAccount(sourceAccount)
             .clientDomain(token.getClientDomain());
 
-    if (!(memo instanceof MemoNone)) {
+    if (memo != null) {
```
JakeUrban, line 147: I thought we identified this as a bug?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -249,8 +248,7 @@ public InteractiveTransactionResponse deposit(
             .transactionId(txnId)
             .status(INCOMPLETE.toString())
             .kind(Sep24Transaction.Kind.DEPOSIT.toString())
-            .amountIn(strAmount)
-            .amountOut(strAmount)
+            .requestedAmount(strAmount)
```
JakeUrban, line 251: Why are we removing amount_in and amount_out here? I think we may want to hold off on adding these requestedAmount changes until we understand the use case.

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
lijamie98: The `amountRequested` has different asset representation with `amountIn` and `amountOut`. We should have the same representation.

lijamie98 (changes requested):

lijamie98:
Can we change the title of the PR to more clearly reflect the contents of the changes?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
Ifropc: Jamie fixed it in another PR

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
Ifropc: Because we don't know amount_in and amount_out during the request time, they should be set by the business server

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
Ifropc: This is fixed in the latest commit. Now assets follow the logic:
1. If is set in db, return db data
2. If not, use `assetService` to query (using request asset code and issuer)
For `amountExpected` field, we always use request data.

core/src/main/java/org/stellar/anchor/sep9/Sep9Fields.java
```diff
@@ -5,61 +5,62 @@
...
-      "organization.photo_proof_address",
-      "organization.address_country_code",
-      "organization.state_or_province",
-      "organization.city",
-      "organization.postal_code",
-      "organization.director_name",
-      "organization.website",
-      "organization.email",
-      "organization.phone");
+  public static final List<String> SEP_9_FIELDS =
```
JakeUrban, line 8: I think the formatter my VS Code IDE is using a different formatter than your Jetbrains IDE. Is there a standard formatter we should use?

docs/03 - Implementing the Anchor Server/Communication/schemas.yml
```diff
@@ -251,6 +251,8 @@ components:
           $ref: '#/components/schemas/Amount'
         amount_fee:
           $ref: '#/components/schemas/Amount'
+        amount_requested:
+          $ref: '#/components/schemas/Amount'
```
JakeUrban, line 255:
```suggestion
        amount_expected:
          $ref: '#/components/schemas/Amount'
```

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
```diff
@@ -119,4 +119,7 @@ public void setRefunds(Sep24Refunds refunds) {
 
   @SerializedName("claimable_balance_supported")
   Boolean claimableBalanceSupported;
+
+  @SerializedName("amount_requested")
+  String amountExpected;
```
JakeUrban, line 124:
```suggestion
  @SerializedName("amount_expected")
  String amountExpected;
```

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -58,20 +59,31 @@ static GetTransactionResponse toGetTransactionResponse(JdbcSep31Transaction txn)
...
         .sep(PlatformTransactionData.Sep.SEP_24)
         .kind(PlatformTransactionData.Kind.from(txn.getKind()))
         .status(SepTransactionStatus.from(txn.getStatus()))
-        .amountIn(new Amount(txn.getAmountIn(), txn.getAmountInAsset()))
-        .amountOut(new Amount(txn.getAmountOut(), txn.getAmountOutAsset()))
-        .amountFee(new Amount(txn.getAmountFee(), txn.getAmountFeeAsset()))
+        .amountIn(Amount.create(txn.getAmountIn(), amountInAsset))
+        .amountOut(Amount.create(txn.getAmountOut(), amountOutAsset))
+        .amountFee(Amount.create(txn.getAmountFee(), amountFeeAsset))
+        .amountExpected(new Amount(txn.getAmountExpected(), amountExpectedAsset))
```
JakeUrban, line 82: Is there a reason why some amounts are using `Amount.new()` and others are using `new Amount()`?

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -81,6 +93,18 @@ static GetTransactionResponse toGetTransactionResponse(JdbcSep24Transaction txn)
         .build();
   }
 
+  private static String makeAsset(
+      @Nullable String dbAsset, AssetService service, JdbcSep24Transaction txn) {
+    if (dbAsset != null) {
+      return dbAsset;
+    }
+
+    AssetInfo info = service.getAsset(txn.getRequestAssetCode(), txn.getRequestAssetIssuer());
```
JakeUrban, line 102: There shouldn't be any function called `getRequestAssetCode()` or `getRequestAssetIssuer()`

core/src/main/java/org/stellar/anchor/sep9/Sep9Fields.java
Ifropc: It's ./gradlew spottlesApply

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
Ifropc: Why not? It's what we store in DB

JakeUrban: This field already exists in the DB? How do we use it today if not for the amount expected?

Ifropc: Used for refunds

JakeUrban: Ok, then I'll resolve this. We may want to include a migration to a column to the same name eventually though.

JakeUrban (approved):

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
Ifropc: I guess we can store expectedAsset instead

api-schema/src/main/java/org/stellar/anchor/api/shared/Amount.java
```diff
@@ -12,4 +13,12 @@ public class Amount {
   String asset;
 
   public Amount() {}
+
```
lijamie98, line 16: 💡 You may use Lombok [@AllArgsConstructor](http://www.javabyexamples.com/delombok-allargsconstructor-noargsconstructor-and-requiredargsconstructor)

lijamie98 (changes requested):

lijamie98 (approved):

api-schema/src/main/java/org/stellar/anchor/api/shared/Amount.java
Ifropc: https://github.com/stellar/java-stellar-anchor-sdk/blob/b22f7b54f11c0c81d515cc947aeae950d75ca8d5/api-schema/src/main/java/org/stellar/anchor/api/shared/Amount.java#L10
It's already annotated with it

PR #730, opened 2023-02-06
SEP-24: add `refund_memo` and `refund_memo_type` support

JakeUrban:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Adds `refund_memo` and `refund_memo_type` support to SEP-24 `POST /transactions/withdraw/interactive` request.

### Why

This was added to SEP-24. ([#1321](https://github.com/stellar/stellar-protocol/pull/1321))

### DIFF
--- a/core/src/main/java/org/stellar/anchor/util/MemoHelper.java
+++ b/core/src/main/java/org/stellar/anchor/util/MemoHelper.java
@@ -48,12 +48,12 @@ public class MemoHelper {
   }
 
   public static Memo makeMemo(String memo, String memoType) throws SepException {
-    if (isEmpty(memoType)) {
-      memoType = "none";
-    }
-
     if (isEmpty(memo)) {
       return null;
+    }
+
+    if (isEmpty(memoType)) {
+      throw new SepValidationException("memo_type is required if memo is specified");
     }
 
     switch (memoType) {
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ target/
 */bin/*
 */generated/*
 *.tgz
-
+.vscode
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -130,19 +130,21 @@ public class Sep24Service {
     }
 
     Memo memo = makeMemo(withdrawRequest.get("memo"), withdrawRequest.get("memo_type"));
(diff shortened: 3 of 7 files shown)


JakeUrban:
@lijamie98 I noticed an inconsistencies with the SEP while making this change:

`account` is not required for SEP-24 `POST /transactions/withdraw/interactive`, but the service raises an error if its not specified. It should default to the authenticated account.

I noticed `account` is required for SEP-24 `POST /transations/deposit/interactive` though. This is an inconsistency and we should update the SEP so that `account` is optional just like the withdraw endpoint, and then we'll update the Anchor Platform to match it.

core/src/main/java/org/stellar/anchor/util/MemoHelper.java
```diff
@@ -48,14 +48,14 @@ public static MemoType memoType(Memo memo) {
...
-      memoType = "none";
-    }
-
     if (isEmpty(memo)) {
       return null;
     }
 
+    if (isEmpty(memoType)) {
+      throw new SepValidationException("memo_type is required if memo is specified");
+    }
```
JakeUrban, line 57: This is a unrelated change, @lijamie98 let me know if you think I should revert. We should probably require memo type to be specified instead of assuming the type is none, because the error message returned will assume memo type was specified.

lijamie98:
> @lijamie98 I noticed an inconsistencies with the SEP while making this change:
> 
> `account` is not required for SEP-24 `POST /transactions/withdraw/interactive`, but the service raises an error if its not specified. It should default to the authenticated account.
> 
> I noticed `account` is required for SEP-24 `POST /transations/deposit/interactive` though. This is an inconsistency and we should update the SEP so that `account` is optional just like the withdraw endpoint, and then we'll update the Anchor Platform to match it.

Hey @JakeUrban, I will submit a PR for the inconsistency.

The SEP does not specify the account in deposit as optional. https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request

I think it is a bug in SEP24. Can you confirm?

core/src/main/java/org/stellar/anchor/util/MemoHelper.java
lijamie98: I think we can keep your change that throws exception unless there are cases that we must return `none`.

lijamie98 (approved):
LGTM.

reecexlm:
not seeing any test coverage

PR #732, opened 2023-02-07
Make account field optional in SEP24 withdraw and deposit

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Make account field optional in SEP24 withdraw and deposit

### Why

To reflect the SEP-24 change.

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -99,10 +99,6 @@ public class Sep24Service {
 
     if (isEmpty(sourceAccount)) {
       sourceAccount = token.getAccount();
-      if (isEmpty(sourceAccount)) {
-        info("missing 'account' field");
-        throw new SepValidationException("'account' is required");
-      }
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -99,9 +97,12 @@ public InteractiveTransactionResponse withdraw(
       throw new SepValidationException("missing 'asset_code'");
     }
 
-    if (sourceAccount == null) {
-      info("missing 'account' field");
-      throw new SepValidationException("'account' is required");
+    if (isEmpty(sourceAccount)) {
+      sourceAccount = token.getAccount();
+      if (isEmpty(sourceAccount)) {
```
JakeUrban, line 102: I don't think we need this check, since SEP-10 is required and SEP-10 tokens always have an account? Same goes for the deposit endpoint.

lijamie98: Removed both unnecessary changes.

JakeUrban (approved):

PR #746, opened 2023-02-11
support assets.json file

JakeUrban:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/core/src/main/java/org/stellar/anchor/asset/DefaultAssetService.java
+++ b/core/src/main/java/org/stellar/anchor/asset/DefaultAssetService.java
@@ -25,6 +25,8 @@ public class DefaultAssetService implements AssetService {
     switch (assetsConfig.getType()) {
       case JSON:
         return fromJson(assetsConfig.getValue());
+      case FILE:
+        return fromResource(assetsConfig.getValue());
       case YAML:
       default:
         infoF("assets type {} is not supported", assetsConfig.getType());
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/stellar/anchor/asset/DefaultAssetService.java
```diff
@@ -25,6 +25,8 @@ public static DefaultAssetService fromAssetConfig(AssetsConfig assetsConfig)
     switch (assetsConfig.getType()) {
       case JSON:
         return fromJson(assetsConfig.getValue());
+      case FILE:
```
lijamie98, line 28: In our previous discussions, we decided not to have external files. Since we are adding it back here, can we take a closer look at the use cases?

`FILE` is too narrow because we should also consider `https://` URLs. 

💡 We can use `EXTERNAL` or `URL` and use the URL string such as `https://` or `file://`

JakeUrban: I think supporting separate documents is better than requiring developers to put json text in a yaml document. If we only support yaml for the assets config, then we could use a single document.

Do we need a `assets.type` config at all? Can't we tell based on the value anyway? For example, if the value starts with `http` then we know its a URL, if it starts with `file` its a file, and otherwise we interpret it as either json or yaml in the same document?

JakeUrban: I feel the same way about other config objects too. Do we really need `type` configs for the stellar.toml file or for the database? Can't we just infer the type based on the value(s)?

PR #748, opened 2023-02-13
[Anchor-169] add bank_account_type in customer fields and SEP-9 fields

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Adds `bank_account_type` to the list of SEP-9 fields, with initial supported values of checking or savings. Some jurisdictions have banking systems that require that the account type be specified in addition to the account number.

Note that `bank_account_type` does not indicate the format of `bank_account_number`, and should not be used to validate `bank_account_number`.

If an anchor needs to be able to validate an account identifier, it should use other SEP-9 fields added specifically for the format they're looking to validate, such as `cbu_number` or `clabe_number`. These fields often contain pieces of information other than the user's account number with their bank.

For example, CBU numbers contain identifiers for the bank, the bank branch, the type of account, and the account number itself. Instead of trying to decompose CBU numbers and passing each piece via the `bank_account_number`, `bank_number`, `bank_branch_number`, and other SEP-9 fields, it is much easier for businesses operating in those jurisdictions to accept account identifiers in the formats that are common in those jurisdictions.

### Why

N/A

### Known limitations

N/A

### DIFF
--- a/anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
+++ b/anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
@@ -150,6 +150,11 @@ public class CustomerService {
       } else {
         fields.put("bank_account_number", createBankAccountNumberField(type));
       }
+      if (customer.getBankAccountType() != null) {
+        providedFields.put("bank_account_type", createBankAccountTypeProvidedField());
+      } else {
+        fields.put("bank_account_type", createBankAccountTypeField(type));
+      }
       if (customer.getBankRoutingNumber() != null) {
         providedFields.put("bank_number", createBankNumberProvidedField());
       } else {
@@ -193,6 +198,9 @@ public class CustomerService {
     if (request.getBankAccountNumber() != null) {
       customer.setBankAccountNumber(request.getBankAccountNumber());
     }
+    if (request.getBankAccountType() != null) {
+      customer.setBankAccountType(request.getBankAccountType());
+    }
     if (request.getBankNumber() != null) {
       customer.setBankRoutingNumber(request.getBankNumber());
     }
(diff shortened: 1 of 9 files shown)


anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
```diff
@@ -250,6 +259,14 @@ public CustomerField createBankAccountNumberField(String type) {
     return field;
   }
 
+  public CustomerField createBankAccountTypeField(String type) {
+    CustomerField field = new CustomerField();
+    field.setType("string");
+    field.setDescription("bank account type of the customer");
+    field.setOptional(!type.equals(Customer.Type.SEP31_RECEIVER.toString()));
+    return field;
+  }
```
JakeUrban, line 268: Since `bank_account_type` is an enum, lets add the `choices` list that can be defined for fields with a limited set of accepted values.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
```diff
@@ -298,6 +315,14 @@ public ProvidedCustomerField createBankAccountNumberProvidedField() {
     return field;
   }
 
+  public ProvidedCustomerField createBankAccountTypeProvidedField() {
+    ProvidedCustomerField field = new ProvidedCustomerField();
+    field.setType("string");
+    field.setDescription("bank account type of the customer");
+    field.setStatus(Customer.Status.ACCEPTED.toString());
+    return field;
+  }
```
JakeUrban, line 324: Same goes here, lets add the `choices` list.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
lijamie98: Added

anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
lijamie98: Added

JakeUrban (approved):

PR #757, opened 2023-02-15
Fixes for stellarTransactionId and Customers

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Small fixes.
1. Make it possible to update `stellarTransactionId` in patch (previously missed updates from payment observer). Also add it to `getTransaction`
2. Fix `amountExpected` in `getTransaction` (see comment)
3. Update customers logic. Now sender is always user account address (regardless of type). Receiver is set to null for deposit, and to `withdraw_anchor_account` for withdraw.

### Why

To make reference server work

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
@@ -57,8 +57,11 @@ public class PlatformTransactionData {
   @SerializedName("stellar_transactions")
   List<StellarTransaction> stellarTransactions;
 
-  @SerializedName("stellar_transaction_id")
-  String stellarTransactionId;
+  @SerializedName("source_account")
+  String sourceAccount;
+
+  @SerializedName("destination_account")
+  String destinationAccount;
 
   @SerializedName("external_transaction_id")
   String externalTransactionId;
--- a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
@@ -196,6 +196,7 @@ public class TransactionService {
     txnUpdated = updateField(patch, txn, "externalTransactionId", txnUpdated);
     // update stellar_transactions
     txnUpdated = updateField(patch, txn, "stellarTransactions", txnUpdated);
+    txnUpdated = updateField(patch, txn, "stellarTransactionId", txnUpdated);
 
     switch (txn.getProtocol()) {
       case "24":
--- a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
@@ -196,7 +196,6 @@ public class TransactionService {
     txnUpdated = updateField(patch, txn, "externalTransactionId", txnUpdated);
     // update stellar_transactions
     txnUpdated = updateField(patch, txn, "stellarTransactions", txnUpdated);
-    txnUpdated = updateField(patch, txn, "stellarTransactionId", txnUpdated);
 
     switch (txn.getProtocol()) {
       case "24":
--- a/platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
@@ -1,6 +1,6 @@
 package org.stellar.anchor.platform.utils;
 
-import static org.stellar.anchor.api.platform.PlatformTransactionData.Kind.RECEIVE;
+import static org.stellar.anchor.api.platform.PlatformTransactionData.Kind.*;
 
 import javax.annotation.Nullable;
 import org.stellar.anchor.api.exception.SepException;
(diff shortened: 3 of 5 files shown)


JakeUrban:
Can we have `sender` and `receiver` refer to the same public key? SEP-24 transactions are often referred to as "me-to-me" since the user is sending to themselves, making themselves the receiver too.

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -196,6 +196,7 @@ void updateSepTransaction(PlatformTransactionData patch, JdbcSepTransaction txn)
     txnUpdated = updateField(patch, txn, "externalTransactionId", txnUpdated);
     // update stellar_transactions
     txnUpdated = updateField(patch, txn, "stellarTransactions", txnUpdated);
+    txnUpdated = updateField(patch, txn, "stellarTransactionId", txnUpdated);
```
JakeUrban, line 199: The intention is to have the business send a `stellarTransactions` list containing the stellar transaction id. It doesn't make sense to have both a single transaction ID field and a list.

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -66,10 +66,18 @@ static GetTransactionResponse toGetTransactionResponse(
       refunds = toRefunds(txn.getRefunds(), txn.getAmountInAsset());
     }
 
+    String sender = DEPOSIT.kind.equals(txn.getKind()) ? txn.getToAccount() : txn.getFromAccount();
```
JakeUrban, line 69: The sending and receiving customers are identified by the account authenticated in SEP-10, not the destination account for the payment. Also note that customers of custodial wallets are identified by an account AND memo.

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -66,10 +66,18 @@ static GetTransactionResponse toGetTransactionResponse(
...
     String amountInAsset = makeAsset(txn.getAmountInAsset(), assetService, txn);
     String amountOutAsset = makeAsset(txn.getAmountOutAsset(), assetService, txn);
     String amountFeeAsset = makeAsset(txn.getAmountFeeAsset(), assetService, txn);
     String amountExpectedAsset = makeAsset(null, assetService, txn);
+    Customers.CustomersBuilder builder =
+        Customers.builder().sender(StellarId.builder().account(sender).build());
+
+    if (WITHDRAWAL.kind.equals(txn.getKind()) && txn.getWithdrawAnchorAccount() != null) {
+      builder.receiver(StellarId.builder().account(txn.getWithdrawAnchorAccount()).build());
+    }
```
JakeUrban, line 80: You're assigning the businesses's destination account for receiving payments from users as `customer.receiver`, which is incorrect. The Stellar Id for the receiver should be the account authenticated via SEP-10, and usually this is also the source of the payment to the anchor.

JakeUrban (changes requested):

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
Ifropc: Yes it can be removed

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
Ifropc: I think there is a misunderstanding on my side. If we use customers for this purpose, where should the payment destination be set?

JakeUrban: Ah.. we don't have field for it. We should probably just add `destination_account` and `source_account` fields to the top-level transaction object.

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -66,10 +66,18 @@ static GetTransactionResponse toGetTransactionResponse(
...
     }
 
+    String sender = DEPOSIT.kind.equals(txn.getKind()) ? txn.getToAccount() : txn.getFromAccount();
+
     String amountInAsset = makeAsset(txn.getAmountInAsset(), assetService, txn);
     String amountOutAsset = makeAsset(txn.getAmountOutAsset(), assetService, txn);
     String amountFeeAsset = makeAsset(txn.getAmountFeeAsset(), assetService, txn);
     String amountExpectedAsset = makeAsset(null, assetService, txn);
+    Customers.CustomersBuilder builder =
+        Customers.builder().sender(StellarId.builder().account(sender).build());
```
lijamie98, line 76: This is too verbose. Can we use constructor instead of builders for simple classes?

api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
```diff
@@ -57,8 +57,11 @@ public class PlatformTransactionData {
   @SerializedName("stellar_transactions")
   List<StellarTransaction> stellarTransactions;
 
-  @SerializedName("stellar_transaction_id")
-  String stellarTransactionId;
+  @SerializedName("source_account")
```
lijamie98, line 60: Can we use `from_account` which is the same as the database?

Ifropc: `source_account` is always user account.
So it's `from` for deposit and `to` for withdrawal

JakeUrban: `source_account` is always the source of the payment. In deposits, this is the anchor's account. In withdraw's, its the user's account.

Ifropc: Ok, I will just mirror to and from account

platform/src/main/java/org/stellar/anchor/platform/utils/TransactionHelper.java
```diff
@@ -79,11 +80,10 @@ static GetTransactionResponse toGetTransactionResponse(
         .amountIn(Amount.create(txn.getAmountIn(), amountInAsset))
         .amountOut(Amount.create(txn.getAmountOut(), amountOutAsset))
         .amountFee(Amount.create(txn.getAmountFee(), amountFeeAsset))
-        .amountExpected(Amount.create(txn.getAmountExpected(), amountExpectedAsset))
-        .customers(
-            new Customers(
-                StellarId.builder().account(txn.getToAccount()).build(),
-                StellarId.builder().account(txn.getFromAccount()).build()))
```
JakeUrban, line 86: Instead of removing these fields, can we just fix their values? Businesses will still need to use this info to identify their users.

JakeUrban (commented):
Almost there, one comment.

lijamie98 (approved):

JakeUrban (approved):

PR #759, opened 2023-02-15
[ANCHOR-173] Removed data.url and add data.server and data.database fields

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This is to replace `data.url` with `data.server` and `data.database` fields.

Here is the original data configuration:
```
  data: 
    url: jdbc:postgresql://127.0.0.1:5432/postgres
```

Here is the data configuration after the refactoring.

```
data:
  type: postgres
  server: 127.0.0.1:5432
  database: postgres
```

### Why

This is to avoid mistakes while constructing database url string to reduce the user friction. 

### Known limitations

N/A

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -492,10 +492,10 @@ data:
   #
 #  url: jdbc:h2:mem:anchor-platform
 
-  ## The hostname of the database server.
+  ## The hostname and port of the database server.
   server:
 
-  ## Name of the database. Optional. Default to anchor-platform or the default name of the type of database.
+  ## Name of the database.
   database:
 
   ## @param: initial_connection_pool_size
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -485,11 +485,19 @@ data:
   #
   #
   type: h2
+
   ## @param: url
   ## @type: string
   ## Location of the database
   #
-  url: jdbc:h2:mem:anchor-platform
+#  url: jdbc:h2:mem:anchor-platform
+
+  ## The hostname of the database server.
+  server:
+
+  ## Name of the database. Optional. Default to anchor-platform or the default name of the type of database.
+  database:
+
   ## @param: initial_connection_pool_size
   ## @type: integer
   ## Initial number of connections
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -486,12 +486,6 @@ data:
   #
   type: h2
 
-  ## @param: url
-  ## @type: string
-  ## Location of the database
-  #
-#  url: jdbc:h2:mem:anchor-platform
-
   ## The hostname and port of the database server.
   server:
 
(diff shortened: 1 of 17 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -485,11 +485,19 @@ data:
...
   ## @type: string
   ## Location of the database
   #
-  url: jdbc:h2:mem:anchor-platform
+#  url: jdbc:h2:mem:anchor-platform
+
+  ## The hostname of the database server.
+  server:
+
+  ## Name of the database. Optional. Default to anchor-platform or the default name of the type of database.
```
JakeUrban, line 498: Does this mean in postgres, it defaults to "postgres"? What do you think about requiring this field instead?

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -485,11 +485,19 @@ data:
...
+
   ## @param: url
   ## @type: string
   ## Location of the database
   #
-  url: jdbc:h2:mem:anchor-platform
+#  url: jdbc:h2:mem:anchor-platform
+
+  ## The hostname of the database server.
+  server:
```
JakeUrban, line 496: what would the value be for databases that don't use a server, like h2 or sqlite?

JakeUrban (commented):
Left some questions, but more generally, aren't we missing changes to other files to process these values?

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: If type is h2, the server and database will be ignore.

sqlite has a database name.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: I think we should make it required. If the database does not exist, AP will fail. 

We can add automatic database creation if that is becoming an issue.

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: Sounds straightforward!

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: Awesome, lets do it.

Ifropc (approved):

JakeUrban (approved):

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -485,11 +485,13 @@ data:
...
-  ## @type: string
-  ## Location of the database
-  #
-  url: jdbc:h2:mem:anchor-platform
+
+  ## The hostname and port of the database server.
+  server:
+
+  ## Name of the database.
+  database:
```
JakeUrban, line 493: What do you think about using `dbName` here?

JakeUrban: Not a big deal, just an idea. Its not totally clear that you're looking for the database name when I use the `DATA_DATABASE` environment variable.

lijamie98: from Postgres doc: https://jdbc.postgresql.org/documentation/use/

They use `database` as the database name. I think we can keep database unless you feel strong about this.

JakeUrban: Nope, thats fine.

PR #761, opened 2023-02-16
[ANCHOR-174] Stop creating SEP service beans when disabled in configuration. (fix Sep10 npe)

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This is to fix:
```
When sep10.enabled is false or omitted, and secret.sep10.signing_seed / SECRET_SEP10_SIGNING_SEED is omitted, a null pointer exception is raised. I noticed that if I enable SEP-10, but omit the signing seed, this results in a user-friendly error message as expected.

The signing seed should not be required if SEP-10 is not enabled.
```
The bug is caused by creating Sep10Service bean when `sep10.enabled` is set to false. 

This fix avoids creating SEP service beans when the configuration is disabled. 

### Why

Bug fixes.

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
@@ -17,6 +17,7 @@ import org.stellar.anchor.config.*;
 import org.stellar.anchor.event.EventService;
 import org.stellar.anchor.filter.JwtTokenFilter;
 import org.stellar.anchor.horizon.Horizon;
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
Was `ConditionalOnAllSepsEnabled` already implemented?

lijamie98:
> Was `ConditionalOnAllSepsEnabled` already implemented?

Yes, it is implemented [here](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/condition/OnAllSepEnabledCondition.java).

PR #771, opened 2023-02-22
Fix database migration file of kyc_verified field

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix database migration file of kyc_verified field

### Why
Bug fix

### Known limitations

N/A

### DIFF
--- a/platform/src/main/resources/db/migration/V6__sep24_field_updates.sql
+++ b/platform/src/main/resources/db/migration/V6__sep24_field_updates.sql
@@ -2,7 +2,7 @@ ALTER TABLE sep24_transaction ADD sep10_account_memo VARCHAR(255);
 
 ALTER TABLE sep24_transaction ADD amount_expected VARCHAR(255);
 
-ALTER TABLE sep24_transaction ADD kyc_verified BOOLEAN SET default FALSE;
+ALTER TABLE sep24_transaction ADD kyc_verified BOOLEAN;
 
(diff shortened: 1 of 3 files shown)


platform/src/main/resources/db/migration/V6__sep24_field_updates.sql
```diff
@@ -2,7 +2,7 @@ ALTER TABLE sep24_transaction ADD sep10_account_memo VARCHAR(255);
 
 ALTER TABLE sep24_transaction ADD amount_expected VARCHAR(255);
 
-ALTER TABLE sep24_transaction ADD kyc_verified BOOLEAN SET default FALSE;
+ALTER TABLE sep24_transaction ADD kyc_verified BOOLEAN;
```
JakeUrban, line 5: I think we need `ADD COLUMN`, not just `ADD`, right?

JakeUrban: Actually it looks like this may be optional.

JakeUrban (approved):

Ifropc (approved):

platform/src/main/resources/db/migration/V6__sep24_field_updates.sql
lijamie98: `ADD COLUMN` is optional.

PR #776, opened 2023-02-23
[ANCHOR-183] Fix sep10account_memo database field and lock the database column names

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix sep10account_memo database field and lock the database column names.

### Why
Bug fixes.

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
@@ -116,12 +116,12 @@ public class JdbcSep24Transaction extends JdbcSepTransaction
   String requestAssetIssuer;
 
   /** The SEP10 account used for authentication. */
-  @SerializedName("sep10account")
+  @SerializedName("sep10_account")
   @Column(name = "sep10account")
   String sep10Account;
 
   /** The SEP10 account memo used for authentication. */
-  @SerializedName("sep10account_memo")
+  @SerializedName("sep10_account_memo")
   @Column(name = "sep10account_memo")
   String sep10AccountMemo;
 
--- a/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
@@ -30,15 +30,19 @@ public class JdbcSep24Transaction extends JdbcSepTransaction
   String kind;
 
   @SerializedName("status_eta")
+  @Column(name = "status_eta")
   String statusEta;
 
   @SerializedName("kyc_verified")
+  @Column(name = "kyc_verified")
   Boolean kycVerified;
 
   @SerializedName("more_info_url")
+  @Column(name = "more_info_url")
   String moreInfoUrl;
 
   @SerializedName("transaction_id")
+  @Column(name = "transaction_id")
   String transactionId;
 
   String message;
@@ -67,6 +71,7 @@ public class JdbcSep24Transaction extends JdbcSepTransaction
    * will transfer) their issued asset to.
    */
   @SerializedName("withdraw_anchor_account")
+  @Column(name = "withdraw_anchor_account")
   String withdrawAnchorAccount;
 
   /** The memo for deposit or withdraw */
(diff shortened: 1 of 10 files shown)


platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
```diff
@@ -30,15 +30,19 @@ public String getProtocol() {
   String kind;
 
   @SerializedName("status_eta")
+  @Column(name = "status_eta")
```
Ifropc, line 33: Are all this `@Column` annotations necessary?

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
```diff
@@ -97,34 +104,44 @@ public void setRefunds(Sep24Refunds refunds) {
...
 
   @SerializedName("request_asset_issuer")
+  @Column(name = "request_asset_issuer")
   String requestAssetIssuer;
 
   /** The SEP10 account used for authentication. */
-  @SerializedName("sep10_account")
+  @SerializedName("sep10account")
+  @Column(name = "sep10account")
   String sep10Account;
```
JakeUrban, line 121: The serialized name still needs to be `sep10_account` right? Same for the serialized name of the SEP-10 memo?

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
lijamie98: It is not necessary. But it is to lock the column name caused by the inconsistency between tools and Spring. 

There are pros and cons of specifying the column name. I personally don't like to specify the column name just because of a bug and being paranoid of tool bugs. 

WDYT?

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
lijamie98: Unfortunately, Spring maps to `sep10Account` to `sep10account` and `sep10AccountMemo` to `sep10account_memo`. That's why the JPA Buddy has this strange bug. 

I am locking the name so that JPA Buddy will be consistent with Spring.

Ifropc: I think `SerializedName` shouldn't be changed

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
Ifropc: I'd prefer to only do it when necessary, otherwise it's too verbose in my opinion

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
JakeUrban: I agree, unless `SerializedName` needs to match `Column`, I would keep `SerializedName` unchanged.

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
JakeUrban: I'm confused, isn't using `@Column` the only way to ensure Spring uses the correct DB column name? Are there alternatives?

Ifropc: Spring has an automatic conversion that's been used before. I guess in rare cases it doesn't match to expected naming

JakeUrban: Ok, then lets lock in the column name. Not ideal but sounds necessary.

lijamie98: locking is verbose but safe. We do not expect too many Entity classes and they are quite stable. I think we can lock it unless we have other strong reasons in the future.

JakeUrban (approved):

PR #804, opened 2023-03-29
Run Stellar Anchor Test in GH actions

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]


.github/workflows/sub_stellar_anchor_tests.yml
```diff
@@ -20,9 +20,7 @@ jobs:
       - name: Find Home Domain (preview or develop or main)
         id: endpoint-finder
         run: |         
-          if [[ ! -z "$PR_NUMBER" ]]; then
-              export HOME_DOMAIN=https://anchor-sep-pr$PR_NUMBER.previews.kube001.services.stellar-ops.com
```
Ifropc, line 24: No pr previews anymore?

.run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml
Comment (Ifropc) [inline .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: How should we rub reference server without example config?

.run/v2.0 - End2End Test - no Docker.run.xml
Ifropc, line 1: Should I be able to run E2E tests using this configuration file?

.github/workflows/sub_stellar_anchor_tests.yml
lijamie98: Nope. All are in the GH actions.

.run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml
Comment (lijamie98) [reply .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: They are replaced by the test profiles. `V2.0 - Run All Servers - with Docker` and `V2.0 - Run All Servers - no Docker`

Comment (Ifropc) [reply .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: Is config used in this test profiles the same as was in the example config?

.run/v2.0 - End2End Test - no Docker.run.xml
lijamie98: Yes. It is easier to run with `V2.0 - End2End Test - with Docker`. 

I did not test in NixOs. But I think `docker-compose-rule` should work on NixOs.

Paulo will be checking how these runs on MacOS

Ifropc: @lijamie98 it doesn't work with current structure.
```
Reason=Unable to connect to database. url=jdbc:postgresql://db:5432/postgres
```
Unless I [update hosts](https://github.com/stellar/java-stellar-anchor-sdk/blob/26b54ff217502beea000727f6f046cd6ed100010/.github/workflows/sub_gradle_test_and_build.yml#L18) I don't think it will work

lijamie98: Yes, the problem is described at: https://github.com/stellar/java-stellar-anchor-sdk/blob/feature/add-stellar-anchor-test-to-gh-action/.github/workflows/sub_gradle_test_and_build.yml#L17

https://stellarorg.atlassian.net/browse/ANCHOR-229 is created for improvement.

.run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml
Comment (lijamie98) [reply .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: There are two profiles, `default` and `sep24`. We can create another Run Configuration for `sep24` which should be similar to the example config.

Eventually, they should be merged if the Java reference server is deprecated.

Comment (Ifropc) [reply .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: Sounds good

Ifropc:
@lijamie98 can you add into README or somewhere that for non-standard docker location this env variables should be defined, for example:
```bash
DOCKER_COMPOSE_LOCATION = "/run/current-system/sw/bin/docker-compose";
DOCKER_LOCATION = "/run/current-system/sw/bin/docker";
```

Ifropc:
@lijamie98 can you add a helper script `/bin/stop-docker.sh`
```bash
#!/usr/bin/env bash
docker stop $(docker ps -aq)
docker rm $(docker ps -aq)
docker network prune
```

.run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml
Comment (lijamie98) [reply .run/v2.0 - Platform Server - ServiceRunner [Example config].run.xml:1]: The `Run All` configuration runs the reference server as well. I can add configurations that may run individual servers based on the test profiles in the next PRs after the release `2.0.0-rc1` is cut.

lijamie98:
> ```shell
> #!/usr/bin/env bash
> docker stop $(docker ps -aq)
> docker rm $(docker ps -aq)
> docker network prune
> ```

Script is added.

Ifropc (approved):

PR #808, opened 2023-04-03
[ANCHOR-228] SEP-24 withdrawal e2e tests

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

SEP-24 Withdrawal end-2-end tests.

### Why

Preparation for `2.0.0-rc1`

### Known limitations

N/A

### DIFF
--- a/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
+++ b/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
@@ -15,9 +15,13 @@ import org.stellar.walletsdk.ApplicationConfiguration
 import org.stellar.walletsdk.StellarConfiguration
 import org.stellar.walletsdk.Wallet
 import org.stellar.walletsdk.anchor.TransactionStatus
+import org.stellar.walletsdk.anchor.TransactionStatus.*
+import org.stellar.walletsdk.anchor.WithdrawalTransaction
 import org.stellar.walletsdk.asset.IssuedAssetId
 import org.stellar.walletsdk.auth.AuthToken
 import org.stellar.walletsdk.horizon.SigningKeyPair
+import org.stellar.walletsdk.horizon.sign
+import org.stellar.walletsdk.horizon.transaction.toTransferTransaction
 
 class Sep24End2EndTest(
   private val config: TestConfig,
@@ -40,7 +44,6 @@ class Sep24End2EndTest(
       }
     }
   private val anchor = wallet.anchor(config.env["anchor.domain"]!!.substring("http://".length))
-
   private val maxTries = 40
 
   private fun `typical deposit end-to-end flow`() = runBlocking {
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
```diff
@@ -15,46 +15,80 @@ import org.stellar.walletsdk.ApplicationConfiguration
...
     print("accessing ${deposit.url}...")
     assertEquals(200, resp.status.value)
-    waitStatus(deposit.id, TransactionStatus.COMPLETED, token)
+    // Wait for the status to change to COMPLETED
+    waitStatus(deposit.id, COMPLETED, token)
+  }
+
+  private fun `typical withdraw end-to-end flow`() = runBlocking {
+    val token = anchor.auth().authenticate(keypair)
+    // TODO: Add the test where the amount is not specified
```
Ifropc, line 68: You should add test endpoint or test function to be called from here (to provide amount) on the Ref server to support that

kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/Sep24TestRoute.kt
```diff
@@ -40,19 +39,21 @@ fun Route.testSep24(
       }
 
       val transaction = sep24.getTransaction(transactionId)
-      val amountIn =
-        transaction.amountExpected?.amount?.toBigDecimal()
-          ?: throw ClientException("Missing amountExpected.amount field")
-
+      var amountExpected = transaction.amountExpected?.amount?.toBigDecimal()
+      if (amountExpected == null) {
```
Ifropc, line 43: I don't think it's used anywhere right now? Amount is always passed in tests

Ifropc:
@lijamie98 code is not formatted. Can you fix github actions so the build fails?

Ifropc (changes requested):

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
lijamie98: The plan is to change them to parameterized tests and pass the amount here. The current design is to support running the tests from CLI. But it looks like we should run the tests as Junit tests from the CLI. 

https://stellarorg.atlassian.net/browse/ANCHOR-230

kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/Sep24TestRoute.kt
lijamie98: This is to support the use case when the `amount` is not specified in POST transactions. The test case will be added in `Sep24End2EndTests.kt`.

Ifropc: Can you remove it from current iteration if the test is not there yet

lijamie98: I think we should handle the case when the amount expected is not specified because the `amount` field is optional in SEP-24. https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request-1

When the amount is not specified, we should give a number to simulate the interactive flow.

lijamie98:
> @lijamie98 code is not formatted. Can you fix github actions so the build fails?

Reformatted.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/Sep24TestRoute.kt
Ifropc: I agree. But I don't think the way it's done now is good. Ideally there should be a function on ref server that you call from your test code. Just hardcoding it doesn't sounds like a good idea to me. 
Can you remove this change because it's out of scope of this PR?

Ifropc (approved):

paulormnas (approved):

PR #824, opened 2023-04-07
fix: getenv so it can handle line breaks `/n` and quotes `/"` when reading System envs

marcelosalloum:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix the getenv method so it can handle line breaks /n and quotes /" when reading System envs

### Why

When providing a multiline env var with `\n`, it was being ingested as `\\n`. The same for quotes.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigEnvironment.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigEnvironment.java
@@ -46,7 +46,11 @@ public class ConfigEnvironment {
    * @return the value of the environment variable.
    */
   public static String getenv(String name) {
-    return env.get(toPosixForm(name));
+    String envValue = env.get(toPosixForm(name));
+    if (envValue != null) {
+      envValue = envValue.replace("\\n", "\n").replace("\\\"", "\"");
+    }
+    return envValue;
   }
 
   public static Collection<String> names() {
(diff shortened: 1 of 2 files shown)


platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigEnvironment.java
```diff
@@ -46,7 +46,11 @@ public static void rebuild() {
    * @return the value of the environment variable.
    */
   public static String getenv(String name) {
-    return env.get(toPosixForm(name));
+    String envValue = env.get(toPosixForm(name));
+    if (envValue != null) {
+      envValue = envValue.replace("\\n", "\n").replace("\\\"", "\"");
```
lijamie98, line 51: I am a little worried about the corner cases that we did not cover such as "\\t" for tab.

Can we use [`StringEscapeUtils` from Apache](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringEscapeUtils.html#unescapeJava-java.lang.String-)?

marcelosalloum: Great suggestion, thank you! Done in https://github.com/stellar/java-stellar-anchor-sdk/pull/824/commits/4f2cdfa9a36837d7d16ca34a7b36bd7ca630c5d6.

lijamie98 (approved):
LGTM

PR #833, opened 2023-04-11
[ANCHOR-212] Skip setting amountExpected when posting SEP24 transactions

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Skip setting amountExpected when posting SEP24 transactions

### Why

The `amountExpected` should be set by the business server through the `PATCH /transaction` API.


Ifropc:
@lijamie98 is it still set in JWT token? 
FYI: it should be always set, if it's present in the request (regardless of configuration)

paulormnas (approved):

JakeUrban:
The JWT has a `data` object that should have any fields passed in the `POST /transactions/deposit/interactive` or `POST /transactions/withdraw/interactive` request bodies that shouldn't be persisted in the anchor platform's database. Right now this includes any of the SEP-9 KYC fields, but it could also include the amount field. 

I think if we're not saving it to `amountExpected` automatically, we need to pass it to the business somehow, and using the JWT makes sense.

lijamie98:
> 

@JakeUrban I think this make sense now. Is `amount` the only field that we should add to JWT? If there other fields in the request that are not saved in database but should be passed to JWT, we can add them all at once.

Here is the list that I can think of. Please comment:

- transaction_id
- all `SEP-9` fields
- `amount`

Ifropc (approved):

PR #834, opened 2023-04-11
[ANCHOR-245] Support spring.flyway.baseline-on-migrate

lijamie98:
https://stellarorg.atlassian.net/browse/ANCHOR-270
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Support spring.flyway.baseline-on-migrate

### Why

To support running AP on the same database as other applications.


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -534,16 +534,14 @@ data:
...
-  ## @param: hikari_max_lifetime
-  ## @type: integer
-  ## Maximum connection time before expiration
-  ## Only valid when database `type` is `aurora`.
-  ## Recommended setting is 14 minutes because IAM tokens are valid for 15 min.
+  ## @param: flyway_baseline_on_migrate
+  ## @type: bool
+  ## Whether to enable flyway baseline on migrate. This is useful when you want to run flyway on an existing database.
   #
-  # hikari_max_lifetime: 840000
```
marcelosalloum, line 548: Are you sure you wanted to remove hikari_max_lifetime? Just as a sanity check here.

marcelosalloum (commented):
I'd like to hold on this PR for now, until I'm able to run some tests in localhost and make sure this is what SDP needs.

I'll get back to you later this week.

marcelosalloum:
> Hi @marcelosalloum were you able to verify this on the SDP side?

I still couldn't do it. I will only be able to check this after Thu (May 4yh).

platform/src/main/java/org/stellar/anchor/platform/configurator/DataConfigAdapter.java
```diff
@@ -131,6 +131,9 @@ void updateSpringEnv(ConfigMap config) throws InvalidConfigException {
           set("spring.flyway.user", SecretManager.getInstance().get(SECRET_DATA_USERNAME));
           set("spring.flyway.password", SecretManager.getInstance().get(SECRET_DATA_PASSWORD));
           set("spring.flyway.url", constructPostgressUrl(config));
+          set(
+              "spring.flyway.baseline-on-migrate",
+              config.getString("data.flyway_baseline_on_migrate").equalsIgnoreCase("true"));
```
marcelosalloum, line 136: 🚨  I've tested it locally and got an error where the [V1_master](https://github.com/stellar/java-stellar-anchor-sdk/blob/6b57d67d9138b9cde78bbd1d8750e356e53a3dc6/platform/src/main/resources/db/migration/V1__master.sql) migration was ignored.

To fix that, you additionally need to tell Flyway what version should be given to the baseline. The official documentation suggests setting it to 0: https://flywaydb.org/documentation/configuration/parameters/baselineVersion

According to [the documentation](https://flywaydb.org/documentation/configuration/parameters/baselineOnMigrate), _"Only migrations above `baselineVersion` will then be applied."_.

marcelosalloum: Here's the fix that worked for me:
```suggestion
          boolean baselineOnMigrate =
              config.getString("data.flyway_baseline_on_migrate").equalsIgnoreCase("true");
          if (baselineOnMigrate) {
            set("spring.flyway.baseline-on-migrate", true);
            set("spring.flyway.baseline-version", "0");
          }
```

It should be fine to default the baseline-version to `0` when baseline-on-migrate is present.

platform/src/main/java/org/stellar/anchor/platform/configurator/DataConfigAdapter.java
```diff
@@ -149,6 +152,9 @@ void updateSpringEnv(ConfigMap config) throws InvalidConfigException {
...
```
marcelosalloum, line 157: It should also be fixed here:

```suggestion
          boolean baselineOnMigrate =
              config.getString("data.flyway_baseline_on_migrate").equalsIgnoreCase("true");
          if (baselineOnMigrate) {
            set("spring.flyway.baseline-on-migrate", true);
            set("spring.flyway.baseline-version", "0");
          }
```

marcelosalloum (commented):
Thanks for working on that.

The `baseline-version` configuration is missing (as explained below).
After you add that the PR should be good to be merged.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: The Hikari configurations are set in the [`DataConfigAdapter`](https://github.com/stellar/java-stellar-anchor-sdk/pull/834/files#diff-286b8e76038fec752de1d40ba945fa30fc32348e487d6afe5fa90c4a2626d21eR250)

I have taken a closer look at the default values that would probably make the system stabler.

marcelosalloum:
This is looking good! I'll test it locally and approve it after I've tested it with the SDP.

Thanks!

marcelosalloum (approved):
LGTM!

PR #839, opened 2023-04-19
Fix stellar transaction id is not set properly

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Fix stellar_transaction_id is not set in DB 

### Why

Due to this bug stellar_transaction_id is not set in the result GET transaction response. Also, it's not possible to query by stellar_transaction_id.

### Known limitations

Only works with the latest transaction's id. If multiple Stellar transactions are associated with the SEP transaction, only latest Stellar transaction will be visible.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
@@ -13,10 +13,7 @@ import static org.stellar.sdk.xdr.MemoType.MEMO_HASH;
 
 import io.micrometer.core.instrument.Metrics;
 import java.time.Instant;
-import java.util.Base64;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 import org.apache.commons.lang3.StringUtils;
 import org.stellar.anchor.api.exception.AnchorException;
 import org.stellar.anchor.api.exception.BadRequestException;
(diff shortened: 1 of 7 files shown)


platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -145,6 +142,13 @@ private GetTransactionResponse patchTransaction(PatchTransactionRequest patch)
           sep24Transaction.setMemo(memo);
           sep24Transaction.setMemoType(memoTypeAsString(MEMO_HASH));
         }
+        if (patch.getTransaction().getStellarTransactions() != null) {
+          patch.getTransaction().getStellarTransactions().stream()
+              .max(Comparator.comparingLong(x -> x.getCreatedAt().toEpochMilli()))
+              .ifPresent(
+                  stellarTransaction ->
+                      sep24Transaction.setStellarTransactionId(stellarTransaction.getId()));
+        }
```
JakeUrban, line 151: Do we not want to do this for SEP-31 transaction as well? Isn't this doing what is supposed to be happening [here](https://github.com/stellar/java-stellar-anchor-sdk/blob/d616253c0157b0bc9b67e16cb99811e09a1bd33a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java#L200)?

Ifropc: We probably do

JakeUrban (approved):

PR #847, opened 2023-04-25
Cleanup transaction schema

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

1. Remove kycVerified 
2. Update transferReceivedAt only for SEP-31
3. Remove custodialTransactionId

### Why

1. Not used 
2. Was only used by SEP-31
3. Not used

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
@@ -33,9 +33,6 @@ public class PlatformTransactionData {
   @SerializedName("amount_fee")
   Amount amountFee;
 
-  @SerializedName("kyc_verified")
-  Boolean kycVerified;
-
   @SerializedName("quote_id")
   String quoteId;
 
@@ -65,10 +62,6 @@ public class PlatformTransactionData {
 
   @SerializedName("external_transaction_id")
   String externalTransactionId;
-
-  // TODO: review the use case of the custodialTransactionId
-  @SerializedName("custodial_transaction_id")
-  String custodialTransactionId;
 
   @SerializedName("memo")
   String memo;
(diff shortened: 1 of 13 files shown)


platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -230,6 +228,8 @@ void updateSepTransaction(PlatformTransactionData patch, JdbcSepTransaction txn)
         }
         break;
       case "31":
+        txnUpdated = updateField(patch, txn, "transferReceivedAt", txnUpdated);
+
```
JakeUrban, line 232: Do we not want to update this field in SEP-24 withdrawals?

api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
```diff
@@ -66,10 +63,6 @@ public class PlatformTransactionData {
   @SerializedName("external_transaction_id")
   String externalTransactionId;
 
-  // TODO: review the use case of the custodialTransactionId
-  @SerializedName("custodial_transaction_id")
-  String custodialTransactionId;
```
JakeUrban, line 71: I don't see this field being removed in a migration file. Did it not exist in the first place?

JakeUrban (commented):
Was the V6 migration file added for the 2.0 release or did it exist for 1.x releases? If it existed in 1.x releases we should make a new migration file for removing the column instead of deleting the add statement.

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
Ifropc: It's not returned in GET response of SEP-24, so I don't think it makes sense to update it at all

api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
Ifropc: It's never been a part of initial SQL script

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
JakeUrban: It should be in the Platform API's transaction response though

Ifropc:
> Was the V6 migration file added for the 2.0 release or did it exist for 1.x releases? If it existed in 1.x releases we should make a new migration file for removing the column instead of deleting the add statement.

It's for v2

JakeUrban (approved):

PR #866, opened 2023-05-02
[ANCHOR-265] Implement GET /transactions Platform API endpoint

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Implement GET /transactions Platform API endpoint

### Why

New feature

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep24/Sep24TransactionStore.java
```diff
@@ -60,4 +63,21 @@ List<Sep24Transaction> findTransactions(
...
+   * Finds multiple transactions that matches following criteria
+   *
+   * @param to upper date to search to
+   * @param from lower date to search from
+   * @param orderBy date type that used for the search
+   * @param limit limit
+   * @param offset offset
+   * @return list of transactions
+   */
+  List<? extends Sep24Transaction> findBulk(
```
JakeUrban, line 77: Should we use the `findTransactions()` function already defined instead of `findBulk()`?

platform/src/main/java/org/stellar/anchor/platform/controller/sep/PlatformController.java
```diff
@@ -45,7 +45,16 @@ public PatchTransactionsResponse patchTransactions(@RequestBody PatchTransaction
...
       value = "/transactions",
       method = {RequestMethod.GET})
-  public GetTransactionResponse getTransactions() throws AnchorException {
-    throw new NotFoundException("Not implemented");
+  public GetTransactionsResponse getTransactions(
+      @RequestParam(value = "sep") TransactionsSeps sep,
+      @RequestParam(required = false, value = "order_by", defaultValue = "created_at")
+          TransactionsOrderBy order_by,
+      @RequestParam(required = false, value = "order", defaultValue = "asc") Sort.Direction order,
+      @RequestParam(required = false, value = "statuses") List<SepTransactionStatus> statuses,
```
JakeUrban, line 53: Can you think of a concrete use case where the business would need to get transactions in multiple statuses? At that point, why not just get all transactions and filter on the client side?

platform/src/main/java/org/stellar/anchor/platform/controller/sep/PlatformController.java
```diff
@@ -45,7 +45,16 @@ public PatchTransactionsResponse patchTransactions(@RequestBody PatchTransaction
...
+      @RequestParam(required = false, value = "limit", defaultValue = "200") Integer limit,
```
JakeUrban, line 54: This should be smaller, say 20?

platform/src/main/java/org/stellar/anchor/platform/controller/sep/PlatformController.java
```diff
@@ -45,7 +45,16 @@ public PatchTransactionsResponse patchTransactions(@RequestBody PatchTransaction
...
+      @RequestParam(required = false, value = "cursor", defaultValue = "0") Integer cursor)
+      throws AnchorException {
+    return transactionService.getTransactionsResponse(
```
JakeUrban, line 57: I think you're passing in `limit` for `pageNumber` and `cursor` for `pageSize`, right?

Also, I'm not sure we want to call our page number `cursor`. It will be easier to understand if we just call it `page`, wdyt?

Ifropc: Yeah i forgot to rename parameters. 
page_number and page_size should be good, right?

platform/src/main/java/org/stellar/anchor/platform/controller/sep/PlatformController.java
Ifropc: Statuses is optional. The behavior you describing can be achieved with not providing statuses at all.
The business may want to filter orders in multiple statuses to expire them, for example. 
If they want to expire both pending_user_transfer_start and incomplete after certain amount without any action

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep31TransactionStore.java
```diff
@@ -79,4 +85,32 @@ public JdbcSep31Transaction findByStellarAccountIdAndMemo(String accountId, Stri
...
+      int pageNumber,
+      int pageSize) {
+
+    Sort.Order primary =
+        order == Sort.Direction.ASC
+            ? Sort.Order.asc(orderBy.getTableName()).nullsLast()
+            : Sort.Order.desc(orderBy.getTableName()).nullsLast();
+    Sort sort = Sort.by(List.of(primary, Sort.Order.asc("id")));
+
+    PageRequest page = PageRequest.of(pageNumber, pageSize, sort);
```
JakeUrban, line 103: Why is this logic different from the SEP-24 method?

JakeUrban (approved):

platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep31TransactionStore.java
Ifropc: I was just testing it out.
(Spoiler: doesn't work cause of https://github.com/spring-projects/spring-data-jpa/issues/1280 )

PR #867, opened 2023-05-03
[ANCHOR-205] Validate PATCH transactions request body.

paulormnas:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add NPE catchs on sending PATCH transactions request

### Why

Return a understandable error response to the user instead of only Null exception error stack trace

### Known limitations

The PATCH transactions endpoint will need more attention to handle failures of one or more transactions inside a list of transactions. Today the TransactionService returns a BadRequestException as soon as the first failure occurs, even if other transactions were already committed.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
@@ -131,16 +131,25 @@ public class TransactionService {
   public PatchTransactionsResponse patchTransactions(PatchTransactionsRequest request)
       throws AnchorException {
     List<PatchTransactionRequest> patchRequests = request.getRecords();
+    if (patchRequests == null) {
+      throw new BadRequestException("Records are missing.");
+    }
 
     List<GetTransactionResponse> txnResponses = new LinkedList<>();
+
     for (PatchTransactionRequest patchRequest : patchRequests) {
       txnResponses.add(patchTransaction(patchRequest));
     }
+
     return new PatchTransactionsResponse(txnResponses);
   }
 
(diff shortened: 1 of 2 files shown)


LuiN55 (approved):

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -131,17 +131,27 @@ public JdbcSepTransaction findTransaction(String txnId) throws AnchorException {
   public PatchTransactionsResponse patchTransactions(PatchTransactionsRequest request)
       throws AnchorException {
     List<PatchTransactionRequest> patchRequests = request.getRecords();
-
     List<GetTransactionResponse> txnResponses = new LinkedList<>();
-    for (PatchTransactionRequest patchRequest : patchRequests) {
-      txnResponses.add(patchTransaction(patchRequest));
+
+    try {
```
Ifropc, line 136: I think body should be validated instead of catching NPE
With body validation, proper message can be created "Missing field X in the request"

paulormnas: Yeah, I talked to @lijamie98 about that and he asked me to validate in the service and not in the controller, that is why I did it that way. Also is better to test.

But I can do something like.
```
if(patchRequests == null){
  throw BadRequestException();
}
```

For me is pretty much the same. 
WDYT?

paulormnas: Updated

lijamie98 (approved):
LGTM

lijamie98 (approved):
LGTM

PR #891, opened 2023-05-18
SDFSOF-25 - Added transaction reconciliation job logic

MazurakIhor:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Added transaction reconciliation job logic

### Why

Reconciliation job should update transaction status in case if platform missed or failed to process webhook message

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/custody/fireblocks/TransactionStatus.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/custody/fireblocks/TransactionStatus.java
@@ -1,5 +1,7 @@
 package org.stellar.anchor.api.custody.fireblocks;
 
+import java.util.Set;
+
 public enum TransactionStatus {
   SUBMITTED,
   QUEUED,
@@ -15,5 +17,13 @@ public enum TransactionStatus {
   CANCELLED,
   REJECTED,
   BLOCKED,
-  FAILED
+  FAILED;
+
+  public boolean isSuccessStatus() {
+    return this.equals(COMPLETED);
+  }
+
+  public boolean isObservableStatus() {
+    return Set.of(FAILED, CANCELLED, BLOCKED, COMPLETED).contains(this);
+  }
 }
--- a/platform/src/main/java/org/stellar/anchor/platform/config/FireblocksConfig.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/config/FireblocksConfig.java
@@ -33,9 +33,9 @@ public class FireblocksConfig implements Validator {
   private String baseUrl;
   private String vaultAccountId;
   private CustodySecretConfig secretConfig;
-  private String transactionsReconciliationCron;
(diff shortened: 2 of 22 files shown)


platform/src/main/java/org/stellar/anchor/platform/config/FireblocksConfig.java
```diff
@@ -46,6 +46,10 @@ public void setRetryConfig(RetryConfig retryConfig) {
     this.retryConfig = retryConfig;
   }
 
+  public void setReconciliation(Reconciliation reconciliation) {
```
rkharevych, line 49: It should work without explicit setter, the same for `retryConfig`

api-schema/src/main/java/org/stellar/anchor/api/custody/fireblocks/TransactionStatus.java
```diff
@@ -15,5 +17,13 @@ public enum TransactionStatus {
...
   REJECTED,
   BLOCKED,
-  FAILED
+  FAILED;
+
+  public boolean isSuccessStatus() {
+    return this.equals(COMPLETED);
+  }
+
+  public boolean isObservableStatus() {
```
rkharevych, line 26: I think, that we can remove `Status` ending from both methods, since Enum has it in it's name

platform/src/main/java/org/stellar/anchor/platform/data/JdbcCustodyTransaction.java
```diff
@@ -84,4 +84,8 @@ public class JdbcCustodyTransaction {
   @SerializedName("kind")
   @Column(name = "kind")
   String kind;
+
+  @SerializedName("attempt_count")
+  @Column(name = "attempt_count")
+  int attemptCount;
```
rkharevych, line 90: What do you think abount `reconciliationAttemptCount`? Because it's not clear which count it is

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
```diff
@@ -136,4 +138,22 @@ class FireblocksPaymentServiceTest {
...
+    every { fireblocksClient.get(capture(urlCapture)) } returns responseJson
+
+    val response = fireblocksPaymentService.getTransactionById(EXTERNAL_TXN_ID)
+
+    Assertions.assertEquals(
+      String.format("/v1/transactions/%s", EXTERNAL_TXN_ID),
+      urlCapture.captured
+    )
+
+    assertNotNull(response)
```
rkharevych, line 157: You can also double-check response here, in case service modifies response etc

rkharevych (changes requested):

rkharevych (approved):

PR #895, opened 2023-05-18
Fix tx_bad_seq error caused at Sep24End2EndTest.kt::`withdraw end-to-end flow()`

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

#### Root cause: 
The Kotlin reference server is using the same account to send deposit money [java-stellar-anchor-sdk/default-config.yaml at develop · stellar/java-stellar-anchor-sdk](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/kotlin-reference-server/src/main/resources/default-config.yaml#L9) as the withdraw flow.

When both transactions are submitted to the same block, the race condition causes tx_bad_seq error.

#### Fix:
The reference server uses the same account as the CLIENT_WALLET_KEY. We should have another account dedicated for the Kotlin reference server.

### DIFF
--- a/kotlin-reference-server/src/main/resources/default-config.yaml
+++ b/kotlin-reference-server/src/main/resources/default-config.yaml
@@ -5,8 +5,9 @@ sep24:
   anchorPlatformUrl: "http://localhost:8080"
   #  URL for horizon server
   horizonUrl: "https://horizon-testnet.stellar.org"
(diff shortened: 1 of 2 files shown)


lijamie98 (approved):
LGTM.

PR #906, opened 2023-05-26
[ANCHOR-301] Add JWT auth for the platform server endpoints

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

1. Add platform server authentication configuration.
2. Re-enable Platform Server JwtAuth

### Why
To secure platform server endpoints.


core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -102,6 +102,21 @@ public String encode(Sep24MoreInfoUrlJwt token) throws InvalidConfigException {
...
+          "Please provide the secret before encoding JWT for API Authentication");
+    }
+
+    Calendar calNow = Calendar.getInstance();
+    Calendar calExp = Calendar.getInstance();
+    calExp.setTimeInMillis(calNow.getTimeInMillis() + 1000L * token.getExp());
+    JwtBuilder builder =
+        Jwts.builder().setIssuedAt(calNow.getTime()).setExpiration(calExp.getTime());
+
+    return builder.signWith(SignatureAlgorithm.HS256, platformAuthSecret).compact();
```
philipliu, line 117: Doesn't the secret we use depend on whether `ApiAuthJwt` is for callback or platform?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -136,8 +155,16 @@ public <T extends AbstractJwt> T decode(String cipher, Class<T> cls)
       return (T) Sep10Jwt.class.getConstructor(Jwt.class).newInstance(jwt);
     } else if (cls.equals(Sep24InteractiveUrlJwt.class)) {
       return (T) Sep24InteractiveUrlJwt.class.getConstructor(Jwt.class).newInstance(jwt);
-    } else {
+    } else if (cls.equals(Sep24MoreInfoUrlJwt.class)) {
       return (T) Sep24MoreInfoUrlJwt.class.getConstructor(Jwt.class).newInstance(jwt);
+    } else if (cls.equals(PlatformAuthJwt.class)) {
+      return (T) PlatformAuthJwt.class.getConstructor(Jwt.class).newInstance(jwt);
+    } else {
```
philipliu, line 162: Nit: let's explicitly check that the class is `CallbackAuthJwt` and throwing an `NotSupportedException` in the else case.

core/src/main/java/org/stellar/anchor/filter/AbstractJwtFilter.java
```diff
@@ -9,21 +10,19 @@
...
 import org.stellar.anchor.util.Log;
 
-public class JwtTokenFilter implements Filter {
+public abstract class AbstractJwtFilter implements Filter {
   public static final String JWT_TOKEN = "token";
-  private static final String OPTIONS = "OPTIONS";
-  public static final String APPLICATION_JSON_VALUE = "application/json";
-  private static final Gson gson = GsonUtils.builder().setPrettyPrinting().create();
-  private final JwtService jwtService;
+  static final String OPTIONS = "OPTIONS";
```
philipliu, line 20: Nit: does this need to be a variable? We are only using it in one place, it can be inlined instead.

core/src/main/java/org/stellar/anchor/filter/PlatformAuthJwtFilter.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class PlatformAuthJwtFilter extends AbstractJwtFilter {
+  public PlatformAuthJwtFilter(JwtService jwtService) {
+    super(jwtService);
+  }
+
+  @Override
+  void check(String jwtCipher, HttpServletRequest request, ServletResponse servletResponse)
+      throws Exception {
+    PlatformAuthJwt token = jwtService.decode(jwtCipher, PlatformAuthJwt.class);
```
philipliu, line 16: We check that `token` is not null in `Sep10JwtFilter`. Do we need a check here as well?

core/src/main/java/org/stellar/anchor/filter/Sep10JwtFilter.java
```diff
@@ -0,0 +1,27 @@
...
+
+  @Override
+  void check(String jwtCipher, HttpServletRequest request, ServletResponse servletResponse)
+      throws Exception {
+    Sep10Jwt token = jwtService.decode(jwtCipher, Sep10Jwt.class);
+    if (token == null) {
+      throw new SepValidationException("JwtToken should not be null");
+    }
+    infoF("token created. account={} url={}", shorter(token.getAccount()), request.getRequestURL());
+    debug(String.format("storing token to request %s:", request.getRequestURL()), token);
```
philipliu, line 24: Nit: we can use `debugF` for consistency.

platform/src/main/java/org/stellar/anchor/platform/config/PlatformApiConfig.java
```diff
@@ -35,7 +35,7 @@ public boolean supports(@NotNull Class<?> clazz) {
   @Override
   public void validate(@NotNull Object target, @NotNull Errors errors) {
     if (List.of(AuthType.API_KEY, AuthType.JWT).contains(auth.getType())) {
-      if (isEmpty(secretConfig.getCallbackApiSecret())) {
+      if (isEmpty(secretConfig.getCallbackAuthSecret())) {
```
philipliu, line 38: This should be `getPlatformAuthSecret`?

core/src/main/java/org/stellar/anchor/filter/PlatformAuthJwtFilter.java
Ifropc: No, token is checked later in the flow. Checking it here will disable support for auth token

platform/src/main/java/org/stellar/anchor/platform/config/PlatformApiConfig.java
```diff
@@ -35,7 +35,7 @@ public boolean supports(@NotNull Class<?> clazz) {
   @Override
   public void validate(@NotNull Object target, @NotNull Errors errors) {
```
Ifropc, line 36: Can we add an extra validation here: if network set to mainnet, throw an error if auth type is set to NONE?
This will prevent unsecured endpoints in production/staging envs

service-runner/src/main/kotlin/org/stellar/anchor/platform/TestProfileRunner.kt
```diff
@@ -60,6 +60,7 @@ class TestProfileExecutor(val config: TestConfig) {
     shouldStartKotlinReferenceServer = config.env["run_kotlin_reference_server"].toBoolean()
 
     startDocker()
+    // TODO: Check server readiness instead of wait for 5 seconds
```
Ifropc, line 63: Can you add a Jira ticket for that?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    @JvmStatic
+    fun breakdown() {
+      testProfileRunner.shutdown()
+    }
+  }
+
+  private val jwtPlatformClient: PlatformApiClient =
+    PlatformApiClient(jwtAuthHelper, "http://localhost:8085")
+
+  // TODO - to be deprecated by platformAPI client
```
Ifropc, line 62: Can usage of nonAuthPlatformClient be added before this PR is merged?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    OkHttpClient.Builder()
+      .connectTimeout(10, TimeUnit.MINUTES)
+      .readTimeout(10, TimeUnit.MINUTES)
+      .writeTimeout(10, TimeUnit.MINUTES)
+      .build()
+
+  @ParameterizedTest
+  @CsvSource(
+    value =
+      [
```
Ifropc, line 73: Can you make this list as a constant somewhere?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        "GET,/transactions",
+        "PATCH,/transactions",
+        "GET,/transactions/my_id",
+        "GET,/exchange/quotes",
+        "GET,/exchange/quotes/id"
+      ]
+  )
+  fun `test JWT protection of the platform server`(method: String, endpoint: String) {
```
Ifropc, line 99: Can you add test where wrong key is used?

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
Ifropc, line 1: Line 45 and onward: can we remove this logic with `registrationBean`? All platform endpoints should be secured. Otherwise, when a developer adds new endpoint to platform server, this filter can be forgotten to be updated.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/AnchorReferenceConfig.java
```diff
@@ -77,15 +77,13 @@ public FilterRegistrationBean<Filter> platformToAnchorTokenFilter(
...
   @Bean
-  AuthHelper authHelper(AppSettings appSettings, IntegrationAuthSettings integrationAuthSettings) {
+  AuthHelper authHelper(IntegrationAuthSettings integrationAuthSettings) {
     String authSecret = integrationAuthSettings.getAnchorToPlatformSecret();
     switch (integrationAuthSettings.getAuthType()) {
       case JWT:
         return AuthHelper.forJwtToken(
-            new JwtService(authSecret, null, null),
-            integrationAuthSettings.getExpirationMilliseconds(),
-            appSettings.getHostUrl());
```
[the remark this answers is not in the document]
lijamie98: The issuer URL is used when it is published by a web site with an URL. We are using JWT for symmetric authentication and the issuer URL may not be available.

core/src/main/java/org/stellar/anchor/auth/AuthHelper.java
```diff
@@ -40,12 +50,14 @@ public AuthHeader<String, String> createAuthHeader() {
...
         long issuedAt = Calendar.getInstance().getTimeInMillis() / 1000L;
         long expirationTime = issuedAt + (jwtExpirationMilliseconds / 1000L);
-        Sep10Jwt token = Sep10Jwt.of(issuerUrl, issuedAt, expirationTime);
-        return new AuthHeader<>("Authorization", "Bearer " + jwtService.encode(token));
-
+        ApiAuthJwt token = new ApiAuthJwt(issuedAt, expirationTime);
+        try {
+          return new AuthHeader<>("Authorization", "Bearer " + jwtService.encode(token));
+        } catch (InvalidConfigException e) {
+          return null;
```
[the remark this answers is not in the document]
lijamie98: I have removed the `return null` and let the exception bubble up.

core/src/main/java/org/stellar/anchor/filter/AbstractJwtFilter.java
lijamie98: Removed.

core/src/main/java/org/stellar/anchor/auth/JwtService.java
philipliu: We can add a  `@SuppressWarnings({"ConstantConditions"})` to the decode method.

core/src/main/java/org/stellar/anchor/filter/PlatformAuthJwtFilter.java
lijamie98: The decode function never returns `null`. I have changed the variable type to `@NonNull`.

platform/src/main/java/org/stellar/anchor/platform/config/PlatformApiConfig.java
lijamie98: Good catch.

platform/src/main/java/org/stellar/anchor/platform/config/PlatformApiConfig.java
lijamie98: @Ifropc I am not sure if we should add the validation here since the platform server is now separated from the sep server. We should be careful because this disables the anchor to have open endpoints in their private network. If you think this is necessary, let's create a separate thread in Slack to discuss.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
lijamie98: Changed.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
lijamie98: Added.

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
lijamie98: I think this is a very good point. I have removed the url patterns.

core/src/main/java/org/stellar/anchor/auth/JwtService.java
lijamie98: That's probably about the same as as the previous approach.

philipliu (approved):
LGTM

platform/src/main/java/org/stellar/anchor/platform/config/PlatformApiConfig.java
Ifropc: That's a good point. I think we should just print WARN when auth set to NONE instead

integration-tests/src/test/kotlin/org/stellar/anchor/platform/AuthIntegrationTest.kt
Ifropc: FYI: you can make a constant a whole array :)

Ifropc (approved):

PR #907, opened 2023-05-26
[ANCHOR-263] Fix: check for asset min and max amount

paulormnas:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Check for assets min and max before validate request's amount limits

### Why

To avoid NPE when assets min and max are not set

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -110,12 +110,23 @@ public class Sep24Service {
       throw new SepValidationException(String.format("invalid operation for asset %s", assetCode));
     }
 
-    // Validate amount
-    if (strAmount != null) {
-      if (decimal(strAmount).compareTo(decimal(asset.getWithdraw().getMinAmount())) < 0
-          || decimal(strAmount).compareTo(decimal(asset.getWithdraw().getMaxAmount())) > 0) {
+    // Validate min amount
+    Long minAmount = asset.getWithdraw().getMinAmount();
+    if (strAmount != null && minAmount != null) {
+      if (decimal(strAmount).compareTo(decimal(minAmount)) < 0) {
         infoF("invalid amount {}", strAmount);
-        throw new SepValidationException(String.format("invalid amount: %s", strAmount));
+        throw new SepValidationException(
+            String.format("amount is less than asset's minimum limit: %s", strAmount));
+      }
+    }
+
+    // Validate max amount
+    Long maxAmount = asset.getWithdraw().getMaxAmount();
+    if (strAmount != null && maxAmount != null) {
+      if (decimal(strAmount).compareTo(decimal(maxAmount)) > 0) {
+        infoF("invalid amount {}", strAmount);
+        throw new SepValidationException(
+            String.format("amount exceeds asset's maximum limit: %s", strAmount));
       }
     }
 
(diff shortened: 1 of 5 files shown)


JakeUrban (approved):
This looks great, thanks for fixing!

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24Tests.kt
```diff
@@ -320,36 +322,24 @@ private const val expectedSep24Info =
   {
     "deposit": {
       "JPYC": {
-        "enabled": true,
-        "min_amount": 1,
-        "max_amount": 1000000
```
philipliu, line 325: It would be good to keep at least one asset with `min_amount` and `max_amount` set. Same comment applies to other tests below.

philipliu: It's also possible for only one of `min_amount` or `max_amount` to be null, right? Can we include tests cases for that?

paulormnas: Hey @philipliu, thanks for the comments
There are 3 assets configured in the project, and I removed the max and min configs from 2 of them and kept it on the third one.

About the test case, it's a good catch, I didn't think about the possibility. I'll work on that

Ifropc: Good catch, it looks like with current version of code, the case when only one of min/max is set won't work

paulormnas: Updated!
I didn't need to add new tests, because when min or max is missing it don't thrown NPE. But I configured the asset with only max or min to check. 👍🏻

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -238,9 +240,11 @@ public InteractiveTransactionResponse deposit(Sep10Jwt token, Map<String, String
     }
 
     // Validate amount
-    if (strAmount != null) {
-      if (decimal(strAmount).compareTo(decimal(asset.getDeposit().getMinAmount())) < 0
-          || decimal(strAmount).compareTo(decimal(asset.getDeposit().getMaxAmount())) > 0) {
+    Long minAmount = asset.getWithdraw().getMinAmount();
+    Long maxAmount = asset.getWithdraw().getMaxAmount();
+    if (strAmount != null && minAmount != null && maxAmount != null) {
```
lijamie98, line 245: I think we may start adding comments for the logics that can be confusing. 
```suggestion
    // check the amount only when the min and max amounts are defined.
    if (strAmount != null && minAmount != null && maxAmount != null) {
```

core/src/main/java/org/stellar/anchor/util/SepHelper.java
```diff
@@ -118,14 +118,17 @@ public static BigDecimal validateAmount(String messagePrefix, String amount, boo
...
       throws AnchorException {
-    BigDecimal sAmount = validateAmount("", amount);
-    BigDecimal bdMin = new BigDecimal(min);
-    BigDecimal bdMax = new BigDecimal(max);
-    if (sAmount.compareTo(bdMin) < 0) {
-      throw new BadRequestException(String.format("%samount less than min limit", messagePrefix));
-    }
-    if (sAmount.compareTo(bdMax) > 0) {
-      throw new BadRequestException(String.format("%samount exceeds max limit", messagePrefix));
+    if (min != null && max != null) {
```
lijamie98, line 121: I think we may start adding comments for the logics that can be confusing. 
```suggestion
    // since min and max are optional in assets, compare only if defined.
    if (min != null && max != null) {
```

lijamie98 (approved):
Feel free to merge when the comments are taken care of.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -111,9 +111,11 @@ public InteractiveTransactionResponse withdraw(
...
 
     // Validate amount
-    if (strAmount != null) {
-      if (decimal(strAmount).compareTo(decimal(asset.getWithdraw().getMinAmount())) < 0
-          || decimal(strAmount).compareTo(decimal(asset.getWithdraw().getMaxAmount())) > 0) {
+    Long minAmount = asset.getWithdraw().getMinAmount();
+    Long maxAmount = asset.getWithdraw().getMaxAmount();
+    if (strAmount != null && minAmount != null && maxAmount != null) {
+      if (decimal(strAmount).compareTo(decimal(minAmount)) < 0
+          || decimal(strAmount).compareTo(decimal(maxAmount)) > 0) {
```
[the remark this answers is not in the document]
paulormnas: I talked to Jake and he agreed that a partner would configure only max or min. However, I separated the code to handle each one.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
paulormnas: I splitted the code to handle min and max separated, so it is less confusing now.

core/src/main/java/org/stellar/anchor/util/SepHelper.java
paulormnas: Same here

PR #910, opened 2023-05-31
[ANCHOR-281] Start SEP-24 reference UI as part of dev stack

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Start the SEP-24 reference UI as part of the dev stack.

### Why

We can run the SEP-24 end-to-end with the UI using the IntelliJ run targets.

### Known limitations

N/A


service-runner/src/main/resources/profiles/default/config.env
```diff
@@ -34,5 +34,5 @@ sep12.enabled=true
 sep31.enabled=true
 sep38.enabled=true
 sep24.enabled=true
-sep24.interactive_url.base_url=http://localhost:8081/sep24/interactive
-sep24.more_info_url.base_url=http://localhost:8081/sep24/transaction/more_info
+sep24.interactive_url.base_url=http://localhost:3000/
```
Ifropc, line 37: IIRC default config is used in integration tests too, it will break them.
Considering UI is not needed for tests, can you just create a new profile for that? (Ideally with IDEA .run configuration)

philipliu: The integration tests that depend on SEP24 interactive URLs use the `sep24` profile. I didn't modify those URLs for this reason.

I discussed this with Jamie last week but we decided to just spin up the UI as part of the dev stack. Introducing a separate run configuration would require some changes to the service runner since we need to introduce a way to optionally start the UI in docker compose. I'm in favor of just spinning it up in the dev stack. What do you think?

Ifropc: Good point.
I agree, it can be a part of a dev stack, and I agree that making it optional sounds like a good idea.

.run/Docker - Run Dev Stack - Zookeeper, Kafka, Postgres, SEP24 Reference UI.run.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Docker - Run Dev Stack - Zookeeper, Kafka, Postgres" type="JetRunConfigurationType">
+  <configuration default="false" name="Docker - Run Dev Stack - Zookeeper, Kafka, Postgres, SEP24 Reference UI" type="JetRunConfigurationType">
```
lijamie98, line 2: This will prevent us from starting a separate debugger of the `sep24-reference-ui`. Updating the pages will require updating the docker image. 

This is probably ok because `sep24-reference-ui` is rarely updated.

lijamie98 (approved):
Please feel free to merge when the comments are taken care of.

lijamie98:
Can you also take care of https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/service-runner/src/main/resources/profiles/sep24/config.env

.run/Docker - Run Dev Stack - Zookeeper, Kafka, Postgres, SEP24 Reference UI.run.xml
philipliu: Yes, it's rarely updated, but also there is no dependency on the UI in tests.

philipliu:
@lijamie98 That config is used by SEP-24 integration tests. It needs to point to the reference server. https://github.com/stellar/java-stellar-anchor-sdk/pull/910#discussion_r1212268787

PR #911, opened 2023-05-31
[ANCHOR-308] fix: Use amounts calculated by rate integration in SEP-38

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This fixes a bug in SEP-38 POST quote where the anchor platform recalculates buy and sell amounts even if a value is returned by `/rate` integration.

### Why

As part of the SEP-38 POST quote flow, `/rate` is called to return buy or sell amounts calculated by the business server.  These amounts should not be recalculated by the platform due to differences in decimal precision between the anchor platform and the business server.

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -360,21 +360,42 @@ public class Sep38Service {
 
     // Calculate amounts: sellAmount = buyAmount * totalPrice
     BigDecimal bTotalPrice = decimal(rate.getTotalPrice());
-    BigDecimal bSellAmount, bBuyAmount;
-    if (request.getSellAmount() != null) {
+    BigDecimal bSellAmount = null;
+    BigDecimal bBuyAmount = null;
+
+    if (rate.getSellAmount() != null) {
+      bSellAmount = decimal(rate.getSellAmount());
+    } else if (request.getSellAmount() != null) {
       bSellAmount = decimal(request.getSellAmount());
-      bBuyAmount = bSellAmount.divide(bTotalPrice, buyAsset.getDecimals(), RoundingMode.HALF_UP);
-    } else {
-      bBuyAmount = decimal(request.getBuyAmount());
-      bSellAmount = bBuyAmount.multiply(bTotalPrice);
     }
+
+    if (rate.getBuyAmount() != null) {
+      bBuyAmount = decimal(rate.getBuyAmount());
+    } else if (request.getBuyAmount() != null) {
+      bBuyAmount = decimal(request.getBuyAmount());
+    }
+
+    // TODO: This shouldn't be needed
+    if (bBuyAmount == null && bSellAmount == null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -360,21 +360,39 @@ public Sep38QuoteResponse postQuote(JwtToken token, Sep38PostQuoteRequest reques
...
+      bSellAmount = decimal(rate.getSellAmount());
+    }
+    if (rate.getBuyAmount() != null) {
+      bBuyAmount = decimal(rate.getBuyAmount());
+    }
+
+    // This should not happen because we previously checked at least one was not null
+    if (bBuyAmount == null && bSellAmount == null) {
+      throw new ServerErrorException(
+          String.format(
```
Ifropc, line 378: Can you change the message that they are both null?

philipliu: Done

JakeUrban (approved):
LGTM. Can we port this change to `develop` as well?

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -360,21 +360,36 @@ public Sep38QuoteResponse postQuote(JwtToken token, Sep38PostQuoteRequest reques
...
+
+    // This should not happen because we previously checked at least one was not null
+    if (bBuyAmount == null && bSellAmount == null) {
+      throw new ServerErrorException("Unable to calculate buy and sell amounts, both were null");
+    }
+
+    if (bSellAmount == null) {
       bSellAmount = bBuyAmount.multiply(bTotalPrice);
+    } else if (bBuyAmount == null) {
+      bBuyAmount = bSellAmount.divide(bTotalPrice, buyAsset.getDecimals(), RoundingMode.HALF_UP);
```
JakeUrban, line 383: I see we're using `HALF_UP` here but we're using `FLOOR` when we [calculate `totalPrice` in 2.0](https://github.com/stellar/java-stellar-anchor-sdk/blob/ba1df433f66ae1a6fd58d32c350b125af3346008/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java#L488). We're also using a hard-coded precision.

In a 2.X release we should probably introduce a set of configuration options that allow the anchor platform to calculate and validate amounts instead.

Ifropc: @JakeUrban we have a task for that
https://stellarorg.atlassian.net/browse/ANCHOR-280

Ifropc: We are using FLOOR because one of your partners (currently on v1) uses this mode to calculate the total price, with precision 10.
So, in should not break them, as it will be a default config.
However, new partners may want to have a different setting, so it should be configurable in the asset config

PR #929, opened 2023-06-06
Fix NPE where the country code isn't specified in the sellAsset of Sep38Service

Ifropc:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

fix npe

### Why

fix npe

### Known limitations

n/a

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -89,7 +89,8 @@ public class Sep38Service {
 
     // countryCode
     if (!Objects.toString(countryCode, "").isEmpty()) {
-      if (!sellAsset.getCountryCodes().contains(countryCode)) {
+      if (sellAsset.getCountryCodes() == null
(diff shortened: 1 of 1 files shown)


lijamie98 (approved):
LGTM

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -89,7 +89,8 @@ public GetPricesResponse getPrices(
 
     // countryCode
     if (!Objects.toString(countryCode, "").isEmpty()) {
-      if (!sellAsset.getCountryCodes().contains(countryCode)) {
+      if (sellAsset.getCountryCodes() == null
+          || !sellAsset.getCountryCodes().contains(countryCode)) {
         throw new BadRequestException("Unsupported country code");
```
lijamie98, line 94: The country code isn't specified or is specified but not supported.

JakeUrban (approved):
Is this only relevant to `getPrices()` or are there other places this should be checked?

Ifropc:
All other places check that it's not null

PR #933, opened 2023-06-07
[ANCHOR-310] Add kyc_fields_forwarding configuration and implementation

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

When the `POST /transactions/deposit/interactive` and `POST /transactions/withdraw/interactive` are called, the request may contain KYC fields. If `kyc_fields_forwarding` is enabled, the SEP server will call the `PUT /customer` endpoint of the business server to create or update the KYC fields.

### Why
To prevent leaking PII in the interactive URL


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -350,6 +350,12 @@ sep24:
     account_creation: false
     claimable_balances: false
 
+  # When the `POST /transactions/deposit/interactive` and `POST /transactions/withdraw/interactive` are called,
+  # the request may contain KYC fields. If `kyc_fields_forwarding` is enabled, the SEP server will call the
+  # `PUT /customer` endpoint of the business server to create or update the KYC fields.
+  kyc_fields_forwarding:
+    enabled: true
+
```
JakeUrban, line 358: I think it would be preferable if this config object wasn't necessary. Instead, the platform can check if SEP-12 is enabled when KYC fields are passed by wallets, and if so, it can make the `PUT /customer` request. What do you think?

lijamie98: I think your proposed would be smoother. 

To be cautious, is there a use case when SEP-12 is enabled but SEP-24 does not want to forward KYC fields?

lijamie98: > What if the anchor supports sep-31? How are they going to differentiate between sep-31 and sep-24 customers? By type? And do they need to do so?

I don't think our callback API differentiate them though.

JakeUrban: The Platform would need to use built-in types for SEP-24 requests since the wallet isn't providing it directly. We can use something like `sep24-deposit-customer`.

> To be cautious, is there a use case when SEP-12 is enabled but SEP-24 does not want to forward KYC fields?

I don't think so. I can't envision a case where the business wouldn't want to offer that functionality. If they don't support it but still have SEP-12 enabled that is fine too. They just won't prefill the UI and the user will have to re-enter the KYC.

lijamie98: Hey @JakeUrban , I think it is possible that the anchor does not want to open SEP-12. But they want SEP-24 to receive the KYC via `kycFieldsForwarding`

JakeUrban: Ok, then we can keep this config and SEP-12 doesn't have to be enabled.

api-schema/src/main/java/org/stellar/anchor/api/callback/PutCustomerRequest.java
```diff
@@ -119,4 +123,9 @@ public class PutCustomerRequest {
   String ip_address;
 
   String sex;
+
+  public static PutCustomerRequest from(Sep12PutCustomerRequest request) {
+    Gson gson = new Gson();
+    return new Gson().fromJson(gson.toJson(request), PutCustomerRequest.class);
```
philipliu, line 129: Can we not rely on Gson for this? It's not clear to see how null fields, missing fields are treated without looking how the `Gson` object was initialized. If the default behavior changes between `Gson` versions, it might end up breaking the build.

Same comment for `to`/`from` methods in the other `*Request` classes.

philipliu: Nit: also this mapping seems like an implementation detail of `Sep12Service`, does it make sense to put this method there instead?

api-schema/src/main/java/org/stellar/anchor/api/sep/sep12/PutCustomerVerificationResponse.java
```diff
@@ -1,10 +1,12 @@
...
+
 /**
  * The response to the PUT /customer/verification endpoint of SEP-12.
  *
  * @see <a
  *     href="https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#fields">Refer
  *     to SEP-12</a>
  */
-public class PutCustomerVerificationResponse extends Sep12GetCustomerResponse {}
+public class PutCustomerVerificationResponse extends GetCustomerResponse {}
```
philipliu, line 12: We don't need to refactor this right now, but this is a little strange to read.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/DatabaseMigrationTests.kt
```diff
@@ -5,33 +5,33 @@ import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 
 val PostgresConfig =
-  TestConfig("default").also {
-    it.env["run_all_servers"] = "false"
-    it.env["run_sep_server"] = "true"
-    it.env["data.flyway_enabled"] = "true"
-  }
+    TestConfig("default").also {
```
philipliu, line 8: Did the `ktfmt` config change?

platform/src/main/java/org/stellar/anchor/platform/service/SimpleInteractiveUrlConstructor.java
```diff
@@ -48,6 +57,8 @@ public String construct(Sep24Transaction txn, Map<String, String> request) {
 
   @SneakyThrows
   String constructToken(Sep24Transaction txn, Map<String, String> request) {
+    forwardKycFields(request);
```
philipliu, line 60: It's interesting that constructing a JWT could fail due to a network error. What do you think about moving `forwardKycFields` out of this method and requiring `Sep24Service` to call it explicitly?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/DatabaseMigrationTests.kt
Ifropc: No, everywhere else it's 2 spaces

Ifropc (approved):

api-schema/src/main/java/org/stellar/anchor/api/callback/PutCustomerRequest.java
philipliu: If we're doing this elsewhere, then I won't block the PR on this. But it would be nice to not depend on `Gson` for these types of conversion.

anchor-reference-server/src/main/java/org/stellar/anchor/reference/service/CustomerService.java
```diff
@@ -217,116 +217,116 @@ public Map<String, CustomerField> getSep31ReceiverFields(String type) {
...
 
   public CustomerField createClabeNumberField(String customerType) {
-    CustomerField field = new CustomerField();
-    field.setType("string");
-    field.setDescription("Bank account number for Mexico");
-    field.setOptional(!customerType.equals(Customer.Type.SEP31_RECEIVER.toString()));
-    return field;
+    return CustomerField.builder()
+        .type("string")
+        .description("clabe number of the customer")
```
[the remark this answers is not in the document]
lijamie98: I have changed the description back to `Bank account number for Mexico`.

api-schema/src/main/java/org/stellar/anchor/api/callback/PutCustomerRequest.java
lijamie98: We use GSON for AP. We have a `GsonUtils` class that handles the conversion. 

This PR is focusing on fixing the security leak. I think we can add another refactor ticket tech-debt ticket to clean up the conversion pattern. I would not make this blocking the PR.

api-schema/src/main/java/org/stellar/anchor/api/sep/sep12/PutCustomerVerificationResponse.java
lijamie98: It is a bug to extend from Sep12GetCustomerResponse.

api-schema/src/main/java/org/stellar/anchor/api/sep/sep12/Sep12Status.java
```diff
@@ -27,4 +27,13 @@ public enum Sep12Status {
   public String getName() {
     return this.name;
   }
+
+  public static Sep12Status fromName(String name) {
```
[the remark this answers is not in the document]
lijamie98: I have removed the function.

core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
```diff
@@ -28,22 +33,22 @@ class Sep12ServiceTest {
...
           .build()
     )
   private val mockProvidedFields =
-    mapOf<String, ProvidedField>(
+    mapOf<String, ProvidedCustomerField>(
       "last_name" to
-        ProvidedField.builder()
-          .type(Field.Type.STRING)
+        ProvidedCustomerField.builder()
+          .type("stirng")
```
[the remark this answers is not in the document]
lijamie98: Added assertion to catch the typo.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/DatabaseMigrationTests.kt
lijamie98: It is strange the ktfmt in the Gradle build does not catch it.

platform/src/main/java/org/stellar/anchor/platform/service/SimpleInteractiveUrlConstructor.java
lijamie98: `forwardKycFields` is a anchor platform specific implementation. And this is related to interactive URL construction. That is why it is placed in the constructor. 

It can fail if the configuration is enabled and the `PUT /customer` endpoint is not enabled.

lijamie98: Further thinking about it, `forwardKycFields` can also be moved up to the `construct` method.

core/src/test/kotlin/org/stellar/anchor/test.json
```diff
@@ -0,0 +1,20 @@
+{
```
philipliu, line 1: Looks like this isn't used

philipliu (commented):
Minor comment, otherwise LGTM

philipliu (approved):

core/src/test/kotlin/org/stellar/anchor/test.json
lijamie98: Thank you! Removed.

PR #942, opened 2023-06-13
[ANCHOR-328] Add homeDomain and webAuthDomain length validation 

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add validation that calls `ManageDataOperation.Builder` to validate the `homeDomain` and `webAuthDomain`.

### Why

Cath the errors early.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
@@ -127,7 +127,7 @@ public class PropertySep10Config implements Sep10Config, Validator {
           "homeDomain",
           "sep10-home-domain-invalid",
           format(
-              "The sep10.home_domain (%s) does not have valid format. Please make sure it is a valid domain name. Error=%s",
+              "The sep10.home_domain (%s) is longer than the maximum length (64) of a domain. Error=%s",
               homeDomain, iaex));
     }
 
@@ -138,7 +138,7 @@ public class PropertySep10Config implements Sep10Config, Validator {
           "webAuthDomain",
           "sep10-web-auth-domain-invalid",
           format(
-              "The sep10.web_auth_domain (%s) does not have valid format. Please make sure it is a valid domain name. Error=%s",
+              "The sep10.web_auth_home_domain (%s) is longer than the maximum length (64) of a domain. Error=%s",
               webAuthDomain, iaex));
     }
   }
--- a/platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
@@ -16,7 +16,7 @@ import org.stellar.anchor.config.SecretConfig;
 import org.stellar.anchor.config.Sep10Config;
 import org.stellar.anchor.util.ListHelper;
 import org.stellar.anchor.util.NetUtil;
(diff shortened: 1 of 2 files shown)


platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -117,6 +117,30 @@ void validateConfig(Errors errors) {
           "sep10-jwt-timeout-invalid",
           "The sep10.jwt_timeout must be greater than 0");
     }
+
+    byte[] nonce = new byte[64];
+
+    try {
+      new ManageDataOperation.Builder(String.format("%s %s", homeDomain, "auth"), nonce).build();
+    } catch (IllegalArgumentException iaex) {
+      errors.rejectValue(
```
Ifropc, line 126: Can we have a special error for when domains are too long? Current error message is hard to debug.

Ifropc (changes requested):

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
lijamie98: Hey Gleb, this is still a draft.

lijamie98: I will modify the error messages.

Ifropc: Ah sorry about that, I thought it wasn't a draft. Sounds good, thank you

lijamie98: No problem. 💯

Ifropc (approved):

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -87,6 +87,18 @@ void validateConfig(Errors errors) {
       errors.rejectValue(
           "homeDomain", "home-domain-empty", "The sep10.home_domain is not defined.");
     } else {
+      try {
+        new ManageDataOperation.Builder(String.format("%s %s", homeDomain, "auth"), new byte[64])
```
philipliu, line 91: Can we make the length check more explicit? Something like:
```java
byte[] bytes = homeDomain.getBytes(StandardCharsets.US_ASCII);
if (bytes.length > 64) {
     errors.reject( ... )
}
```

lijamie98: That's a good alternative. However, the size check (<=64) happens in the `ManageDataOperation` constructor. It would be more robust if we call the `ManageDataOperation`.

philipliu: I'm not sure if I'm convinced that is more robust. For example, if there are new validations added to the `ManageDataOperation` constructor, we will fail with a message complaining about the size. Or if the allowed since increases from 64, we will need to update the message here.

lijamie98: This is actually intended. If there is a new validation added to `ManageDataOperation`, the validation should fail because AP uses `ManageDataOperation` to create SEP-10 challenge.

PR #946, opened 2023-06-15
[ANCHOR-342] Add `clients` to the configuration

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add client_attribution to the configuration.

### Why

This is to support the client status callback feature.

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -282,12 +282,15 @@ sep10:
   #
   client_attribution_required: false
   ## @param: client_attribution_allow_list
-  ## Set the white list of the client domain. The domains are comma-separated and must be defined in the
-  #  `client_attribution` section of the configuration.
+  ## Set the white list of the client domain. The domains are comma-separated. If a domain is defined in the
+  #  `client_attribution` section of the configuration, the public_key defined in the client_attribution configuration
+  #  mush match the `SIGNING_KEY` of the client domain's TOML file.
   client_attribution_allow_list:
   ## @param: client_attribution_deny_list
   ## Set the black list of the client domain. The domains are comma-separated and must be defined in the
-  #  `client_attribution` section of the configuration.
+  #  `client_attribution` section of the configuration. The domains are comma-separated. If a domain is defined in the
+  #  #  `client_attribution` section of the configuration, the public_key defined in the client_attribution configuration
+  #  #  mush match the `SIGNING_KEY` of the client domain's TOML file.
   client_attribution_deny_list:
   ## @param: auth_timeout
   ## @type: integer
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -133,21 +133,17 @@ payment_observer:
     # The initial backoff (cool-down) time (in seconds) before reconnecting to the event publisher 
     max_event_backoff_time: 300
 
-
 event_processor:
   # The context path of the event processing server
   context_path: /
   # The listening port of the event processing server
   port: 8084
-  # The flag to enable the status callback to the client domain
-  client_domain_status_callback_enabled: true
-  # The flag to enable the event delivery to the anchor business server
-  callback_api_event_delivery_enabled: true
-  # The queue where the event processor will
-  queue:
-    type: kafka
-    kafka:
-      bootstrap_server: localhost:29092
+  # The configuration of the status callback to the client domain
+  client_domain_status_callback:
+    enabled: true
+  # The configuration of the event delivery to the anchor business server
+  callback_api_event_delivery:
+    enabled: true
 
 # The list of clients for the Anchor server to safely communicate with the outside client domains.
 # Each item in the list may contain the following fields:
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -282,13 +282,12 @@ sep10:
   #
   client_attribution_required: false
   ## @param: client_attribution_allow_list
-  ## Set the white list of the client domain. The domains are comma-separate and must be defined in the
+  ## Set the white list of the client domain. The domains are comma-separated and must be defined in the
   #  `client_attribution` section of the configuration.
   client_attribution_allow_list:
   ## @param: client_attribution_deny_list
-  ## Set the black list of the client domain. The domains are comma-separated. The domains are comma-separate and must be defined in the
+  ## Set the black list of the client domain. The domains are comma-separated and must be defined in the
   #  `client_attribution` section of the configuration.
-  #
   client_attribution_deny_list:
   ## @param: auth_timeout
   ## @type: integer
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -139,10 +139,10 @@ event_processor:
   # The listening port of the event processing server
   port: 8084
   # The configuration of the status callback to the client domain
-  client_domain_status_callback:
+  client_status_callback:
     enabled: true
   # The configuration of the event delivery to the anchor business server
-  callback_api_event_delivery:
+  callback_api_request:
     enabled: true
 
 # The list of clients for the Anchor server to safely communicate with the outside client domains.
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -158,13 +158,11 @@ event_processor:
 #
 # Examples:
 #     - domain: lobstr.com
-#       type: non-custodial
 #       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
-#       url: https://callback.lobstr.com/api/v1/anchor/callback
+#       callback_url: https://callback.lobstr.com/api/v1/anchor/callback
 #     - domain: mgi.com
-#       type: custodial
 #       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
-#       url: https://callback.mgi.com/api/v2/anchor/callback
+#       callback_url: https://callback.mgi.com/api/v2/anchor/callback
 client_attribution:
 
 ## @param: languages
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -159,7 +159,13 @@ event_processor:
 #     - domain: mgi.com
 #       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
 #       callback_url: https://callback.mgi.com/api/v2/anchor/callback
-client_attribution:
+clients:
+#     - domain: domain1.com
+#       public_key:
+#       callback_url:
+#     - domain: domain2.com
+#       public_key:
+#       callback_url:
 
 ## @param: languages
 ## @supported_values: en
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -139,17 +139,33 @@ event_processor:
   context_path: /
   # The listening port of the event processing server
   port: 8084
+  # The flag to enable the status callback to the client domain
+  client_domain_status_callback_enabled: true
+  # The flag to enable the event delivery to the anchor business server
+  callback_api_event_delivery_enabled: true
   # The queue where the event processor will
   queue:
     type: kafka
     kafka:
       bootstrap_server: localhost:29092
-      event_type_to_topic:
-        all: dev_ap_event_single_queue
-        quote_created: ap_event_quote_created
-        transaction_created: ap_event_transaction_created
-        transaction_status_updated: ap_event_transaction_status_changed
-        transaction_error: ap_event_transaction_error
+
+# The list of clients for the Anchor server to safely communicate with the outside client domains.
+# Each item in the list may contain the following fields:
+#   - domain: the domain of the client
+#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
+#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
+#   - url: the URL of the client's callback API endpoint
+#
+# Examples:
+#     - domain: lobstr.com
+#       type: non-custodial
+#       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
+#       url: https://callback.lobstr.com/api/v1/anchor/callback
+#     - domain: mgi.com
+#       type: custodial
+#       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
+#       url: https://callback.mgi.com/api/v2/anchor/callback
+client_attribution:
 
 ## @param: languages
 ## @supported_values: en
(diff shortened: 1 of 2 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -266,12 +282,12 @@ sep10:
   #
   client_attribution_required: false
   ## @param: client_attribution_allow_list
-  ## Set the white list of the client domain. The domains are comma-separated.
-  #
+  ## Set the white list of the client domain. The domains are comma-separated and must be defined in the
```
Ifropc, line 285: >and must be defined in the `client_attribution` section of the configuration.

This is not backward compatible...

lijamie98: Good catch. Updated.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -266,12 +282,15 @@ sep10:
   #
   client_attribution_required: false
   ## @param: client_attribution_allow_list
-  ## Set the white list of the client domain. The domains are comma-separated.
-  #
+  ## Set the white list of the client domain. The domains are comma-separated. If a domain is defined in the
+  #  `client_attribution` section of the configuration, the public_key defined in the client_attribution configuration
```
Ifropc, line 286: I think we should use domain instead of public key, as it's more user friendly

lijamie98: Can you help to make the suggestion to elaborate?

Ifropc:
Can we deprecate
```
  ## @param: omnibus_account_list
  ## A comma-separated list of omnibus accounts (comma separated).
  ## The SEP-10 omnibus account is described at:
  ##     https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#memos
  #
  omnibus_account_list:
  ## @param: require_known_omnibus_account
  ## @type: bool
  ## Whether to require authenticating clients to be in the omnibusAccountList
  #
  require_known_omnibus_account: false
```
Because with new schema we can always just use `client_attribution`

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
...
+# Each item in the list may contain the following fields:
+#   - domain: the domain of the client
+#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
+#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
+#   - url: the URL of the client's callback API endpoint
+#
+# Examples:
+#     - domain: lobstr.com
+#       type: non-custodial
+#       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
```
Ifropc, line 162: Does non-custodial wallet even have a public key? I think it should only be required for custodial wallets

platform/src/main/resources/config/anchor-config-default-values.yaml
Ifropc: Sorry please check this comment instead https://github.com/stellar/java-stellar-anchor-sdk/pull/946/files/8be50e02f4c33999fb895b84251f888a043bb32f#r1231540626

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Doesn't lobstr.com server has a `SIGNING_KEY` in the TOML file?

cc: @JakeUrban

lijamie98:
> Can we deprecate
> 
> ```
>   ## @param: omnibus_account_list
>   ## A comma-separated list of omnibus accounts (comma separated).
>   ## The SEP-10 omnibus account is described at:
>   ##     https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#memos
>   #
>   omnibus_account_list:
>   ## @param: require_known_omnibus_account
>   ## @type: bool
>   ## Whether to require authenticating clients to be in the omnibusAccountList
>   #
>   require_known_omnibus_account: false
> ```
> 
> Because with new schema we can always just use `client_attribution`

Is there a use case where a wallet uses `omnibus_account` without `client_attribution`?

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: Noncustodial wallets do have a public key, published on their stellar.toml file. This is nice because the non-custodial wallet could rotate their signing key and the anchor's authentication process would still work because it should fetch the `SIGNING_KEY` on each authentication request.

For this same reason, I don't think we should require each wallet entry to have a public key attribute. Only custodial wallets will need the `public_key` attribute.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
   context_path: /
   # The listening port of the event processing server
   port: 8084
+  # The flag to enable the status callback to the client domain
+  client_domain_status_callback_enabled: true
+  # The flag to enable the event delivery to the anchor business server
+  callback_api_event_delivery_enabled: true
```
JakeUrban, line 145: I don't love the naming here. I think something generalized would be better:

```
  events:
    transaction_status_updated:
      wallet_callbacks_enabled: true
      callback_api_requests_enabled: true
```

This way, we can add more event types and more handlers of those events using a predefined structure.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
...
-        quote_created: ap_event_quote_created
-        transaction_created: ap_event_transaction_created
-        transaction_status_updated: ap_event_transaction_status_changed
-        transaction_error: ap_event_transaction_error
+
+# The list of clients for the Anchor server to safely communicate with the outside client domains.
+# Each item in the list may contain the following fields:
+#   - domain: the domain of the client
+#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
+#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
```
JakeUrban, line 156: I suggest naming this something more specific in case we want to add public keys for sending & receiving payments. We should also design the structure to support multiple keys just in case. For example:

```
wallets:
  - domain: whitebit.com
    auth_accounts:
      - GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
    source_accounts:
      - GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
    destination_accounts:
      - GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
```

I don't think we should add source / destination accounts right now, but I think its good to prepare for them if requested by a partner.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
...
+#
+# Examples:
+#     - domain: lobstr.com
+#       type: non-custodial
+#       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
+#       url: https://callback.lobstr.com/api/v1/anchor/callback
+#     - domain: mgi.com
+#       type: custodial
+#       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
+#       url: https://callback.mgi.com/api/v2/anchor/callback
```
JakeUrban, line 167: We should make the name more specific.

```
wallets:
  - domain: beansapp.com
    callback_url: https://beansapp.com/callbacks
```

JakeUrban (commented):
Made some comments / suggestions. Regarding backwards compatibility, I think we should deprecate the allowlist & denylist attributes but keep the `client_attribution_required` flag. Then we'll validate that everything matches on startup if the `wallets` object is present.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
...
+# The list of clients for the Anchor server to safely communicate with the outside client domains.
+# Each item in the list may contain the following fields:
+#   - domain: the domain of the client
+#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
+#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
+#   - url: the URL of the client's callback API endpoint
+#
+# Examples:
+#     - domain: lobstr.com
+#       type: non-custodial
```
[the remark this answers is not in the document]
lijamie98: +1

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: I think we can add `type` when there is an immediate use case. In that case, the `public_key` will stay.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: I avoided `wallets` because it actually represents `clients` and this section is to identify the client attribution.

lijamie98: The purpose of this section is to identify the client (wallet). The detail (`source_accounts`, `destination_accounts`, etc) of the wallet should be obtained from the TOML file from `https://domain/.well-known/stellar.toml` to avoid multiple sources of truth and avoid unnecessary inconsistency.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -139,17 +139,33 @@ event_processor:
...
+#       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
+#       url: https://callback.lobstr.com/api/v1/anchor/callback
+#     - domain: mgi.com
+#       type: custodial
+#       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
+#       url: https://callback.mgi.com/api/v2/anchor/callback
+client_attribution:
```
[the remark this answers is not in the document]
lijamie98: Can the attribution represents some entities other than a `wallet` such as a sending anchor? 

If we have a non-wallet client, the naming will be off.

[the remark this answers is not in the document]
JakeUrban: Good point. `clients:`?

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: The challenge there is that custodial wallets may not have toml files.

Ifropc: We can have accounts etc to be optional, but I agree in general toml file is preferred.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: There is a `event` section that defines the existing `EventService`. This is specific to `event-processor`. 

I would probably swap the structure like. Please refer to the updated commit.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: What's the purpose of 
`source_accounts: - GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU`
and `destination_accounts: - GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU`

platform/src/main/resources/config/anchor-config-default-values.yaml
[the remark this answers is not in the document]
lijamie98: This is for client attribution purposes. `clients` feels too generic and can be confusing. I don't have better naming though. Can we have another one?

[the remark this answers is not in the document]
JakeUrban: I think I disagree, I think including `_attribution` or something that implies that the list of clients is used for a specific purpose might get us into trouble when we find another use case for the list. I'd rather use `clients` for the list and then reference the clients by domain elsewhere in the configuration if necessary for attribution or another use case. wdyt @Ifropc ?

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: They don't have a functional purpose today, but MoneyGram whitelists custodial wallets source & destination accounts today, so I could see them requesting this.

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: I see, I don't mind using the `event-processor` top-level object.

What do you think about using event type as a sub-object though? That way the business could subscribe to a subset of events, instead of having just a single `true`/`false` attribute for all event types.

For example,

```
event_processor:
  event_type:
    transaction_status_updated:
      client_callback_enabled: true
      callback_api_requests_enabled: true
    transaction_created:
      client_callback_enabled: false
      callback_api_requests_enabled: true
```

JakeUrban: I don't feel strongly about this. If we want to use a single attribute for all requests that it ok, we'll just have less flexibility and need to make assumptions about what anchors and clients will want.

platform/src/main/resources/config/anchor-config-default-values.yaml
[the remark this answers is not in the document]
Ifropc: I'd prefer it to be `wallets`, but as Jamie said it could be anchors too. So IMO `clients` is better than attribution, not sure if attribution is a good word.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Adding configuration won't break the backward compatibility. IMO, we may skip these two fields until requested.

platform/src/main/resources/config/anchor-config-default-values.yaml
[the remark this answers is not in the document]
lijamie98: Since I don't have better wording either, let's use `clients:`.

lijamie98:
> 

I am not sure if we may deprecate the `omnibus_account_list`. 
- `omnibus_account_list` seems to be inconsistent with `client`.
- I think it is possible `omnibus_account_list` can be different from the `wallet_address`s defined in the `clients` section.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Not all callbacks are interested in all event types. How about we do this:
```
event_processor:
  # The context path of the event processing server
  context_path: /
  # The listening port of the event processing server
  port: 8084
  # The configuration of the status callback to the client domain
  client_domain_status_callback:
    enabled: true
  # The configuration of the event delivery to the anchor business server
  callback_api_event_delivery:
    enabled: true
```
This will leave more room for customization and we don't have to define the behavior of the events unless it is necessary to customize.

JakeUrban: Hey Jamie, can we use the following names, `client_status_callback` and `callback_api_requests`?

JakeUrban:
> I am not sure if we may deprecate the omnibus_account_list.
> omnibus_account_list seems to be inconsistent with client.
> I think it is possible omnibus_account_list can be different from the wallet_addresss defined in the clients section.

Can you provide an example of how it is inconsistent? I think the omnibus account list is directly associated with the `auth_accounts` of a given client.

Ifropc (approved):

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: Works for me!

JakeUrban (approved):

PR #958, opened 2023-06-21
[ANCHOR-257] Validate that each asset is unique on startup

JiahuiWho:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Ticket: https://stellarorg.atlassian.net/browse/ANCHOR-257

### Why

Validated PropertyAssetConfig on start to avoid duplicate ticker being added.

### Known limitations

N/A
See tests for test plan


platform/src/main/java/org/stellar/anchor/platform/config/PropertyAssetsConfig.java
```diff
@@ -66,6 +71,18 @@ public void validate(@NotNull Object target, @NotNull Errors errors) {
               String.format("assets.type:%s is not supported.", this.getType()));
           break;
       }
+
+      assert assetService != null;
```
lijamie98, line 75: Can we do `errors.reject()` instead of stopping the JVM?

platform/src/test/resources/test_assets_duplicate_asset.json
```diff
@@ -0,0 +1,168 @@
+{
+  "assets": [
```
lijamie98, line 2: To simplify and focus the test, can we use minimal JSON file. 

For example, remove the `deposit`, `withdraw`, `send`, `sep31`, `sep38` fields?

platform/src/test/resources/test_assets_duplicate_asset.yaml
```diff
@@ -0,0 +1,121 @@
+assets:
```
lijamie98, line 1: Same as above.

platform/src/main/java/org/stellar/anchor/platform/config/PropertyAssetsConfig.java
JiahuiWho: This line is to get away with the NullPointerException warning from the following line `assetService.listAllAssets()`.
assetService here will never be null actually, the prior switch case have captured all the null case.

Let me know if this makes sense to you :) Or we can remove this line

philipliu:
Alternatively, we could enforce the uniqueness constraint in the initialization of `DefaultAssertService`. The underlying data structure is a `List` of assets which can be swapped to a `Map`. I'm thinking this is something we would want for performance reasons anyway for Anchors supporting a large number of assets.

JiahuiWho:
Yeah I also prefer to move uniqueness constrain to `DefaultAssetService ` initialization phase. No assetService should be returned if config is invalid.
Regarding to swap `List` to `Map`, let's do it in separate PR due to it and its underlying object being referenced in multiple places. @lijamie98 what do you think ?

philipliu (approved):
LGTM

PR #967, opened 2023-06-28
[ANCHOR-342] Add `type` to clients configuration

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Add `type` field to the client configuration. The required fields will change according to the `type` assigned.

### Why

To clearly distinguish the `custodial` and `noncustodial` wallet clients.

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -145,27 +145,42 @@ event_processor:
   callback_api_request:
     enabled: true
 
-# The list of clients for the Anchor server to safely communicate with the outside client domains.
+# The list of clients for the Anchor server to safely communicate with the outside wallet servers or clients.
 # Each item in the list may contain the following fields:
-#   - domain: the domain of the client
-#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
-#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
-#   - url: the URL of the client's callback API endpoint
+#   - type: `custodial` or `noncustodial`
+#
+#   If the type is `custodial`,
+#   - account: (required) the custodial account of the client. If the `account` of the transaction is the same as this
+#              account, the callback will be activated.
+#   - url: (required) the URL of the client's callback API endpoint
+#
+#   If the type is `noncustodial`,
+#   - domain: (required) the domain of the client. If the `client_domain` field of the transaction is the same as this
+#             domain, the callback will be activated.
+#   - url: (required) the URL of the client's callback API endpoint
+#   - public_key: (optional) the public key of the client. If the public key is specified, the client domain's TOML
+#                 file will be fetched and the `SIGNING_KEY` of the TOML file will be compared with the public key.
+#                 If they don't match, a validation error will be thrown.
 #
 # Examples:
-#     - domain: lobstr.com
-#       public_key: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
-#       callback_url: https://callback.lobstr.com/api/v1/anchor/callback
-#     - domain: mgi.com
+#     - type: custodial
+#       account: GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2
+#       url: https://callback.circle.com/api/v1/anchor/callback
+#     - type: noncustodial
+#       domain: lobstr.co
+#       callback_url: https://callback.lobstr.co/api/v2/anchor/callback
+#     - type: noncustodial
+#       domain: vibrant.co
+#       callback_url: https://callback.vibrant.com/api/v2/anchor/callback
 #       public_key: GA22WORKYRXB6AW7XR5GIOAOQUY4KKCENEAI34FN3KJNWHKDZTZSVLTU
-#       callback_url: https://callback.mgi.com/api/v2/anchor/callback
 clients:
-#     - domain: domain1.com
+#     - type: custodial
+#       account:
+#       url:
+#     - type: noncustodial
+#       domain:
+#       url:
 #       public_key:
-#       callback_url:
-#     - domain: domain2.com
-#       public_key:
-#       callback_url:
 
 ## @param: languages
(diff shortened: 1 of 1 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -145,27 +145,42 @@ event_processor:
...
-#   - domain: the domain of the client
-#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
-#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
-#   - url: the URL of the client's callback API endpoint
+#   - type: `custodial` or `noncustodial`
+#
+#   If the type is `custodial`,
+#   - account: (required) the custodial account of the client. If the `account` of the transaction is the same as this
+#              account, the callback will be activated.
+#   - url: (required) the URL of the client's callback API endpoint
```
JakeUrban, line 155: Lets update the comments here to use the `callback_url` attribute name.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -145,27 +145,42 @@ event_processor:
...
 # Each item in the list may contain the following fields:
-#   - domain: the domain of the client
-#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
-#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
-#   - url: the URL of the client's callback API endpoint
+#   - type: `custodial` or `noncustodial`
+#
+#   If the type is `custodial`,
+#   - account: (required) the custodial account of the client. If the `account` of the transaction is the same as this
+#              account, the callback will be activated.
```
JakeUrban, line 154: Can we clarify this definition to be "the public key used to authenticate the client". Custodial wallets may use different accounts to send & receive payments.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Thanks for catching this. Updated.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Updated.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -145,27 +145,43 @@ event_processor:
...
+#   - account: (required) the custodial account of the client. If the `account` of the transaction is the same as this
+#              account, the callback will be activated. The `account` field of the transaction is the account that is
+#              used for the SEP-10 authentication.
+#   - callback_url: (required) the URL of the client's callback API endpoint
+#
+#   If the type is `noncustodial`,
+#   - domain: (required) the domain of the client. If the `client_domain` field of the transaction is the same as this
+#             domain, the callback will be activated.
+#   - callback_url: (required) the URL of the client's callback API endpoint
+#   - public_key: (optional) the public key of the client. If the public key is specified, the client domain's TOML
```
Ifropc, line 162: Is there a reason we need a public key in config, if we can always get it from TOML?

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -145,27 +145,43 @@ event_processor:
...
+# The list of clients for the Anchor server to safely communicate with the outside wallet servers or clients.
 # Each item in the list may contain the following fields:
-#   - domain: the domain of the client
-#   - type: the type of the client. The supported values are `custodial` and `non-custodial`
-#   - public_key: the public key of the client. This must match the public key of the TOML file of the client domain
-#   - url: the URL of the client's callback API endpoint
+#   - type: `custodial` or `noncustodial`
+#
+#   If the type is `custodial`,
+#   - account: (required) the custodial account of the client. If the `account` of the transaction is the same as this
```
Ifropc, line 153: How is this field going to be used? Is it for whitelisting SEP-10/SEP-24 calls?
If yes, I think we need to have a list of accounts, as wallets may use different account for withdrawal/deposit/authentication

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: It is for verification process. If not specified, it won't be verified.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: @Ifropc @JakeUrban Can a custodial wallet have multiple accounts?

JakeUrban: They will likely only use one account to authenticate, but also likely will have many accounts for sending & receiving payments.

lijamie98: In this case, should we change `account` to `accounts` to allow multiple accounts?

JakeUrban: I would use this scheme:

```
- type: noncustodial
  domain: lobstr.co
  callback_url: https://lobstr.co/callback
  signing_key: G...  # optional
- type: custodial
  public_key: G...
  callback_url: https://whitebit.com/callback
```

In short:
- use signing_key instead of public_key for noncustodial wallets to avoid name overloading
- using public_key instead of account for custodial wallets because the stellar account doesn't need to exist for SEP-10 auth
- eventually, we may want to add `destination_accounts` and `source_accounts` lists to the custodial client config, but we don't need to do that until requested

Ifropc: +1 to Jake

JakeUrban (approved):

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: We will do ``destination_accounts` and `source_accounts` when they are requested.

JiahuiWho (approved):

platform/src/main/resources/config/anchor-config-default-values.yaml
Ifropc: Yeah I think we should separate auth_key from other types of keys

PR #969, opened 2023-07-03
SDFSOF-63 - Added implementation of endpoint for handling RPC calls with corresponding request validation and basic error handling

MazurakIhor:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Implementation of endpoint for handling RPC calls with request validation and error handling

### Why

This endpoint will handle RPC calls that should replace PATCH /transactions REST endpoint

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/exception/RpcException.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/exception/RpcException.java
@@ -1,0 +1,22 @@
+package org.stellar.anchor.api.exception;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class RpcException extends AnchorException {
+  private final int errorCode;
+  private Object additionalData;
+
+  public RpcException(int errorCode, String message) {
+    super(message);
+    this.errorCode = errorCode;
+  }
+
+  public RpcException(int errorCode, String message, Object additionalData) {
+    super(message);
+    this.errorCode = errorCode;
+    this.additionalData = additionalData;
+  }
+}
--- a/api-schema/src/main/java/org/stellar/anchor/api/exception/RpcException.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/exception/RpcException.java
@@ -2,19 +2,20 @@ package org.stellar.anchor.api.exception;
 
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import org.stellar.anchor.api.rpc.RpcErrorCode;
 
 @Getter
 @EqualsAndHashCode(callSuper = false)
 public class RpcException extends AnchorException {
-  private final int errorCode;
+  private final RpcErrorCode errorCode;
   private Object additionalData;
 
-  public RpcException(int errorCode, String message) {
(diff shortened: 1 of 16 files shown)


MazurakIhor:
softserve_develop branch is 17 commits behind develop but changes are small and easy to merge.
I am going to merge latest develop into softserve_develop branch after this PR is merged, just to not mix my changes with changes from develop branch. But if you think it's better to merge latest develop right now - NP, I will merge it.

JakeUrban:
Either approach is fine by me, whatever you think is best.

api-schema/src/main/java/org/stellar/anchor/api/rpc/RpcRequest.java
```diff
@@ -0,0 +1,13 @@
+package org.stellar.anchor.api.rpc;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class RpcRequest {
+  private Object id;
```
Ifropc, line 9: Pretty sure it's just String

> id
    An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [[1]](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain fractional parts [[2]](https://www.jsonrpc.org/specification#id2)

(from specs)

api-schema/src/main/java/org/stellar/anchor/api/rpc/RpcRequest.java
```diff
@@ -0,0 +1,13 @@
...
+  private String jsonrpc;
+  private String method;
+  private Object params;
```
Ifropc, line 12: I'm guessing it's not possible to make it generic, on what step this going to be converted to an actual object?

Ifropc (approved):

api-schema/src/main/java/org/stellar/anchor/api/exception/RpcException.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class RpcException extends AnchorException {
+  private final int errorCode;
+  private Object additionalData;
+
+  public RpcException(int errorCode, String message) {
```
Ifropc, line 12: Should it has RpcErrorCode as parameter instead of int?

MazurakIhor: Changed `int errorCode` to `RpcErrorCode errorCode`

api-schema/src/main/java/org/stellar/anchor/api/rpc/RpcRequest.java
MazurakIhor: Request may contain `"id": "1"` or `"id": 1`.
String will handle both cases and convert to String with value "1"
Response will contain String as well, even if request contained Number. But it's shoun't be a problem to convert it back to Number.

Ifropc: Oh yeah you are right gson treats it differently. Let's keep it like this then

PR #970, opened 2023-07-03
[Anchor-338] Implement static status callback configuration

JiahuiWho:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Create client status call back configuration, validate both `custodial` and `non-custodial` type

### Why

### Known limitations

Note that for non-custodial client with signing_key provided, we need to check it against TOML file. 

The function `fetchSigningKey` is a copy of `getClientAccountId`(which returns signing_key read from toml file) in `Sep10Service`,  I would love to extract this function to a helper class for reuse.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
+++ b/core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
@@ -1,0 +1,46 @@
+package org.stellar.anchor.sep10;
+
+import static org.stellar.anchor.util.Log.debugF;
+import static org.stellar.anchor.util.Log.infoF;
+
+import java.io.IOException;
+import org.stellar.anchor.api.exception.SepException;
+import org.stellar.anchor.util.Sep1Helper;
+import org.stellar.sdk.FormatException;
+import org.stellar.sdk.KeyPair;
+
+public class Sep10Helper {
+
+  /**
+   * Fetch SIGNING_KEY from clint_domain by reading the stellar.toml content.
+   *
+   * @param clientDomain The client's domain. E.g. lobstr.co.
+   * @return The SIGNING_KEY presented in client's TOML file.
+   * @throws SepException if SIGNING_KEY not present or error happens
+   */
+  public static String fetchSigningKeyFromClientDomain(String clientDomain) throws SepException {
+    String clientSigningKey = "";
+    String url = "https://" + clientDomain + "/.well-known/stellar.toml";
+    try {
+      debugF("Fetching {}", url);
+      Sep1Helper.TomlContent toml = Sep1Helper.readToml(url);
+      clientSigningKey = toml.getString("SIGNING_KEY");
+      if (clientSigningKey == null) {
+        infoF("SIGNING_KEY not present in 'client_domain' TOML.");
+        throw new SepException("SIGNING_KEY not present in 'client_domain' TOML");
+      }
+
+      // client key validation
+      debugF("Validating client_domain signing key: {}", clientSigningKey);
+      KeyPair.fromAccountId(clientSigningKey);
+      return clientSigningKey;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.stellar.anchor.util.Log.infoF;
+
+import java.io.IOException;
+import org.stellar.anchor.api.exception.SepException;
+import org.stellar.anchor.util.Sep1Helper;
+import org.stellar.sdk.FormatException;
+import org.stellar.sdk.KeyPair;
+
+public class Sep10Helper {
+  public static String getClientAccountId(String clientDomain) throws SepException {
```
lijamie98, line 13: This function is complex. Let's start adding comments.

lijamie98: How about rename to `fetchSigningKey()`?

platform/src/main/java/org/stellar/anchor/platform/config/ClientConfig.java
```diff
@@ -0,0 +1,87 @@
...
+import lombok.Data;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.validation.Errors;
+import org.springframework.validation.ValidationUtils;
+import org.springframework.validation.Validator;
+import org.stellar.anchor.api.exception.SepException;
+import org.stellar.anchor.sep10.Sep10Helper;
+
+@Data
+public class ClientConfig implements Validator {
```
lijamie98, line 15: 💡 Rename to `ClientsConfig` to be consistent with `clients:`

lijamie98 (changes requested):

core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
JiahuiWho: Yeah this function is moved from Sep10Service for reuse, yeah I agree `fetchSigningKey` would be a more proper name

core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.stellar.anchor.util.Log.infoF;
+
+import java.io.IOException;
+import org.stellar.anchor.api.exception.SepException;
+import org.stellar.anchor.util.Sep1Helper;
+import org.stellar.sdk.FormatException;
+import org.stellar.sdk.KeyPair;
+
+public class Sep10Helper {
+  public static String fetchSigningKey(String clientDomain) throws SepException {
```
lijamie98, line 13: This function is complex. Let's start adding comments.

platform/src/main/java/org/stellar/anchor/platform/config/ClientsConfig.java
```diff
@@ -0,0 +1,85 @@
...
+        "signingKey",
+        "empty-client-signing-key",
+        "The client.signingKey cannot be empty and must be defined");
+
+    ValidationUtils.rejectIfEmptyOrWhitespace(
+        errors,
+        "callbackUrl",
+        "empty-client-callback-url",
+        "The client.callbackUrl cannot be empty and must be defined");
+  }
```
JakeUrban, line 53: Callback URLs should be optional. Some wallets may opt to poll instead, or may use polling first before implementing callbacks at a later stage. Same goes for noncustodial wallets.

lijamie98: In that case, please also update the comments of the config YAML file.

lijamie98 (approved):
LGTM

PR #973, opened 2023-07-05
Fix: Increase strictness of SEP-24 info test

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Update the test so that expected info response includes all assets and fields specified in `assets.yaml`.

### Why

This test would have quietly failed if `native` asset was not included in the response for some reason.

### Known limitations

N/A

### DIFF
--- a/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24Tests.kt
+++ b/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24Tests.kt
@@ -6,6 +6,7 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.assertThrows
 import org.skyscreamer.jsonassert.JSONAssert
 import org.skyscreamer.jsonassert.JSONCompareMode.LENIENT
+import org.skyscreamer.jsonassert.JSONCompareMode.NON_EXTENSIBLE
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24Tests.kt
```diff
@@ -327,21 +328,35 @@ private const val expectedSep24Info =
         "enabled": true
       },
       "USD": {
-        "enabled": true
+        "enabled": true,
```
lijamie98, line 331: We are not implementing the `min_amount` and `max_amount` in AP and we actually decided to move the `fee` model to the SEP-38 endpoint. We should not add these fields back.

philipliu: Not sure I understand, we use these fields for request validation. https://github.com/stellar/java-stellar-anchor-sdk/blob/a126540141af5e32de03ca00f5ef70a831b26988/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java#L124-L131

lijamie98: These are legacy. We should remove these fields from the code.

PR #975, opened 2023-07-06
[ANCHOR-365]

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This implements the POJOs for the SEP-6 related configuration.

### Why

We are implementing SEP-6 in the coming weeks.

### Known limitations

N/A

### DIFF
--- a/core/src/test/resources/test_assets.json
+++ b/core/src/test/resources/test_assets.json
@@ -22,6 +22,53 @@
         "min_amount": 1,
         "max_amount": 1000000
       },
+      "sep6": {
+        "deposit": {
+          "exchange_enabled": true,
+          "fields": {
+            "email_address": {
+              "description": "email address for transaction status updates",
+              "optional": true
+            }
+          }
+        },
+        "withdraw": {
+          "exchange_enabled": true,
+          "types": {
+            "bank_account": {
+              "fields": {
+                "dest": {
+                  "description": "your bank account number"
+                },
+                "dest_extra": {
+                  "description": "your routing number"
+                },
+                "bank_branch": {
+                  "description": "address of your bank branch"
+                },
+                "phone_number": {
+                  "description": "your phone number in case there's an issue"
+                },
+                "country_code": {
+                  "description": "ISO-3166-1 alpha-3 code of the user's current address",
+                  "choices": [
+                    "USA",
+                    "PRI"
+                  ]
+                }
+              }
+            },
+            "cash": {
+              "fields": {
+                "dest": {
+                  "description": "your email address. Your cashout PIN will be sent here.",
+                  "optional": true
+                }
+              }
+            }
+          }
+        }
+      },
       "sep31": {
         "quotes_supported": true,
         "quotes_required": true,
@@ -74,6 +121,7 @@
           "iso4217:USD"
         ]
       },
+      "sep6_enabled": true,
       "sep24_enabled": true,
       "sep31_enabled": true,
       "sep38_enabled": true
@@ -143,6 +191,7 @@
           "iso4217:USD"
         ]
       },
+      "sep6_enabled": false,
       "sep24_enabled": true,
       "sep31_enabled": true,
       "sep38_enabled": true
@@ -189,6 +238,8 @@
           }
         ]
       },
+      "sep6_enabled": false,
+      "sep24_enabled": true,
       "sep31_enabled": false,
       "sep38_enabled": true
     },
@@ -262,8 +313,10 @@
         "exchangeable_assets": [
           "stellar:USDC:GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP"
         ],
-       "decimals": 7
+        "decimals": 7
       },
+      "sep6_enabled": false,
+      "sep24_enabled": true,
       "sep31_enabled": false,
       "sep38_enabled": false
     }
--- a/core/src/test/resources/test_assets.yaml
+++ b/core/src/test/resources/test_assets.yaml
@@ -17,6 +17,36 @@ assets:
       fee_percent: 0
       min_amount: 1
       max_amount: 1000000
+    sep6:
+      deposit:
+        exchange_enabled: true
+        fields:
+          email_address:
+            description: email address for transaction status updates
+            optional: true
+      withdraw:
+        exchange_enabled: true
+        types:
+          bank_account:
+            fields:
+              dest:
+                description: your bank account number
+              dest_extra:
+                description: your routing number
+              bank_branch:
+                description: address of your bank branch
+              phone_number:
+                description: your phone number in case there's an issue
+              country_code:
+                description: ISO-3166-1 alpha-3 code of the user's current address
+                choices:
+                  - USA
+                  - PRI
+          cash:
+            fields:
+              dest:
+                description: your email address. Your cashout PIN will be sent here.
+                optional: true
     sep31:
       quotes_supported: true
       quotes_required: true
@@ -52,6 +82,7 @@ assets:
       exchangeable_assets:
         - stellar:JPYC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5
         - iso4217:USD
+    sep6_enabled: true
     sep24_enabled: true
     sep31_enabled: true
     sep38_enabled: true
@@ -100,6 +131,7 @@ assets:
       exchangeable_assets:
         - stellar:USDC:GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
         - iso4217:USD
+    sep6_enabled: false
     sep24_enabled: true
     sep31_enabled: true
     sep38_enabled: true
@@ -132,6 +164,8 @@ assets:
       buy_delivery_methods:
         - name: WIRE
           description: Have USD sent directly to your bank account.
+    sep6_enabled: false
+    sep24_enabled: true
     sep31_enabled: false
     sep38_enabled: true
   - schema: stellar
@@ -186,5 +220,7 @@ assets:
       exchangeable_assets:
         - stellar:USDC:GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
       decimals: 7
+    sep6_enabled: false
+    sep24_enabled: true
     sep31_enabled: false
     sep38_enabled: false
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -264,6 +264,23 @@ sep1:
     #
     value: |
 
+sep6:
+  ## @param: enabled
+  ## @type: bool
+  ## Whether to enable SEP-6 for this instance of AP server.
+  #
+  enabled: false
+  transactions:
+    enabled: true
+    authentication_required: true
+  transaction:
+    enabled: true
+    authentication_required: true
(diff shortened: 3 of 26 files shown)


philipliu:
The config is currently missing `more_info_url`. I'll update this PR to include it.

api-schema/src/main/java/org/stellar/anchor/api/sep/operation/Field.java
```diff
@@ -0,0 +1,13 @@
+package org.stellar.anchor.api.sep.operation;
+
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class Field {
```
lijamie98, line 9: 💡 I am interested in the reason to extract the class to separate files. Can you elaborate?

core/src/test/resources/test_assets.json
```diff
@@ -22,6 +22,53 @@
         "min_amount": 1,
         "max_amount": 1000000
       },
+      "sep6": {
+        "deposit": {
```
lijamie98, line 26: For withdraw and deposit, we do not use SEP specific configurations. This may results in inconsistency with SEP24 configuration.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -264,6 +264,34 @@ sep1:
...
+sep6:
+  ## @param: enabled
+  ## @type: bool
+  ## Whether to enable SEP-6 for this instance of AP server.
+  #
+  enabled: false
+  ## @param: transactions
+  ## Configures the GET transactions endpoint
+  #
+  transactions:
```
lijamie98, line 276: We should not configure whether to enable or disable the `GET /transactions` endpoint because it is required by SEP-6.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -264,6 +264,34 @@ sep1:
...
+    ## @param: enabled
+    ## Whether to enable the GET transactions endpoint
+    #
+    enabled: true
+  ## @param: transaction
+  ## Configures the GET transaction endpoint
+  #
+  transaction:
```
lijamie98, line 284: We should not configure whether to enable or disable the `GET /transaction` endpoint because it is required by SEP-6.

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -14,6 +14,36 @@ assets:
       fee_fixed: 0
       fee_percent: 0
       max_amount: 1000000
+    sep6:
+      deposit:
+        exchange_enabled: true
```
JakeUrban, line 19: 💡 I think we can always make the `*-exchange` endpoints available. Ideally, the fact that SEP-6 has a different set of endpoints for nonequivalent assets can be abstracted away.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -264,6 +264,34 @@ sep1:
...
+    ## @param: enabled
+    ## Whether to enable the GET transaction endpoint
+    #
+    enabled: true
+  # Configures the features flag returned by the SEP-6 /info endpoint
+  # For details, please refer to https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#response-4
+  #
+  features:
```
lijamie98, line 292: We are not supporting the `account_creation` and `claimable_balances` in the first version. We do not need to configure these flags until we decide to support them and they may need extra fields.

lijamie98 (changes requested):

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -14,6 +14,36 @@ assets:
...
+        fields:
+          email_address:
+            description: email address for transaction status updates
+            optional: true
```
JakeUrban, line 23: ❗ SEP-6 technically supports passing KYC in deposit & withdraw requests, but passing KYC data in URLs is a security concern. Lets assume all KYC data is passed via SEP-12 prior to requests to initiate transactions, and we can remove this `fields` object here.

There is a [sentence](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fields) is the spec that backs up this approach:

> If a field is part of the KYC/AML flow handled by SEP-12 [...], there's no need to list it in /info

api-schema/src/main/java/org/stellar/anchor/api/sep/operation/Field.java
philipliu: This is mostly preference, but I found that prefixing the types with `Sep*` difficult to read when they can be grouped together into separate files instead.

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -14,6 +14,36 @@ assets:
...
+              country_code:
+                description: ISO-3166-1 alpha-3 code of the user's current address
+                choices:
+                  - USA
+                  - PRI
+          cash:
+            fields:
+              dest:
+                description: your email address. Your cashout PIN will be sent here.
+                optional: true
```
JakeUrban, line 46: ❗ Same thing here -- lets remove this and assume its all configured and handled via SEP-12

core/src/test/resources/test_assets.json
philipliu: Yes, but these fields are specific to SEP-6. `fields` can be moved into the root `deposit` and `types` can be moved into the root `withdraw`, however, they are only relevant if `sep6_enabled: true`. 

I'm fine with this approach, but what about `exchange_enabled`? That is only relevant to SEP-6 at the moment.

platform/src/main/resources/config/anchor-config-default-values.yaml
philipliu: If I'm not mistaken, we don't support these in SEP-24 either but we still allow partners to configure this. I can remove this though.

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -14,6 +14,36 @@ assets:
       fee_fixed: 0
       fee_percent: 0
       max_amount: 1000000
+    sep6:
```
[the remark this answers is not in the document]
philipliu: I don't think we are validating assets.yaml at all. I can add validation though.

service-runner/src/main/resources/config/assets.yaml
philipliu: That makes sense. Shall we update the protocol to remove passing in KYC data then?

core/src/test/resources/test_assets.json
philipliu: >  but what about `exchange_enabled`? That is only relevant to SEP-6 at the moment.

If we follow Jake's suggestion below and always assume `exchange_enabled` is true, then this is not a problem.

service-runner/src/main/resources/config/assets.yaml
JakeUrban: Yes, although updating the protocol isn't time-sensitive. Can you create a PR when you have some time?

service-runner/src/main/resources/config/assets.yaml
philipliu: I think you mean only removing the KYC fields here, right? We would need to keep the financial account information (`type`, `dest`, `dest_extra`). SEP-9 does not currently support having multiple types of financial accounts.

service-runner/src/main/resources/config/assets.yaml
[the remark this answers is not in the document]
philipliu: I don't think we should define defaults because it's hard to define reasonable ones. Rather, we can have the validation catch any missing required fields.

service-runner/src/main/resources/config/assets.yaml
philipliu: Discussed this with @JakeUrban @lijamie98. We will delegate the fields to SEP-12 and ignore the `dest` and `dest_extra` parameters for now.

core/src/test/resources/test_assets.yaml
```diff
@@ -12,6 +12,9 @@ assets:
       enabled: true
       min_amount: 1
       max_amount: 10000
+      types:
```
lijamie98, line 18: It seems a lot of efforts to maintain both `json` and `yaml`. 
💡 I think we can have the `yaml` for asset configuration testing and `json` just to test the format.

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -10,6 +10,9 @@ assets:
     withdraw:
       enabled: true
       max_amount: 1000000
+      types:
```
lijamie98, line 16: Another duplication with `test_assets.yaml`. Can we simplify?

core/src/test/resources/test_assets.json
```diff
@@ -14,7 +14,11 @@
       "withdraw": {
         "enabled": true,
         "min_amount": 1,
```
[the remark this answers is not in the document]
philipliu: This would require use to remove the validation from SEP-24 implementation. I would prefer to keep the scope of this PR specific to SEP-6. I can create a Jira to track it though.

core/src/test/resources/test_assets.json
```diff
@@ -14,7 +14,11 @@
...
-        "max_amount": 10000
+        "max_amount": 10000,
+        "types": [
+          "bank_account",
```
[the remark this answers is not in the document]
philipliu: These are anchor defined types.

core/src/test/resources/test_assets.json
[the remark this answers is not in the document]
philipliu: ANCHOR-381

core/src/test/resources/test_assets.json
```diff
@@ -74,6 +80,7 @@
           "iso4217:USD"
         ]
       },
+      "sep6_enabled": true,
```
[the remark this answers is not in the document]
philipliu: ANCHOR-382

api-schema/src/main/java/org/stellar/anchor/api/sep/operation/Field.java
philipliu: > Can you help to point out which classes have the type with Sep* prefix?

For example, Sep31TxnFieldSpecs can be moved into a class called Sep31Operation.

philipliu: > These classes are specific to the AssetInfo file format. IMO, it would keep the class namespace cleaner.

Will move this into AssetInfo inner class.

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: I think we should keep them because they are used in softserve's work with Fireblocks. If fireblocks is configured then account creation is supported. Alternatively we can remove them and ask softserve to add them back as a part of their changes.

core/src/test/resources/test_assets.json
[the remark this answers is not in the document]
JakeUrban: Yes

core/src/test/resources/test_assets.json
```diff
@@ -53,10 +61,16 @@
         "fields": {
           "transaction": {
```
JakeUrban, line 62: Can we remove all uses of `fields.transaction`? We don't support this in the platform so it is confusing to require this object.

philipliu: Let's address this in a separate PR as well. I created ANCHOR-383

JakeUrban (approved):
The config interface looks good to me. I'll defer to Jamie on details of the implementation.

lijamie98 (approved):
LGTM

PR #988, opened 2023-07-12
[Anchor-244] Implement the webhook callback in the event processor

JiahuiWho:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
Implement the SEP-24 and SEP-31 callback function

1. Use OkHttp3 client
2. sign the ClientStatusCallbackHandler
3. OkHttp will try to retry a request when:
 -- The error is a connection error and there is alternative route 
 -- The request was attempted but the response indicates it is safe to retry and the request is repeatable

### Known limitations

Left the retry mechanism implementation aside and will do in a separate PR. Right now we will just keep retry until event processed successfully


platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
```diff
@@ -17,8 +29,44 @@ public ClientStatusCallbackHandler(ClientConfig clientConfig) {
...
+            .callTimeout(10, TimeUnit.MINUTES)
+            .build();
+
+    // Prepare the payload to sign
+    String currentTs = String.valueOf(TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()));
+    String domain;
+    try {
+      domain = getDomainFromURL(clientConfig.getCallbackUrl());
+    } catch (MalformedURLException e) {
+      throw new RuntimeException(e);
```
philipliu, line 46: Do we do validation on the callback URL?

JiahuiWho: We don't. `callback_url` is optional (see comments [here](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/resources/config/anchor-config-default-values.yaml#L155)). But will make change to the callsite to make sure `callback` is present

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
```diff
@@ -17,8 +29,44 @@ public ClientStatusCallbackHandler(ClientConfig clientConfig) {
...
+    // Sign the payload using the Anchor private key
+    KeyPair signer = KeyPair.fromSecretSeed(new PropertySecretConfig().getSep10SigningSeed());
+
+    // Base64 encode the signature
+    String encodedSignature = new String(Base64.getEncoder().encode(signer.sign(payload.getBytes())));
+
+    // Build the X-Stellar-Signature header
+    Request request = new Request.Builder()
+            .url(clientConfig.getCallbackUrl())
+            .header("X-Stellar-Signature", String.format("t=%s, s=%s", currentTs, encodedSignature))
```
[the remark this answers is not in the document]
JiahuiWho: Discussed with @lijamie98 about this and decided to proceed with `X-Stellar-Signature` for forward compatibility.

platform/src/main/java/org/stellar/anchor/platform/event/CallbackApiHandler.java
```diff
@@ -17,8 +24,25 @@ public class CallbackApiHandler extends EventHandler {
...
+
+    Request request = new Request.Builder()
+            .url(callbackApiConfig.getBaseUrl() + "/callback")
+            .post(buildJsonRequestBody(json(event)))
+            .build();
+
+    try {
+      Response response = client.newCall(request).execute();
+    } catch (IOException e) {
+      debugF("Callback service not available: {}", callbackApiConfig.getBaseUrl());
```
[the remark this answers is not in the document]
JiahuiWho: Discussed with @lijamie98 will move the try catch mechanism to the callsite so that the try/catch logic can be applied to both handler

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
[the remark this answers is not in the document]
philipliu: I think either set `Signature` or both `Signature` and `X-Stellar-Signature`. It's a bit strange to only supported the deprecated header. What do you think?

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
philipliu: Maybe we should do validation then? I think it's a bad experience to find out the URL is malformed only during runtime. Doesn't have to be in this PR.

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
```diff
@@ -1,24 +1,62 @@
...
+    String payload = currentTs + "." + domain + "." + eventJson;
+    // Sign the payload using the Anchor private key
+    // Base64 encode the signature
+    String encodedSignature =
+        new String(Base64.getEncoder().encode(signer.sign(payload.getBytes())));
+
+    // Build the X-Stellar-Signature header
+    return new Request.Builder()
+        .url(clientConfig.getCallbackUrl())
+        .header("X-Stellar-Signature", String.format("t=%s, s=%s", currentTs, encodedSignature))
```
lijamie98, line 58: The `X-Stellar-Signature` is a [to-be-deprecated header](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#compute-signature). 

Can we set `Signature` instead?

platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
```diff
@@ -59,6 +59,10 @@ public void start() {
       // clientsConfig
       if (eventProcessorConfig.getClientStatusCallback().isEnabled()) {
         for (ClientsConfig.ClientConfig clientConfig : clientsConfig.getClients()) {
+          if (clientConfig.getCallbackUrl().isEmpty()) {
+            continue;
```
lijamie98, line 63: Can we add a `info` call to log that the client's callback is skipped?

platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
```diff
@@ -141,7 +145,21 @@ public void run() {
...
+            boolean isProcessed = false;
+            while (!isProcessed) {
+              try{
+                eventHandler.handleEvent(event);
+                isProcessed = true;
+              } catch (Exception e) {
+                // TODO: handle retry according to response
+                // if downstream cannot consume event -> keep retry
+                // if it's code issue -> event goes to dead letter queue
+                Log.errorEx(e);
```
lijamie98, line 159: Add backoff timer in the straight-forward retry PR.

platform/src/test/kotlin/org/stellar/anchor/platform/event/ClientStatusCallbackHandlerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+  fun setUp() {
+    val clientConfig = ClientsConfig.ClientConfig()
+    clientConfig.type = ClientsConfig.ClientType.CUSTODIAL
+    clientConfig.signingKey = "GBI2IWJGR4UQPBIKPP6WG76X5PHSD2QTEBGIP6AZ3ZXWV46ZUSGNEGN2"
+    clientConfig.callbackUrl = "https://callback.circle.com/api/v1/anchor/callback"
+    handler = ClientStatusCallbackHandler(clientConfig)
+
+    secretConfig = mockk()
+    every { secretConfig.sep10SigningSeed } returns
+      "SDNMFWJGLVR4O2XV3SNEJVF53MMLQWYFYFC7HT7JZ5235AXPETHB4K3D"
```
lijamie98, line 35: I will send you some information of the secrets.

platform/src/test/kotlin/org/stellar/anchor/platform/event/ClientStatusCallbackHandlerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    val originSignature = signer.sign(payload.toByteArray())
+
+    // header example: "X-Stellar-Signature": "t=....., s=......"
+    val requestHeader =
+      handler.buildSignedCallbackRequest(signer, event).headers["X-Stellar-Signature"]
+    val parsedSignature = requestHeader?.split(", ")?.get(1)?.substring(2)
+    val decodedSignature = Base64.getDecoder().decode(parsedSignature)
+
+    Assertions.assertArrayEquals(decodedSignature, originSignature)
+  }
```
lijamie98, line 54: Please also implement the verification in SEP-24: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#verify-signature

JiahuiWho: Just for notes, I don't think the freshness can be verified here since we are not actually making the api call. Will verify the signature decoded from request against the signature re-compsoed from request info

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
JiahuiWho: The callback_url in config is optional (see comments [here](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/resources/config/anchor-config-default-values.yaml#L155), and disussion [here](https://github.com/stellar/java-stellar-anchor-sdk/pull/970#discussion_r1251134903)). I have change the callsite to check if callback_url is empty.

Validating callback_url in format won't help here because the exception thrown here is the result of `getDomainFromURL()` trying to create a `URL` object from string, we will need that try catch block anyways

lijamie98: Hey Jiahui, if `callback_url` is not empty (although it is optional), we should check the URL format in the [ClientsConfig ](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/config/ClientsConfig.java#L42)validator.

lijamie98 (approved):
Please feel free to merge after the feedbacks are taken care of.

PR #994, opened 2023-07-18
SDFSOF-67 - Updated Custody Server and Observer to use new RPC endpoint

MazurakIhor:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Updated Custody Server and Observer to use new RPC endpoint

As for now, target branch of this PR is set to fireblocks_integration but it will be changed to softserve_develop as soon as https://github.com/stellar/java-stellar-anchor-sdk/pull/1002 is approved and merged.

Changed validation of input payments (assets, amounts) to not set transaction to error status but just log message since most likely business will need to make refund of received payments if something went wrong. Business can still change state to error using notify_transaction_error action if needed.

### Why

All transaction updates should be done using new RPC actions endpoint. In scope of this task PATCH /transactions requests sent by Custody Server and Observer were replaced with corresponding RPC calls


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -711,3 +711,15 @@ data:
   #
   flyway_baseline_on_migrate: false
 
+################################
+## RPC Configuration
+################################
+rpc:
```
rkharevych, line 717: How about `rpc.actions.custom_mesages`?

platform/src/main/java/org/stellar/anchor/platform/custody/CustodyPaymentHandler.java
```diff
@@ -89,9 +119,6 @@ protected void validatePayment(JdbcCustodyTransaction txn, CustodyPayment paymen
           payment.getAssetType(),
           payment.getId(),
           payment.getExternalTxId());
-      payment.setStatus(CustodyPaymentStatus.ERROR);
```
rkharevych, line 92: I think, that we can go without this validation right now. We will update it, if it is needed. Check whether we can remove `message` and `status` from `CustodyPayment`

rkharevych (changes requested):

rkharevych (dismissed):

platform/src/main/resources/config/anchor-config-default-values.yaml
MazurakIhor: changed

platform/src/main/java/org/stellar/anchor/platform/custody/CustodyPaymentHandler.java
```diff
@@ -75,7 +73,39 @@ protected void updateTransaction(
     txn.setStatus(getCustodyTransactionStatus(payment.getStatus()).toString());
     custodyTransactionRepo.save(txn);
 
-    patchTransaction(txn, payment, newSepTxnStatus);
+    if (SepTransactionStatus.ERROR.equals(newSepTxnStatus)) {
```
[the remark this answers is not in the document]
MazurakIhor: Changed

platform/src/main/java/org/stellar/anchor/platform/custody/CustodyPaymentHandler.java
MazurakIhor: Status is still used.

Ifropc (approved):

platform/src/main/java/org/stellar/anchor/platform/custody/CustodyPaymentHandler.java
```diff
@@ -75,7 +72,27 @@ protected void updateTransaction(
...
+    } else {
+      switch (Kind.from(txn.getKind())) {
+        case DEPOSIT:
+          platformApiClient.notifyOnchainFundsSent(
+              txn.getSepTxId(),
+              payment.getTransactionHash(),
+              rpcConfig.getActions().getCustomMessages().getOutgoingPaymentSent());
+          break;
+        case WITHDRAWAL:
+        case RECEIVE:
```
Ifropc, line 87: Won't think throw an exception on platform server as SEP-31 not supported yet?

PR #998, opened 2023-07-20
[ANCHOR-369] Migrate SEP-10 to use clients configuration

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
Migrate SEP-10 to use clients configuration 

### Why
Feature implementation

### DIFF
--- a/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
+++ b/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
@@ -25,7 +25,7 @@ public interface Sep10Config {
 
   /**
    * The `home_domain` property of <a
-   * href="https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request">SEP-10<a>.
+   * href="https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request">SEP-10</a>.
    * `home_domain` value must be equal to the host of the toml file. If sep1 is enabled, toml file
    * will be hosted on the SEP server.
    *
--- a/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
+++ b/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
@@ -19,5 +19,5 @@ public interface Sep10Config {
 
   List<String> getOmnibusAccountList();
 
-  boolean isRequireKnownOmnibusAccount();
+  boolean isKnownCustodialAccountRequired();
 }
--- a/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
+++ b/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
@@ -15,8 +15,6 @@ public interface Sep10Config {
 
   Integer getJwtTimeout();
 
-  List<String> getClientAttributionDenyList();
-
   List<String> getClientAttributionAllowList();
 
   List<String> getOmnibusAccountList();
--- a/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
+++ b/core/src/main/java/org/stellar/anchor/config/Sep10Config.java
@@ -3,21 +3,91 @@ package org.stellar.anchor.config;
 import java.util.List;
 
 public interface Sep10Config {
-  String getWebAuthDomain();
 
-  String getHomeDomain();
-
-  boolean isClientAttributionRequired();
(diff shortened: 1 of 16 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -332,17 +323,14 @@ sep10:
...
+  ## @param: require_known_custodial_account
   ## @type: bool
-  ## Whether to require authenticating clients to be in the omnibusAccountList
+  ## Whether to require authenticating clients to be in the list of known custodial accounts.
+  #
+  # If the flag is set to true, the client must be one of the custodial clients defined in the clients section
+  # of this configuration file.
   #
-  require_known_omnibus_account: false
+  require_known_custodial_account: false
```
JakeUrban, line 333: This is a breaking change, right?

lijamie98: Yes... I assume the impact would be minimum if no anchors on version `2.x` are using this flag. 

WDYT?

JakeUrban: Don't think any custodial wallets have integrated with any of the SEP-24 anchors live on the platform yet, so we may be ok to make this change.

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
```diff
@@ -3,23 +3,75 @@
...
+  /**
+   * The timeout in seconds of the authenticated JSON Web Token. An expired JWT will be rejected.
+   * This is the timeout period after the client has authenticated.
+   *
+   * @return the timeout in seconds of the authenticated JSON Web Token
+   */
   Integer getJwtTimeout();
 
-  List<String> getClientAttributionDenyList();
+  boolean isRequireKnownOmnibusAccount();
```
JakeUrban, line 62: Don't we need to change these methods to match the config?

lijamie98: Ah... missed it in PR merge. It is now fixed.

philipliu (approved):
LGTM

Ifropc (approved):

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
```diff
@@ -3,23 +3,75 @@
...
+  /**
+   * Set if the client attribution is required. Client Attribution requires clients to verify their
+   * identity by passing a domain in the challenge transaction request and signing the challenge
+   * with the ``SIGNING_KEY`` on that domain's SEP-1 stellar.toml. See the SEP-10 section `Verifying
+   * Client Application Identity` for more information, refer to <a
+   * href="https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#verifying-client-application-identity">SEP-10</a>.
+   *
+   * @return true if the client attribution is required
+   */
   boolean isClientAttributionRequired();
```
JakeUrban, line 36: Lets add that `isClientAttributionRequired()` is only relevant for noncustodial wallets, and `isKnownCustodialAccountRequired()` is reciprocal method for custodial wallets.

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
```diff
@@ -3,23 +3,75 @@
...
+  /**
+   * The timeout in seconds of the authenticated JSON Web Token. An expired JWT will be rejected.
+   * This is the timeout period after the client has authenticated.
+   *
+   * @return the timeout in seconds of the authenticated JSON Web Token
+   */
   Integer getJwtTimeout();
 
-  List<String> getClientAttributionDenyList();
+  boolean isRequireKnownCustodialAccount();
```
JakeUrban, line 62: nit: can we do `isKnownCustodialAccountRequired()`?

platform/src/main/resources/config/anchor-config-default-values.yaml
JakeUrban: nit: can we match the phrasing of `client_attribution_required` and use `known_custodial_account_required`?

lijamie98: `custodial_account_required`? WDYT?

PR #999, opened 2023-07-20
[ANCHOR-359] Implement SEP-6 info endpoint

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR implements the SEP-6 info endpoint.

### Why

We are implementing SEP-6.

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -0,0 +1,95 @@
...
+                  .description("type of deposit to make")
+                  .choices(methods)
+                  .build();
+
+          DepositAssetResponse deposit =
+              DepositAssetResponse.builder()
+                  .enabled(true)
+                  .authenticationRequired(true)
+                  .fields(ImmutableMap.of("type", type))
+                  .build();
```
JakeUrban, line 68: `type` is optional in the spec today, should we be checking if `methods` is empty and omitting the `type` field if so? Alternatively, we should have validation that there is always at least one type provided in the config.

JakeUrban (approved):
Looks great other than the one comment on deposit types.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: > Alternatively, we should have validation that there is always at least one type provided in the config.

We took this approach. If SEP-6 and deposit/withdraw is enabled, AP will require that at least one method is specified. The validation is done as part of the `AssetService` initialization. https://github.com/stellar/java-stellar-anchor-sdk/blob/96d215de79e3acc51fb65b691a2e376b57e51aa3/core/src/test/kotlin/org/stellar/anchor/asset/DefaultAssetServiceTest.kt#L73-L99

PR #1011, opened 2023-07-27
[ANCHOR-393] Set JsonProperty for nested AnchorEvent types

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This adds the `JsonProperty` annotation to the nested `AnchorEvent` types so that Kafka's `JsonSerializer` can correctly serialize camel case fields.

### Why

Field names are incorrectly serialized for some parts of the AnchorEvent.

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/shared/StellarPayment.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/shared/StellarPayment.java
@@ -1,5 +1,6 @@
 package org.stellar.anchor.api.shared;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.gson.annotations.SerializedName;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


JakeUrban (approved):

api-schema/src/main/java/org/stellar/anchor/api/shared/StellarPayment.java
```diff
@@ -18,19 +19,24 @@ public class StellarPayment {
 
   Amount amount;
 
+  @JsonProperty("payment_type")
```
lijamie98, line 22: Can we stick with gson?

lijamie98 (changes requested):
Can we stick with `gson`?

api-schema/src/main/java/org/stellar/anchor/api/shared/StellarPayment.java
philipliu: We are mixing Gson and Jackson (in Spring and Kafka) throughout the project. This bug is one example where using both caused us some pain. Using Gson here would mean writing a new SerDe for it, which may get thrown away if we move everything to Jackson.

I want to get this patch release out so we can unblock our partners, but I'll create a Jira to address this serialization issue long term.

PR #1016, opened 2023-07-28
[ANCHOR-394] Add whitelist to SEP-10 and add the client name field.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

- Add `client_allow_list` to SEP-10
- Add name field to client config
- When `client_allow_list` is not defined, the domains of all NONCUSTODIAL clients are allowed if `domain` is defined.


platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -155,14 +164,14 @@ void validateConfig(Errors errors) {
 
   void validateClientAttribution(Errors errors) {
     if (clientAttributionRequired) {
-      if (ListHelper.isEmpty(getClientAttributionAllowList())) {
+      if (ListHelper.isEmpty(getDefaultAllowClientDomain())) {
```
Ifropc, line 167: What if anchor support only custodial wallets?

Ifropc (approved):

platform/src/test/kotlin/org/stellar/anchor/platform/config/Sep10ConfigTest.kt
```diff
@@ -75,19 +78,44 @@ class Sep10ConfigTest {
...
     config.validateClientAttribution(errors)
     assertFalse(errors.hasErrors())
-    println(config.clientAttributionAllowList)
-    println(config.knownCustodialAccountList)
   }
 
   @Test
-  fun `test both client attribution deny and allow lists are empty`() {
-    // Create a cofnig with empty ClientsConfig lists
+  fun `test validation of empty client allow list when client attribution is required`() {
```
philipliu, line 84: Can we add tests for all the validations?

lijamie98: Call `config.validate()` only tests the all validations are called in `config.validate()` function. If we think it is necessary, a test that `config.validate()` call all validation can be added. 

This is unit testing the `validateClientAttribution` function.

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
lijamie98: The validation is only called when the client attribution is used. The custodial wallet does not require client attribution.

platform/src/test/kotlin/org/stellar/anchor/platform/config/Sep10ConfigTest.kt
lijamie98: If we still feel strong about this, we can add another test that `config.validate()` calls all validations in the future PR.

PR #1021, opened 2023-08-01
Skip spotless tasks in Github actions to accelerate the CI

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

Skip spotless tasks in Github actions to accelerate the CI

### Why

Optimization

### DIFF
--- a/.github/workflows/sub_gradle_test_and_build.yml
+++ b/.github/workflows/sub_gradle_test_and_build.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Gradle test and build. (unit tests, integration tests, end-2-end tests and build)
         env:
           run_docker: false
-        run: ./gradlew clean build --no-daemon --stacktrace -x spotlessApply -x spotlessCheck -x spotlessKotlin -x spotlessKotlinApply -x javadoc -x javadocJar -x sourcesJar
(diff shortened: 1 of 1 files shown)


.github/workflows/sub_gradle_test_and_build.yml
```diff
@@ -47,7 +51,7 @@ jobs:
       - name: Gradle test and build. (unit tests, integration tests, end-2-end tests and build)
         env:
           run_docker: false
-        run: ./gradlew clean build --no-daemon --stacktrace
+        run: ./gradlew clean build --no-daemon --stacktrace -x spotlessApply -x spotlessCheck -x spotlessKotlin -x spotlessKotlinApply -x javadoc -x javadocJar -x sourcesJar
```
philipliu, line 54: I think we can skip `spotlessApply`, but not check. If a developer tries to submit unformatted code, we should be catching it in CI.

lijamie98: @philipliu If `spotlessCheck` fails, does it fail the CI?

philipliu: Sorry, I missed this earlier, but we should keep `spotlessKotlin` too right?

lijamie98: No problem at all. Thanks for catching these gotchas. I had removed `-x spotlessKotlin`.

PR #1035, opened 2023-08-03
[ANCHOR-355] Implement SEP-6 deposit

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This implements the SEP-6 `GET /deposit` endpoint _mostly_ according to the following sequence diagram.

```mermaid
sequenceDiagram
  participant Wallet
  participant Platform
  participant Anchor

  Wallet->>Platform: GET /deposit
  Platform-->>Platform: Create deposit transaction with id abcd-1234 with status pending_anchor
  Platform-->>Wallet: Returns transaction with deposit instructions missing
  Platform--)Anchor: Sends an AnchorEvent with type transaction_created and transaction_id abcd-1234
  Anchor->>Platform: PATCH /transaction with status pending_customer_info_needed, type=sep6-deposit-usd-small and KYC fields
  Wallet->>Platform: GET /transaction with id abcd-1234
  Platform-->>Wallet: Returns transaction with status pending_customer_info_needed, type=sep6-deposit-usd-small and KYC fields
  Wallet->>Platform: POST [SEP-12]/customer?type=sep6-deposit-usd-small with missing fields
  loop Over all ongoing transactions submitted by the user
    Platform->>Anchor: Sends an AnchorEvent with transaction_id abcd-1234 and type customer_updated
    Anchor->>Anchor: Verifies provided KYC fields are sufficient for transaction
    Anchor-->>Platform: PATCH /transaction with status pending_user_transfer_start and deposit instructions in transaction object
  end
  Wallet->>Platform: GET /transaction with id abcd-1234 to fetch deposit instructions
  Platform-->>Wallet: Returns transaction with status pending_user_transfer_start and deposit instructions
```

Where the implementation deviates is how `CUSTOMER_UPDATED` anchor events are structured. This implementation sends out 1 event whenever a SEP-12 customer is updated instead of fanning them out a `TRANSACTION_STATUS_CHANGED` event for each ongoing transaction submitted by the customer. This was done to keep the scope of this change small, but we can consider whether to move this into the platform later.

### Why

SEP-6 implementation.

### Known limitations

SEP-12 will now emit CUSTOMER_UPDATED events which need to be put behind a feature flag to prevent possible deserialization issues in the business server.


core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -56,8 +67,19 @@ public Sep12PutCustomerResponse putCustomer(Sep10Jwt token, Sep12PutCustomerRequ
...
+    // Only publish event if the customer was updated.
+    eventSession.publish(
+        AnchorEvent.builder()
+            .id(UUID.randomUUID().toString())
+            .sep("12")
+            .type(AnchorEvent.Type.CUSTOMER_UPDATED)
+            .customer(CustomerUpdatedResponse.builder().account(request.getAccount()).build())
+            .build());
+
+    return PutCustomerResponse.to(response);
```
JakeUrban, line 82: So is the idea that the flow for the business would be as follows?

1. receives `PUT /customer` callback request, validates and persists KYC information for the customer, and returns
2. consumes an event from the queue that a customer was updated, finds all transactions pending for the customer, and updates their statuses

This feels fairly opinionated to me, because we're effectively suggesting to the business to process the same event twice. The business is made aware of the customer update via the callback and can decide on their own how they want to manage updating the customers' pending transactions. 

I would imagine most anchors will update the customer's pending transactions synchronously in the request/response cycle of the `PUT /customer` call, in which case they don't need to listen for the event. 

If they want to update transactions asynchronously, which is probably the better approach, I guess monitoring the queue is a reasonable way to do that. Given that we're replacing the practice of reading directly from the queue with receiving HTTP webhooks, are we going to update the reference server's implementation?

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -56,8 +67,19 @@ public Sep12PutCustomerResponse putCustomer(Sep10Jwt token, Sep12PutCustomerRequ
...
+    PutCustomerResponse response =
+        customerIntegration.putCustomer(PutCustomerRequest.from(request));
+
+    // Only publish event if the customer was updated.
+    eventSession.publish(
+        AnchorEvent.builder()
+            .id(UUID.randomUUID().toString())
+            .sep("12")
+            .type(AnchorEvent.Type.CUSTOMER_UPDATED)
+            .customer(CustomerUpdatedResponse.builder().account(request.getAccount()).build())
```
JakeUrban, line 79: I don't think this is the right interface for the event. We want to move away from using Stellar accounts to identify the customer. Can we use the customer ID returned in the response instead?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -1,38 +1,118 @@
...
+    // Pre-validation
+    if (token == null) {
+      throw new SepNotAuthorizedException("missing token");
+    }
+    if (request == null) {
+      throw new SepValidationException("missing request");
+    }
+    if (!request.getAccount().equals(token.getAccount())) {
+      throw new SepNotAuthorizedException("account does not match token");
+    }
```
JakeUrban, line 62: This should be allowed for custodial wallets that have the account from the request whitelisted as a destination account.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -1,38 +1,118 @@
...
+            .id(UUID.randomUUID().toString())
+            .sep("6")
+            .type(AnchorEvent.Type.TRANSACTION_CREATED)
+            .transaction(TransactionHelper.toGetTransactionResponse(txn))
+            .build());
+
+    return GetDepositResponse.builder()
+        .how("Check the transaction for more information about how to deposit.")
+        .id(txn.getId())
+        .build();
```
JakeUrban, line 113: The `200` response in SEP-6 is meant to be a "no additional info necessary" response, but you're using the `how` attribute to guide the developer to checking the transaction for more information. We could go with this approach, but I think we would need to update the SEP to clarify that 200 doesn't mean "ready to receive funds".

I think we should consider using the [customer_info_needed](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#3-customer-information-status) response instead. We would need to add the ID attribute to the response, but this response type is more accurate to the situation. What do you think?

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
philipliu: > So is the idea that the flow for the business would be as follows?
> 1. receives PUT /customer callback request, validates and persists KYC information for the customer, and returns
> 2. consumes an event from the queue that a customer was updated, finds all transactions pending for the customer, and updates their statuses

Yes, that's right.

> I would imagine most anchors will update the customer's pending transactions synchronously in the request/response cycle of the PUT /customer call, in which case they don't need to listen for the event.

Hmm... I'd imagine that asking anchors to update their existing integrations would introduce _more_ adoption friction, but I don't have the experience with talking to anchors to say otherwise. If they want to handle deposits synchronously in the `PUT /customer` request path, I don't think emitting an `AnchorEvent` would prevent them from doing so.

> Given that we're replacing the practice of reading directly from the queue with receiving HTTP webhooks, are we going to update the reference server's implementation?

We could. Although, I don't think this is blocking for the SEP-6 launch.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: That's a good idea. Thanks for the suggestion!

JakeUrban: Hmm actually, I think I like your approach more 😅. I don't think it makes sense to return 403 and a transaction ID in the response body.

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
JakeUrban: Yes it doesn't need to block your progress, just a thought. Lets keep the approach you defined because it doesn't prevent businesses from updating deposit transactions synchronously  👍

philipliu:
Thanks for your review @JakeUrban. I've updated the PR to address your comments.

api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
```diff
@@ -80,7 +80,23 @@ public class PlatformTransactionData {
...
+  @SerializedName("required_info_updates")
+  String requiredInfoUpdates;
+
+  @SerializedName("required_customer_info_message")
+  String requiredCustomerInfoMessage;
+
+  @SerializedName("required_customer_info_updates")
+  String requiredCustomerInfoUpdates;
+
+  String instructions;
```
philipliu, line 95: This should be a structured type. I'll update it when I address the comments.

kotlin-reference-server/src/main/resources/default-config.yaml
```diff
@@ -12,6 +12,7 @@ app:
   distributionWallet: GBN4NNCDGJO4XW4KQU3CBIESUJWFVBUZPOKUZHT7W7WRB7CWOA7BXVQF
   distributionWalletMemo:
   distributionWalletMemoType:
+  secret: SAJW2O2NH5QMMVWYAN352OEXS2RUY675A2HPK5HEG2FRR2NXPYA4OLYN
```
lijamie98, line 15: Does this require an account in the testnet? If yes, we need to add the key creation to the [useful-scirpts](https://github.com/stellar/useful-scripts/tree/main/anchor-platform-testnet-reset) repository.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/di/ReferenceServerContainer.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.stellar.reference.event.event
+import org.stellar.reference.plugins.RequestExceptionHandlerPlugin
+import org.stellar.reference.plugins.RequestLoggerPlugin
+import org.stellar.reference.sep24.sep24
+import org.stellar.reference.sep24.testSep24
+
+const val AUTH_CONFIG_ENDPOINT = "endpoint-auth"
+
+object ReferenceServerContainer {
+  private val config = ConfigContainer.getInstance().config
```
lijamie98, line 32: ❓ Why call these a `container`?

lijamie98 (approved):
Small comments. LGTM.

lijamie98:
Actually, we talked about having `sep-6` to be in a feature branch. WDYT?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/di/ReferenceServerContainer.kt
philipliu: I think DI frameworks usually refer to a group of dependencies as a container.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -1,38 +1,115 @@
...
+      throws AnchorException {
+    // Pre-validation
+    if (token == null) {
+      throw new SepNotAuthorizedException("missing token");
+    }
+    if (request == null) {
+      throw new SepValidationException("missing request");
+    }
+
+    AssetInfo asset = assetService.getAsset(request.getAssetCode());
```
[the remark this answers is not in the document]
philipliu: `assetCode` is a `@NonNull` field. https://github.com/stellar/java-stellar-anchor-sdk/pull/1035/files#diff-df0463cc298733ddfaa78de1a3147301970a4fc25e0254d9f529d5f3e9c29cf5R17-R19

It's also a required parameter by the controller.

kotlin-reference-server/src/main/resources/default-config.yaml
philipliu: Yeah, I'll update the scripts.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -1,38 +1,115 @@
...
+    Sep6TransactionBuilder builder =
+        new Sep6TransactionBuilder(txnStore)
+            .id(id)
+            .transactionId(id)
+            .status(SepTransactionStatus.INCOMPLETE.toString())
+            .kind(Sep6Transaction.Kind.DEPOSIT.toString())
+            .type(request.getType())
+            .assetCode(request.getAssetCode())
+            .assetIssuer(asset.getIssuer())
+            .amountOut(request.getAmount())
```
Ifropc, line 84: We use amount in jwt in sep-24, should we use amount_requested here?
Because the final amount_out may differ from what user requested

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -1,38 +1,115 @@
...
+    try {
+      KeyPair.fromAccountId(request.getAccount());
+    } catch (RuntimeException ex) {
+      throw new SepValidationException(String.format("invalid account %s", request.getAccount()));
+    }
+    Memo memo = MemoHelper.makeMemo(request.getMemo(), request.getMemoType());
+    String id = SepHelper.generateSepTransactionId();
+
+    Sep6TransactionBuilder builder =
+        new Sep6TransactionBuilder(txnStore)
```
Ifropc, line 76: We store client_domain in sep24, is it not needed to store for sep-6?
(I'm guessing not -> we can get in from SEP-10 token)

core/src/main/java/org/stellar/anchor/util/TransactionHelper.java
```diff
@@ -47,6 +48,35 @@ public static GetTransactionResponse toGetTransactionResponse(Sep31Transaction t
...
+        .updatedAt(txn.getUpdatedAt())
+        .completedAt(txn.getCompletedAt())
+        .message(txn.getMessage())
+        .refunds(txn.getRefunds())
+        .stellarTransactions(txn.getStellarTransactions())
+        .sourceAccount(txn.getFromAccount())
+        .destinationAccount(txn.getToAccount())
+        .externalTransactionId(txn.getExternalTransactionId())
+        .memo(txn.getMemo())
+        .memoType(txn.getMemoType())
```
Ifropc, line 71: refund_memo is missing

kotlin-reference-server/src/main/kotlin/org/stellar/reference/ReferenceServer.kt
```diff
@@ -1,73 +1,36 @@
...
 
-  val cfgBuilder = ConfigLoaderBuilder.default()
-  // Add environment variables as a property source.
-  cfgBuilder.addPropertySource(PropertySource.environment())
-  envMap?.run { cfgBuilder.addMapSource(this) }
-  // Add config file as a property source if valid
-  locationCfg.fold({}, { cfgBuilder.addFileSource(it.ktReferenceServerConfig) })
-  // Add default config file as a property source.
-  cfgBuilder.addResourceSource("/default-config.yaml")
+  Thread {
```
Ifropc, line 21: Start a coroutine instead?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/di/EventConsumerContainer.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.kafka.clients.consumer.ConsumerConfig
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.common.serialization.StringDeserializer
+import org.stellar.reference.event.EventConsumer
+import org.stellar.reference.event.processor.AnchorEventProcessor
+import org.stellar.reference.event.processor.InMemoryTransactionStore
+import org.stellar.reference.event.processor.NoOpEventProcessor
+import org.stellar.reference.event.processor.Sep6EventProcessor
+
+object EventConsumerContainer {
```
Ifropc, line 12: Can we listen to webhooks instead of kafka events?
If webhook is not yet supported, can you add a ticket to get rid of listening to kafka events on the business server?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -0,0 +1,204 @@
...
+      )
+    runBlocking {
+      patchTransaction(
+        PlatformTransactionData.builder()
+          .id(event.transaction.id)
+          .status(SepTransactionStatus.PENDING_ANCHOR)
+          .build()
+      )
+    }
+    activeTransactionStore.addTransaction(customer.id, event.transaction.id)
```
Ifropc, line 67: What is activeTransactionStore for? Can't we always get transaction from AP?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+
+  private fun submitStellarTransaction(
+    source: String,
+    destination: String,
+    asset: Asset,
+    amount: String
+  ): StellarTransaction {
+    val account = server.accounts().account(source)
```
Ifropc, line 172: [nit] you can use wallet SDK for it, it will be 5 lines of code instead of 20

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -0,0 +1,204 @@
...
+              )
+              .stellarTransactions(listOf(stellarTxn))
+              .build()
+          )
+        }
+      }
+    }
+  }
+
+  private fun toStandardFormat(asset: String): String {
```
Ifropc, line 157: [nit] `asset.toAssetId()`

kotlin-reference-server/src/main/kotlin/org/stellar/reference/ReferenceServer.kt
philipliu: Hm.... the `KafkaConsumer` API is blocking so is coroutine the right choice here?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/di/EventConsumerContainer.kt
philipliu: Created ANCHOR-439, will make the change after this PR is merged.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/ReferenceServer.kt
Ifropc: If we remove it it doesn't matter :)
But in general you just dispatch a coroutine and it can work in the background. Whatever happens in coroutine can be blocking

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: `amount_requested` doesn't exist, but I think `amount_expected` is what you meant? If so then yeah, we should use that. I think I made an assumption that business server would amend the `amount_out` when they process the transaction.

Ifropc: Yes

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
philipliu: Yeah, that would work too. If it's all the same to you, let's keep the current implementation to minimize additional changes.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: I think so, but SEP-31 also stores it. Let's add it later if we find we need it.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
philipliu: Ah, nice. I'll mark this as a TODO for now and get to it in ANCHOR-439.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
Ifropc: I think we can remove it later to keep it cleaner. If anybody look at the code they would think they should store transaction data on the business server

philipliu: Yeah, I'm going to remove it in ANCHOR-439.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -154,12 +154,12 @@ class Sep6EventProcessor(
     }
   }
 
-  private fun toStandardFormat(asset: String): String {
-    val parts = asset.split(":")
+  private fun String.toAssetId(): String {
```
Ifropc, line 157: I meant that String.toAssetId is already in the wallet SDK :)

Ifropc (approved):

PR #1043, opened 2023-08-05
[ANCHOR-403] Prevent SSRF through SEP-1 TOML redirect

reecexlm:
This PR addresses SDP-001. If an error is encountered while fetching stellar.toml via http, the exception is obfuscated with an error message to avoid ability to use the method to leak aws instance metadata via ssrf.  

### PR Structure
  otherwise).   

### Thoroughness

To test this I used the Anchor Platform e2e test.  I updated the profile/config.env to use stellar toml type url and value to be localhost:9000
```
sep1.toml.type=url
sep1.toml.value=http://localhost:9000/.well-known/stellar.toml
```
I ran the local flask code which performed redirect to itself and simulated a metadata endpoint.  

First to reproduce the error I received the exception in the to local flask app. can see the redirect and then returns a 200 for the fake metadata.
```
❯ python3 app1.py
 * Serving Flask app 'app1'
 * Debug mode: off
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on all addresses (0.0.0.0)
 * Running on http://127.0.0.1:9000
 * Running on http://192.168.1.106:9000
Press CTRL+C to quit
127.0.0.1 - - [09/Aug/2023 15:26:30] "GET /.well-known/stellar.toml HTTP/1.1" 302 -
127.0.0.1 - - [09/Aug/2023 15:26:30] "GET /hostname HTTP/1.1" 200 -
```
the fake metadata sent looks like this
```
{"amiId":"ami-12345678","instanceId":"i-1234567890abcdef0","instanceType":"t2.micro","region":"us-west-2"}
```

the sep service had exception in toml parse
```
AnchorPlatformEnd2EndTest > initializationError FAILED
    java.lang.IllegalStateException: Key is not followed by an equals sign on line 1: {"amiId":"ami-12345678","instanceId":"i-1234567890abcdef0","instanceType":"t2.micro","region":"us-west-2"}
        at com.moandjiezana.toml.Toml.read(Toml.java:140)
        at org.stellar.anchor.util.Sep1Helper$TomlContent.<init>(Sep1Helper.java:25)
        at org.stellar.anchor.util.Sep1Helper.parse(Sep1Helper.java:13)
        at org.stellar.anchor.platform.AbstractIntegrationTest.setupTests(AbstractIntegrationTest.kt:44)
```
Throwing an exception with a customer error message obfuscates the exception message, preventing the leak
```
AnchorPlatformEnd2EndTest > initializationError FAILED
    java.lang.RuntimeException: Failed to parse TOML content
        at org.stellar.anchor.util.Sep1Helper.parse(Sep1Helper.java:17)
        at org.stellar.anchor.platform.AbstractIntegrationTest.setupTests(AbstractIntegrationTest.kt:44)
        at org.stellar.anchor.platform.AbstractIntegrationTest.setUp(AbstractIntegrationTest.kt:34)
        at org.stellar.anchor.platform.AnchorPlatformEnd2EndTest$Companion.construct(AnchorPlatformEnd2EndTest.kt:14)
        at org.stellar.anchor.platform.AnchorPlatformEnd2EndTest.construct(AnchorPlatformEnd2EndTest.kt)
```

### DIFF
--- a/core/src/main/java/org/stellar/anchor/util/NetUtil.java
+++ b/core/src/main/java/org/stellar/anchor/util/NetUtil.java
@@ -1,6 +1,5 @@
 package org.stellar.anchor.util;
 
-import static org.stellar.anchor.util.Log.debugF;
 import static org.stellar.anchor.util.StringHelper.isEmpty;
 
 import io.jsonwebtoken.lang.Strings;
@@ -20,10 +19,12 @@ public class NetUtil {
 
     // Check if response was successful (status code 200)
     if (!response.isSuccessful()) {
-      throw new IOException("Server returned HTTP error code " + response.code());
+      throw new IOException(String.format("Unable to fetch data from %s", url));
     }
 
-    if (response.body() == null) return "";
+    if (response.body() == null) {
+      throw new IOException("Response body is empty.");
+    }
     return Objects.requireNonNull(response.body()).string();
   }
 
--- a/core/src/main/java/org/stellar/anchor/util/NetUtil.java
+++ b/core/src/main/java/org/stellar/anchor/util/NetUtil.java
@@ -1,5 +1,6 @@
 package org.stellar.anchor.util;
 
+import static org.stellar.anchor.util.Log.debugF;
 import static org.stellar.anchor.util.StringHelper.isEmpty;
 
 import io.jsonwebtoken.lang.Strings;
@@ -15,7 +16,14 @@ import okhttp3.Response;
 public class NetUtil {
   public static String fetch(String url) throws IOException {
     Request request = OkHttpUtil.buildGetRequest(url);
+    debugF("request:{}", request.toString());
     Response response = getCall(request).execute();
+    debugF("response:{}", response.toString());
+
+    // Check if response was successful (status code 200)
+    if (!response.isSuccessful()) {
+      throw new IOException("Server returned HTTP error code " + response.code());
+    }
 
     if (response.body() == null) return "";
     return Objects.requireNonNull(response.body()).string();
--- a/core/src/main/java/org/stellar/anchor/util/NetUtil.java
+++ b/core/src/main/java/org/stellar/anchor/util/NetUtil.java
@@ -16,9 +16,7 @@ import okhttp3.Response;
 public class NetUtil {
   public static String fetch(String url) throws IOException {
     Request request = OkHttpUtil.buildGetRequest(url);
-    debugF("request:{}", request.toString());
     Response response = getCall(request).execute();
-    debugF("response:{}", response.toString());
 
     // Check if response was successful (status code 200)
     if (!response.isSuccessful()) {
--- a/core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
+++ b/core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
@@ -10,7 +10,13 @@ public class Sep1Helper {
   }
 
   public static TomlContent parse(String tomlString) {
-    return new TomlContent(tomlString);
+    try {
+      return new TomlContent(tomlString);
+    } catch (Exception e) {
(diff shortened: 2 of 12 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -219,19 +219,19 @@ sep1:
...
     ##     `string`: value contains the content of the stellar toml file
     ##     `file`: value contains the path to the stellar toml file
     ##     `url`: value contains the url to the stellar toml file
-    type: string
+    type: url
     ## @param: value
     ## @type: string
     #
-    value: |
+    value: http://169.254.169.254/latest/meta-data/
```
lijamie98, line 234: We should not use the hard-coded URL in the `anchor-config-default-values.yaml`. 

If this is for test, it should be configured in the test cases instead of the default values.

core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -17,7 +17,14 @@ public static String fetch(String url) throws IOException {
     Request request = OkHttpUtil.buildGetRequest(url);
     Response response = getCall(request).execute();
 
-    if (response.body() == null) return "";
+    // Check if response was successful (status code 200)
+    if (!response.isSuccessful()) {
+      throw new IOException(String.format("Unable to fetch data from %s", url));
```
lijamie98, line 22: I am not sure how this would fix the SSRF issue. But if you think it is the fix, please feel free to merge.

lijamie98 (changes requested):

core/src/main/java/org/stellar/anchor/util/NetUtil.java
reecexlm: just to be consistent. returning "" will be more confusing i think. if the response is empty, we want to throw an exception as well.

philipliu (approved):
LGTM

core/src/main/java/org/stellar/anchor/util/NetUtil.java
reecexlm: also, this code is within a method that is responsible for handling I/O operations, i feel appropriate to throw an IOException. This would signal to callers that an I/O error occurred, which could include an empty response body from a network operation.  that is throwing, IOException, is typically used to signal an error related to Input/Output operations, such as reading or writing to files or network resources. In this context, we are checking if the response body is empty, thus an appropriate exception to throw in the context of reading data from a network resource.

core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -17,7 +17,14 @@ public static String fetch(String url) throws IOException {
...
+    }
```
JakeUrban, line 23: Are we sure we want to throw an exception in this case? We aren't interested in the response body of a 400 transaction?

reecexlm: previously we were just returning "".   I agree with you though. Will you be ok if i file a ticket for this and make a bug fix for this?

core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
```diff
@@ -10,7 +10,13 @@ public static TomlContent readToml(String url) throws IOException {
   }
 
   public static TomlContent parse(String tomlString) {
-    return new TomlContent(tomlString);
+    try {
+      return new TomlContent(tomlString);
+    } catch (Exception e) {
+      // obfuscate exception message to prevent metadata leaks
```
Ifropc, line 16: Should we still log it on ERR level? Otherwise it's hard to debug what went wrong

PR #1044, opened 2023-08-07
[ANCHOR-378] Status callback retry implementation.

lijamie98:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

The event delivery should retry in one of the 3 ways:

1. If connection error (IOException) happens, we will retry 3 times with `networkBackoffTimer`. After that we will skip the event and proceed to next sub-sequent events.

1. If the business server returns HTTP status code other than 200s or 300s, we will retry 3 times with `httpErrorBackoffTimer`. After 3 retries, we will skip the event and proceed to next sub-sequent events.

1. If the event processor encounters an un-expected error caused by the un-seen bugs, the event will be delivered to a no-op DLQ class which has Log.error implementation. 

### Why

Implement reliable retry.

### Known limitations

Needs unit tests.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
@@ -1,5 +1,6 @@
 package org.stellar.anchor.platform.event;
 
+import static java.lang.Thread.currentThread;
 import static org.stellar.anchor.event.EventService.*;
 import static org.stellar.anchor.util.Log.*;
 import static org.stellar.anchor.util.MetricConstants.*;
@@ -76,7 +77,8 @@ public class EventProcessorManager {
           new EventProcessor(
               CALLBACK_API_EVENT_PROCESSOR_NAME,
               EventQueue.TRANSACTION,
-              new CallbackApiEventHandler(callbackApiConfig)));
+              new CallbackApiEventHandler(callbackApiConfig),
+              eventService));
     }
     // Create a processor of the client status callback handler for each client defined in the
     // clientsConfig
@@ -112,7 +114,8 @@ public class EventProcessorManager {
                     clientConfig,
                     sep24TransactionStore,
                     assetService,
-                    moreInfoUrlConstructor)));
+                    moreInfoUrlConstructor),
+                eventService));
       }
     }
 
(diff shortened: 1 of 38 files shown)


platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
```diff
@@ -74,43 +77,45 @@ public void start() {
...
-            continue;
-          }
+              new CallbackApiEventHandler(callbackApiConfig),
+              eventService));
+    }
+    // Create a processor of the client status callback handler for each client defined in the
+    // clientsConfig
+    if (eventProcessorConfig.getClientStatusCallback().isEnabled()) {
+      for (ClientsConfig.ClientConfig clientConfig : clientsConfig.getClients()) {
+        if (clientConfig.getCallbackUrl().isEmpty()) {
```
JiahuiWho, line 87: Nit:  if( **_!_** clientConfig.getCallbackUrl().isEmpty()) {...}

JiahuiWho (approved):

JakeUrban (commented):
I know we've discussed this a few times, but I'm still not sure it makes sense to have different retry behavior depending on if we experience a connection error or http error. AWS SNS via HTTP doesn't seem to distinguish between these cases.

Unrelatedly, I noticed that `BaseApiClient`'s `client` object's timeouts seem way too long. I would have expected something around 15 seconds, rather than 10 minutes.

lijamie98:
@philipliu Please feel free to take care of Jake's comments and submit commits during my absence.

JiahuiWho:
Just curious that how important it is if we missed delivering an event? With current implementation, we could miss delivering an event with http error, but why not with connection error (since we keep retrying until succeed)? If clients service is down, they should be able to pull all events  during the downtime by themselves?

JakeUrban:
I think thats the core of the argument I'm trying to make; it is the client & business' responsibility to ensure events are received and processed correctly, and if they don't, assuming it is not the fault of the anchor platform itself due to an unexpected error, the client or business should have a system in place to ensure it eventually gets that update.

When we built the original version of the SDP, we listened for Circle's SNS notifications to get updates on payments we initiated via their API. In case we missed an event or didn't receive it, we also had a reoccurring check for transactions that hadn't been updated within a timeframe we felt was reasonable. If we found transactions that hadn't been updated in the configured amount of time, we'd make requests to Circle's API to get the latest status.

Now, clients and businesses wouldn't have to implement that logic if the Anchor Platform retried indefinitely for all events, but right now it is only making that guaranteed for a subset of events. I don't think retrying indefinitely is a viable approach in the long-term, so I think we should use the same logic we have for http errors for connection errors.

JiahuiWho:
Thanks for clarification. I will take care of this

JakeUrban (approved):
This looks good @JiahuiWho ! Can you update the PR description to accurately describe the functionality? After that we should be good to merge 🚀

PR #1046, opened 2023-08-08
[ANCHOR-407] Fix client config override

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

All configs are loaded into a map before being sent to Spring. Currently, client configs are being overwritten at the client level with the format `clients[0].property` which will not overwrite the `clients` property from the default config. This commit fixes the property merging logic such that if it discovers an indexed property, it will attempt to remove any non-indexed property with the same name.

### Why

This is a bug in the configuration override logic.

### Known limitations

The fix is ugly but I didn't want to spend too much time refactoring the configurator. A better fix would be to set the overriding properties at different precedence levels and just let Spring handle it.

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
@@ -5,16 +5,19 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import lombok.Data;
-import lombok.Getter;
 import org.stellar.anchor.api.exception.InvalidConfigException;
 
 public class ConfigMap {
-  @Getter int version;
+  int version;
   final Map<String, ConfigEntry> data;
 
   // ConfigMap keys will be in normalized form (dot separated hierarchy)
   public ConfigMap() {
     data = new HashMap<>();
+  }
+
+  public int getVersion() {
+    return version;
   }
 
   public void setVersion(int version) {
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
@@ -1,21 +1,20 @@
 package org.stellar.anchor.platform.configurator;
 
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import lombok.Data;
+import lombok.Getter;
 import org.stellar.anchor.api.exception.InvalidConfigException;
 
 public class ConfigMap {
-  int version;
+  @Getter int version;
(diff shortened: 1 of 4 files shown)


platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
```diff
@@ -70,7 +72,16 @@ public Map<String, String> toStringMap() {
   }
 
   public void merge(ConfigMap config) {
+    // TODO: instead of merging, create a property source with higher precedence
+    // Matches any string of the form: <listName>[<index>].<elementName>
```
Ifropc, line 76: How does it work with .yaml files?

Ifropc, line 76: what about my.property[0]?

philipliu: YAML will produce a config map such that the keys of the map are the flattened property names. 

For example, this will create the key `platform.server`:
```
platform:
  server: 8085
```

platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
```diff
@@ -69,8 +71,19 @@ public Map<String, String> toStringMap() {
         .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getValue()));
   }
 
+  // TODO: instead of merging, create a property source with higher precedence
   public void merge(ConfigMap config) {
+    // Matches any string of the form: <listName>[<index>].<elementName>
+    // <listName> can be a dot separated hierarchy of names.
+    Pattern pattern =
+        Pattern.compile("^([a-zA-Z0-9_]+(?:\\.[a-zA-Z0-9_]+)*)\\[(\\d+)](?:\\.[a-zA-Z0-9_]+)*$");
```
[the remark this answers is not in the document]
Ifropc: Agree

platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigMap.java
Ifropc: I'm more curious about arrays, will it also create `client[0].name` property?

philipliu: It does.

```yaml
clients:
  - name: referenceCustodial
    type: custodial
    signing_key: GDJLBYYKMCXNVVNABOE66NYXQGIA5AC5D223Z2KF6ZEYK4UBCA7FKLTG
    callback_url: http://wallet-server:8092/callbacks
```
will create 
```json
        "clients[0].name": {
            "value": "referenceCustodial",
            "source": "DEFAULT"
        },
...
```

PR #1051, opened 2023-08-10
ANCHOR-403 Refactor TOML Exception obfuscation for URL/Parsing to Toml Functions our of NetUtil

reecexlm:
This PR moves the exception obfuscation for the stellar toml file handling to the Toml module and out of NetUtil.  The reason is that NetUtil fetch is a lower level function that is shared.  Obfuscating the exception message may not be applicable for other usages.  THis PR also logs the actual exception message so that it is available for troubleshooting. Added some additional test coverage for Sep1Helper as well.

### PR Structure

* [x ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
  otherwise).
* [ x] This PR's title starts with name of package that is most changed in the PR, ex.
  packages.


core/src/main/java/org/stellar/anchor/sep1/Sep1Service.java
```diff
@@ -46,6 +45,17 @@ public Sep1Service(Sep1Config sep1Config) throws IOException, InvalidConfigExcep
     }
   }
 
+  private String fetchTomlFromURL(String url) throws IOException {
+    try {
+      return NetUtil.fetch(url);
+    } catch (IOException e) {
+      // Obfuscate the message and rethrow
+      String obfuscatedMessage = String.format("Unable to fetch data from %s", url);
+      debugF(obfuscatedMessage); // Log the obfuscated message using the debugF method
```
Ifropc: If response is not successful this will produce 3 debug logs with the same message

reecexlm: ah yea, im logging at each layer. let me look at that. thanks!

core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
```diff
@@ -1,24 +1,43 @@
...
+          obfuscatedMessage); // Preserve the original exception as the cause
+    }
   }
 
   public static class TomlContent {
     private final Toml toml;
 
     TomlContent(URL url) throws IOException {
-      String tomlValue = NetUtil.fetch(url.toString());
-      toml = new Toml().read(tomlValue);
```
philipliu, line 21: It looks like we are handling the exceptions in both in the `TomlContent` constructor and the `Sep1Helper` methods. Do you think it would be cleaner to move all the exception handling into `Sep1Helper` and removing `TomlContent(URL url)`?

It would look something like this:
```java
public class Sep1Helper {
    public static TomlContent readToml(String url) throws IOException {
        try {
            String tomlValue = NetUtil.fetch(url);
            return TomlContent.parse(tomlValue);
        } catch (IOException e) {
            debugF("Failed to fetch toml file from {} with message: {}", url, e.toString());
            throw new IOException("Failed to fetch toml file from " + url);
        }
    }

    public static TomlContent parse(String tomlString) throws InvalidConfigException {
        try {
            return new TomlContent(tomlString);
        } catch (Exception e) {
            debugF("Failed to parse toml file with message: {}", e.toString());
            throw new ("Failed to parse TOML content. Invalid Config.");
        }
    }
}
```

core/src/main/java/org/stellar/anchor/sep1/Sep1Service.java
```diff
@@ -50,6 +49,17 @@ public Sep1Service(Sep1Config sep1Config) throws IOException, InvalidConfigExcep
     }
   }
 
+  private String fetchTomlFromURL(String url) throws IOException {
```
philipliu, line 52: I don't think it's necessary to obfuscate here, right? If we throw an exception in the `Sep1Service` constructor, the Spring application would fail to initialize.

build.gradle.kts
```diff
@@ -135,7 +135,7 @@ subprojects {
 
 allprojects {
   group = "org.stellar.anchor-sdk"
-  version = "2.1.2"
+  version = "2.1.3"
```
philipliu, line 138: I don't think we bumped the version or updated the badges in `main`. Can we do that an merge `main` into `develop`?

philipliu (commented):
I think the changes are correct I just have some comments about code structure.

core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
reecexlm: nice one

core/src/main/java/org/stellar/anchor/sep1/Sep1Service.java
reecexlm: Yea I think your right. :thumbsup:

core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -13,12 +14,30 @@
...
+      // Check if response was unsuccessful (ie not status code 2xx) and throw IOException
+      if (!response.isSuccessful()) {
+        debugF(message);
+        throw new IOException(message);
+      }
+
+      if (response.body() == null) {
+        debugF(message);
+        throw new IOException(message);
+      }
```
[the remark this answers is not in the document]
reecexlm: I think if we use Objects.requireNonNull we will throw an NPE instead of being able to throw a IOException which I think is more appropriate in this context?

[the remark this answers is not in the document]
reecexlm: ah i see your comment. yea, the previous code had both checks. i think we still need the check for null and throwing the IOException but no longer need Objects.requireNotNull

core/src/main/java/org/stellar/anchor/util/Sep1Helper.java
```diff
@@ -1,26 +1,39 @@
...
+import org.stellar.anchor.api.exception.InvalidConfigException;
 
 public class Sep1Helper {
   public static TomlContent readToml(String url) throws IOException {
-    return new TomlContent(new URL(url));
+    try {
+      String tomlValue = NetUtil.fetch(url);
+      return new TomlContent(tomlValue);
+    } catch (IOException e) {
+      String obfuscatedMessage =
```
Ifropc, line 15: Exception message is swallowed, we should log it. Same for method below

reecexlm: whoops i think i confused myself. meant to log the exception. :) fixed that. thanks

core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -1,24 +1,43 @@
...
-    return Objects.requireNonNull(response.body()).string();
+    try {
+      Request request = OkHttpUtil.buildGetRequest(url);
+      Response response = getCall(request).execute();
+      String message =
+          String.format("Error fetching from URL: %s response: %s", url, response.toString());
+
+      // Check if response was unsuccessful (ie not status code 2xx) and throw IOException
+      if (!response.isSuccessful()) {
+        debugF(message);
```
philipliu, line 27: Nit: I guess if we're going to throw the exception with the same message, there's no need to log it.

platform/src/test/kotlin/org/stellar/anchor/Sep1ServiceTest.kt
```diff
@@ -66,9 +69,58 @@ class Sep1ServiceTest {
...
     val config = PropertySep1Config(true, TomlConfig(URL, mockAnchorUrl))
     sep1 = Sep1Service(config)
     assertEquals(sep1.stellarToml, stellarToml)
   }
+
+  // this test is not expected to raise an exception. given the re-direct to a malicious
+  // endpoint still returns a 200 the exception will be raised/obfuscated
+  // when the toml is parsed.
+  @Test
+  fun `getStellarToml fetches invalid data during malicious re-direct`() {
```
philipliu, line 81: Nit: I think the intention is to make it obvious this _can_ happen, but it seems a bit weird given that this is something we would want to prevent. Maybe we could change the test name to something like `getStellarToml fetches data during re-direct`. What do you think?

philipliu (approved):
A couple of nits but otherwise LGTM.

core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -7,18 +7,39 @@
...
+
     Request request = OkHttpUtil.buildGetRequest(url);
     Response response = getCall(request).execute();
 
-    if (response.body() == null) return "";
-    return Objects.requireNonNull(response.body()).string();
+    // Check if response was unsuccessful (ie not status code 2xx) and throw IOException
+    if (!response.isSuccessful()) {
+      throw new IOException(response.toString());
+    }
```
JakeUrban, line 35: Are we sure don't want to return 4xx or 5xx response bodies?

If this function only has one reference at the moment, I would suggest changing the return value to include both the response body and status code, that way callers can decide for themselves if they're interested in non-success response bodies.

PR #1052, opened 2023-08-11
ANCHOR-403 Obfuscate Exception in Sep1Helper instead of NetUtil

reecexlm:
This PR moves the exception obfuscation for the stellar toml file handling to the Toml module and out of NetUtil. The reason is that NetUtil fetch is a lower level function that is shared. Obfuscating the exception message may not be applicable for other usages. THis PR also logs the actual exception message so that it is available for troubleshooting. Added some additional test coverage for Sep1Helper as well.

PR Structure
[x ] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
 This PR avoids mixing refactoring changes with feature changes (split into two PRs
otherwise).
[ x] This PR's title starts with name of package that is most changed in the PR, ex.
paymentservice.stellar, or all or doc if the changes are broad or impact many
packages.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/util/NetUtil.java
+++ b/core/src/main/java/org/stellar/anchor/util/NetUtil.java
@@ -7,25 +7,44 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Objects;
 import okhttp3.Call;
 import okhttp3.Request;
 import okhttp3.Response;
 
 public class NetUtil {
+
+  /**
+   * Fetches the content from the specified URL using an HTTP GET request.
+   *
+   * <p>This method expects a response body to be present in the HTTP response. If the response is
+   * unsuccessful (i.e., not a 2xx status code) or if the response body is null, an IOException will
+   * be thrown.
+   *
+   * @param url The URL to fetch content from.
+   * @return The content of the response body as a string.
+   * @throws IOException If the response is unsuccessful, or if the response body is null.
+   */
   public static String fetch(String url) throws IOException {
+
     Request request = OkHttpUtil.buildGetRequest(url);
     Response response = getCall(request).execute();
 
-    // Check if response was successful (status code 200)
+    // Check if response was unsuccessful (ie not status code 2xx) and throw IOException
     if (!response.isSuccessful()) {
-      throw new IOException(String.format("Unable to fetch data from %s", url));
+      throw new IOException(
+          "Unsuccessful response code: " + response.code() + ", message: " + response.message());
     }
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/stellar/anchor/util/NetUtil.java
```diff
@@ -7,25 +7,44 @@
...
+   */
   public static String fetch(String url) throws IOException {
+
     Request request = OkHttpUtil.buildGetRequest(url);
     Response response = getCall(request).execute();
 
-    // Check if response was successful (status code 200)
+    // Check if response was unsuccessful (ie not status code 2xx) and throw IOException
     if (!response.isSuccessful()) {
-      throw new IOException(String.format("Unable to fetch data from %s", url));
```
philipliu, line 22: Nit: It might be difficult to extract the code from a String. It might be better to introduce a custom Exception class that represents unsuccessful responses and include the code and body as a field. I believe we are only using `NetUtil` in a couple of places in the code though so we don't really need to block on this IMO.

platform/src/test/kotlin/org/stellar/anchor/Sep1ServiceTest.kt
```diff
@@ -73,11 +74,21 @@ class Sep1ServiceTest {
     assertEquals(sep1.stellarToml, stellarToml)
   }
 
+  // this test is not expected to raise an exception. given the re-direct to a malicious
+  // endpoint still returns a 200 the exception will be raised/obfuscated
+  // when the toml is parsed.
   @Test
-  fun `getStellarToml throws exception when redirect is encountered`() {
+  fun `getStellarToml fetches invalid data during malicious re-direct`() {
```
philipliu, line 81: Nit: maybe call this `getStellarToml fetches data during redirect`?

philipliu (approved):
Minor comments. LGTM

Can we make sure the versions are updated as well? https://github.com/stellar/java-stellar-anchor-sdk/commit/26b3057fc1d50e58b9a05266f7b012b2083a1158

core/src/main/java/org/stellar/anchor/util/NetUtil.java
reecexlm: thanks.  i will file a ticket for the custom exception class due to time constraints and as well jakes comment about return value.  can accomodate and improve this together.

reecexlm: recapturing his comment here
JakeUrban2 hours ago
Are we sure don't want to return 4xx or 5xx response bodies?

If this function only has one reference at the moment, I would suggest changing the return value to include both the response body and status code, that way callers can decide for themselves if they're interested in non-success response bodies.

PR #1060, opened 2023-08-22
[ANCHOR-295] Implement Kotlin reference server

philipliu:
<details>

### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR implements the callback endpoints in the Kotlin reference server. It moves all the tests to use the `sep24` service runner profile, which points to the Kotlin reference server. Another PR will delete the `anchor-reference-server` project. 

### Why

The Java reference server is deprecated.

### Known limitations

N/A


kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/ConfigureAuth.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.ktor.server.auth.jwt.*
+import io.ktor.server.response.*
+import org.stellar.reference.data.AuthSettings
+import org.stellar.reference.data.Config
+
+fun Application.configureAuth(cfg: Config) {
+  when (cfg.authSettings.type) {
+    AuthSettings.Type.JWT ->
+      authentication {
+        jwt("integration-auth") {
```
JakeUrban, line 17: ❓ Should we name this something different considering we'd use the same auth type in our persistent deployment eventually?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/rate/RateRoute.kt
```diff
@@ -0,0 +1,53 @@
...
+          .sellAsset(call.parameters["sell_asset"])
+          .sellAmount(call.parameters["sell_amount"])
+          .sellDeliveryMethod(call.parameters["sell_delivery_method"])
+          .buyAsset(call.parameters["buy_asset"])
+          .buyAmount(call.parameters["buy_amount"])
+          .buyDeliveryMethod(call.parameters["buy_delivery_method"])
+          .countryCode(call.parameters["country_code"])
+          .expireAfter(call.parameters["expire_after"])
+          .clientId(call.parameters["client_id"])
+          .id(call.parameters["id"])
```
JakeUrban, line 36: ❓ `id` is not a parameter to our `GET /rate` callback, is this used by our integration tests?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/customer/CustomerRoute.kt
```diff
@@ -0,0 +1,85 @@
...
+          call.respond(HttpStatusCode.NoContent)
+        } catch (e: NotFoundException) {
+          call.respond(HttpStatusCode.NotFound, e)
+        } catch (e: Exception) {
+          call.respond(HttpStatusCode.InternalServerError)
+        }
+      }
+    }
+    route("/invalidate_clabe") {
+      get("{id}") {
```
JakeUrban, line 72: ❓ I'm assuming this is used to setup an integration test? Any chance we can make the endpoint available only when deployed in that context? We could do this later when we add a persistent deployment too, so I'll leave it to you.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/rate/RateService.kt
```diff
@@ -0,0 +1,231 @@
...
+import org.stellar.reference.dao.QuoteRepository
+import org.stellar.reference.model.Quote
+
+class RateService(private val quoteRepository: QuoteRepository) {
+  private val scale = 4
+  fun getRate(request: GetRateRequest): GetRateResponse {
+    val rate =
+      when {
+        request.id != null -> getRate(request.id)
+        request.sellAmount != null || request.buyAmount != null -> {
```
JakeUrban, line 25: ❓ I'm not sure I understand why we have this condition, won't `validate_request()` catch the case where this condition is not true?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/uniqueaddress/UniqueAddressService.kt
```diff
@@ -0,0 +1,54 @@
...
+            "Invalid distributionWalletMemoType: ${settings.distributionWalletMemoType}"
+          )
+      }
+    }
+  }
+  fun getUniqueAddress(request: GetUniqueAddressRequest): GetUniqueAddressResponse {
+    val builder =
+      GetUniqueAddressResponse.UniqueAddress.builder().stellarAddress(settings.distributionWallet)
+
+    if (settings.distributionWallet.isBlank() || settings.distributionWalletMemoType.isBlank()) {
```
JakeUrban, line 44: ❓ I think we should be checking if `settings.distributionWalletMemo` is blank, not the wallet address, right?

JakeUrban (commented):
🙌 I left a few questions but this looks great. I'm interested to hear your perspective on what its like to integrate with the anchor platform's interfaces generally.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/ConfigureAuth.kt
lijamie98: Change to `endpoint-auth`.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/rate/RateRoute.kt
lijamie98: The `id` is used to store quote at the reference server. I think we need more time to discuss whether to store the quote at AP or at the business server.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/customer/CustomerRoute.kt
lijamie98: I put a comment in the code to consider enabling this endpoint only when testing.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/rate/RateService.kt
lijamie98: This is to simplify the `validate_request()`. The validate request does not check the condition where both amounts are `null`. So, we need this condition outside `validate_request()` function. I think this is legit.

JakeUrban: Rather than have one validation check outside the request, why can't this check be the first within the `validate_request()` function?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/uniqueaddress/UniqueAddressService.kt
lijamie98: fixed.

JakeUrban (approved):

PR #1062, opened 2023-08-23
[ANCHOR-413] Add jacoco test report to PR workflow

JiahuiWho:
### Description

This is to add jacococ test report to PR workflow, to monitor test coverage change

### Testing

After PR being published, there should be a github-action bot comment in your PR with test coverage.

Please see https://github.com/stellar/java-stellar-anchor-sdk/pull/1081#issuecomment-1703260373 for example.
Note that this PR is published from a branch on original repo instead of a forked repo.

### Known limitations

1. For now Github only grant apps write permission (which is needed to make comment) to original repo, and read permission to forked repo, that mean this action will only works on PR created from original repo. I have created a ticket https://stellarorg.atlassian.net/browse/ANCHOR-433 to follow up on this  and try to find a work around for forked repo
For more info about github permission please refer to https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

2. Currently the min-coverage-overall is set to 40 and min-coverage-changed-files is 60, this is just for testing, will be changed accordingly.


.github/workflows/sub_gradle_test_and_build.yml
```diff
@@ -51,7 +54,18 @@ jobs:
...
+        run: ./gradlew clean build jacocoTestReport --no-daemon --stacktrace -x spotlessApply -x spotlessKotlinApply -x javadoc -x javadocJar -x sourcesJar
+
+      - name: Add coverage to PR
+        id: jacoco
+        uses: madrapps/jacoco-report@v1.6.1
+        with:
+          paths: ${{ github.workspace }}/**/build/reports/jacoco/test/jacocoTestReport.xml
+          token: ${{ secrets.GITHUB_TOKEN }}
+          min-coverage-overall: 40
+          min-coverage-changed-files: 60
```
philipliu, line 66: Can this check be moved into the gradle build instead? That way, the developer can get feedback locally without publishing a PR.

build.gradle.kts
```diff
@@ -36,6 +38,7 @@ subprojects {
 
   /** Specifies JDK-11 */
   java { toolchain { languageVersion.set(JavaLanguageVersion.of(11)) } }
+  jacoco { toolVersion = "0.8.10" }
```
philipliu, line 41: nit: Can we move the versions into [`libs.versions.toml`](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/gradle/libs.versions.toml)?

build.gradle.kts
```diff
@@ -62,6 +65,15 @@ subprojects {
     }
 
     kotlin { ktfmt("0.42").googleStyle() }
+
+    tasks.jacocoTestReport {
+      dependsOn(tasks.test) // tests are required to run before generating the report
+      reports {
+        xml.required.set(true)
+        csv.required.set(false)
+        html.required.set(false)
```
philipliu, line 74: Let's enable HTML reports too. I think it would be nice to be able to view a rendered report in the browser.

.github/workflows/sub_gradle_test_and_build.yml
```diff
@@ -9,6 +9,9 @@ jobs:
   gradle_test_and_build:
     name: Gradle Test and Build
     runs-on: ubuntu-22.04
+    permissions:
```
philipliu, line 12: nit: Can we comment that this is required for the coverage report?

.github/workflows/sub_gradle_test_and_build.yml
philipliu: Also, do we want to set coverage minimums by project? I think it would make sense to have the `core` project require higher coverage than the reference server, for example.

philipliu (commented):
The changes look good, just a few minor comments.

.github/workflows/sub_gradle_test_and_build.yml
JiahuiWho: > Can this check be moved into the gradle build instead? That way, the developer can get feedback locally without publishing a PR.

Yes The JacocoCoverageVerification task can be used to verify if code coverage metrics are met based on configured rules. However the build fails if any of the configured rules are not met and JaCoCo only reports the first violated rule. I will gradually implemented in the following PR.

Right now you can simply run jacocoTestReport in local to get an report but without the delta.

> Also, do we want to set coverage minimums by project? I think it would make sense to have the `core` project require higher coverage than the reference server, for example.

I think that's part of the first question where we want enforce more granular test coverage rules.

philipliu: > Yes The JacocoCoverageVerification task can be used to verify if code coverage metrics are met based on configured rules. However the build fails if any of the configured rules are not met and JaCoCo only reports the first violated rule. I will gradually implemented in the following PR.

So once we enable `JacocoCoverageVerification`, do we still want to set `min-coverage-overall` and `min-coverage-changed-files` in the Github action?

lijamie98 (approved):
LGTM

PR #1065, opened 2023-08-25
[ANCHOR-416] Remove `more_info_url` from sep24 database table

lijamie98:
### Description

Remove `more_info_url` from sep24 database table. 

### Context

This field is not used.

### Testing

./gradlew test


JakeUrban (approved):

Ifropc:
I believe we should also remove it from SEP24 POJO

lijamie98:
> I believe we should also remove it from SEP24 POJO

The good thing is `PojoSep24Transaction` class does not have the `moreInfoUrl` field.

PR #1068, opened 2023-08-28
[ANCHOR-380] Add client_id to `getRate()` in `Sep31Service`

lijamie98:
### Description

Fix the bug where `client_id` was not passed to `getRate()` when the customer of the wallet exists.

### Context
Bug fix

### Testing

./gradlew test


core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -248,6 +258,21 @@ public GetPriceResponse getPrice(Sep38GetPriceRequest getPriceRequest) throws An
...
+      GetCustomerResponse customer =
+          customerIntegration.getCustomer(
+              GetCustomerRequest.builder().account(token.getAccount()).build());
+      return customer.getId();
+    } catch (NotFoundException nfex) {
+      // skip only if customer not found.
+      debugF("Unable to get customer {} from customer integration server.", token.getAccount());
+    }
+    return null;
+  }
```
JakeUrban, line 274: I think there is a misunderstanding of what `clientId` is. Its not the ID of the customer requesting the quote.

From the documentation:

> Client IDs will be the Stellar public key the public key used to authenticate via SEP-10. Anchors must ensure that > the public key specified in the request matches a public key known to belong to the sending anchor

This is only true for SEP-31 though. In SEP-24 & SEP-6, the client ID is either the Stellar public key of the wallet OR the wallet's domain.

JakeUrban: @Ifropc wants to replace usage of Stellar public keys with domains, even for custodial applications. For 1.x, we probably need to continue using Stellar public keys, but for 2.x we could use home domains if we require the `clients.domain` field for each client.

Ifropc: To clarify, not domains, but client name that's defined in the config

Ifropc: From the spec
> An identifier for the client application making the request

So I believe using `client.name` is appropriate here, as initially the idea of introducing `client.name` was to use it as application identification

lijamie98: Oh, yeah, I was wondering about this and forgot to check the documentation. 

I think the name `client_id` is very confusing though. Should we change to `client_account`? or `client_stellar_account`?

lijamie98: I am not familiar with the context of the `client.name`. Can we add a Jira ticket for discussion?

platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
```diff
@@ -212,9 +212,15 @@ Sep38Service sep38Service(
...
       Sep38QuoteStore sep38QuoteStore,
       EventService eventService) {
     return new Sep38Service(
-        sep38Config, assetService, rateIntegration, sep38QuoteStore, eventService);
+        sep38Config,
+        assetService,
+        rateIntegration,
+        customerIntegration,
+        sep38QuoteStore,
+        eventService);
```
JakeUrban, line 224: We don't need `customerIntegration` here anymore, right?

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -358,11 +368,17 @@ public Sep38QuoteResponse postQuote(Sep10Jwt token, Sep38PostQuoteRequest reques
...
-            .build();
-    GetRateResponse.Rate rate = this.rateIntegration.getRate(getRateRequest).getRate();
+            .expireAfter(request.getExpireAfter());
+
+    // Sets the clientId if the customer exists
+    String clientId = getClientIdFromToken(token);
+    if (clientId != null) {
+      getRateRequestBuilder.clientId(clientId);
+    }
+    GetRateRequest getRateRequest = getRateRequestBuilder.build();
```
JakeUrban, line 378: Firm quotes will always be authenticated, so we don't need to call `getClientIdFromToken()`. If the token is null, we should throw an error, rather than handle it gracefully.

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -507,4 +524,9 @@ private String getTotalPrice(BigDecimal bSellAmount, BigDecimal bBuyAmount) {
 
     return formatAmount(bTotalPrice, pricePrecision);
   }
+
+  private String getClientIdFromToken(Sep10Jwt token) {
+    if (token == null) return null;
+    return token.getAccount();
+  }
```
JakeUrban, line 531: This works until we add support for using SEP-38 prices & quotes for SEP-24 & SEP-6 transactions. If we want to replace the usage of public keys and domains with client names, we can't merge this because it would be a breaking change to change `clientId` from a public key to names defined in the config. cc @Ifropc

JakeUrban (changes requested):

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
lijamie98: The check of the authentication with the token is always done at the controller. We don't need to worry about them in the service layer. It makes simpler not to worry about authentication in the service layer.

platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
lijamie98: That's right. Changed.

core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
```diff
@@ -215,7 +215,7 @@ public GetPriceResponse getPrice(Sep38GetPriceRequest getPriceRequest) throws An
       validateAmountLimit("sell_", sellAmount, sendMinLimit, sendMaxLimit);
     }
 
-    GetRateRequest request =
+    GetRateRequest.GetRateRequestBuilder rrBuilder =
```
philipliu, line 218: nit: rename this to `getRateRequestBuilder` since that's what we call it elsewhere.

core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
```diff
@@ -615,13 +613,11 @@ internal class Sep10ServiceTest {
         .build()
 
     assertDoesNotThrow { sep10Service.createChallenge(cr) }
-    verify(exactly = 1) { sep10Config.isKnownCustodialAccountRequired }
     verify(exactly = 2) { sep10Config.knownCustodialAccountList }
   }
 
   @Test
   fun `Test createChallenge() when isKnownCustodialAccountRequired is not enabled`() {
```
philipliu, line 620: This test name seems to be incorrect.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -508,13 +517,40 @@ void updateFee() throws SepValidationException, AnchorException {
...
             .getFee();
     infoF("Fee for request ({}) is ({})", request, fee);
     Context.get().setFee(fee);
   }
 
+  String getClientName() throws BadRequestException {
+    return getClientName(Context.get().getSep10Jwt().getAccount());
+  }
+
+  String getClientName(String account) throws BadRequestException {
```
philipliu, line 531: Is `Sep31Service` the best place for this class?

JiahuiWho: This function is to find the name of a client (whose domain is listed in SEP-10 config) through clientsConfig (listed in SEP-31). So it has to be here, or we can move it to a helper class

lijamie98 (commented):
LGTM

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -508,13 +517,40 @@ void updateFee() throws SepValidationException, AnchorException {
...
+      }
+    } else {
+      // client attribution not required
+      if (client == null) {
+        return null;
+      }
+    }
+    return client.getName();
+  }
+
```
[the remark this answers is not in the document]
JiahuiWho: The second if check is no longer necessary since we put validation in PropertySep10Config to make sure `getClientAttributionAllowList` (renamed to `getAllowedClientDomains`) is never empty and always returns the intersection of `clientAllowList` and clientsConfig.clilents list

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
```diff
@@ -67,22 +67,18 @@ public interface Sep10Config {
   boolean isClientAttributionRequired();
 
   /**
-   * Get the list of allowed client domains if the client attribution is required.
+   * Get the list of allowed client domains.
    *
    * @return the list of allowed client domains.
    */
-  List<String> getClientAttributionAllowList();
+  List<String> getAllowedClientDomains();
```
Ifropc, line 74: Isn't it a breaking change?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -557,13 +564,29 @@ void updateFee() throws SepValidationException, AnchorException {
...
+    if (sep10Config.isClientAttributionRequired() && client == null) {
+      throw new BadRequestException("Client not found");
+    }
+    if (client != null &&
+        !sep10Config.getAllowedClientDomains().contains(client.getDomain())) {
+      client = null;
+    }
+    return client == null ? null : client.getName();
+  }
+
```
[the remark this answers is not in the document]
JiahuiWho: I think this is still missing the logic that, even if we find the client, we should still return null if it's not in getAllowedClientDomains()

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
JiahuiWho: This should works just fine. Only the function name changed, the name of the config field stays`clientAllowList`,

JakeUrban:
Please don't force-push if avoidable @JiahuiWho

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
Ifropc: How did you achieve that?

JiahuiWho: Can you elaborate more on the question? We didn't touch the field `clientAllowList` in Sep10Config..

The `getClientAttributionAllowList` returns a list of allowed client domains (which is the intersection of `sep10config.clientAllowList` and `clientsConfig.clients`), since we are introducing a new function to return the name of same group of clients, it's better to make the naming

JiahuiWho (approved):

JakeUrban (approved):

core/src/main/java/org/stellar/anchor/config/Sep10Config.java
Ifropc: Sorry didn't see SEP10 config is an interface

Ifropc (approved):

PR #1071, opened 2023-08-30
[ANCHOR-295] Cleanup Kotlin referencer server implementation

philipliu:
### Description

The changes include:
1. Introduces a new plugin to catch and map exceptions to a HTTP error code.
2. Put the `/invalidate_clabe` endpoint behind a configuration.
3. Re-introduces the SEP-12 validation from stellar-anchor-tests. Previously, when a GET /customer request does not include an account/memo, the query does not assert that they are null. This caused a race when updating an account that maps to many customers.

### Context

Address the comments from the previous PR https://github.com/stellar/java-stellar-anchor-sdk/pull/1060.

### Testing

`./gradlew test`

### Known limitations

N/A

### DIFF
--- a/kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/ConfigureRouting.kt
+++ b/kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/ConfigureRouting.kt
@@ -10,6 +10,7 @@ import org.stellar.reference.callbacks.fee.fee
 import org.stellar.reference.callbacks.interactive.sep24Interactive
 import org.stellar.reference.callbacks.rate.RateService
 import org.stellar.reference.callbacks.rate.rate
+import org.stellar.reference.callbacks.test.testCustomer
 import org.stellar.reference.callbacks.uniqueaddress.UniqueAddressService
 import org.stellar.reference.callbacks.uniqueaddress.uniqueAddress
 import org.stellar.reference.dao.JdbcCustomerRepository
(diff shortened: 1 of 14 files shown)


JakeUrban (approved):
The changes here look great, but I think we're missing one thing. We're still trying to fetch an `id` parameter from the get rate request, but the parameter is not passed by the platform.

philipliu:
@JakeUrban I purposely omitted that change as it requires architectural changes. You and @ lijamie98 will have a discussion later this week to decide what to do here. We can address this after your meeting.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/plugins/ConfigureRouting.kt
```diff
@@ -53,4 +54,7 @@ fun Application.configureRouting(cfg: Config) = routing {
   if (cfg.sep24.enableTest) {
     testSep24(helper, depositService, withdrawalService, cfg.sep24.interactiveJwtKey)
   }
+  if (cfg.appSettings.isTest) {
```
lijamie98, line 57: Do we need to keep two settings for `sep24` and `appSettings`. They seems to serve the same purpose.

philipliu: Hm...not really? I will merge them.

philipliu: Actually, I think we need to keep them. If we are deploying the reference server, we would want to enable the SEP-24 routes.

lijamie98: Hm... I guess the configuration `sep24.enableTest` is confusing from my (user's) perspective.... It actually simulate the anchor's sep-24 flow.

lijamie98 (approved):

PR #1090, opened 2023-09-05
ANCHOR-427: add status filter and use destination accounts for matching transactions in observer

JakeUrban:
### Description

This PR does two things:

- The queries to fetch transactions from the database now uses the status that the transaction should be in as an additional filter in addition to the account and memo filters
- They also check for matches using the destination account of the payment rather than the source account of the payment

### Context

There is no reason to check for matches against transactions that are not in the status that indicates the transaction is ready to receive the payment.

We also need to match against the destination account because that is the account we're monitoring. The source accounts of payments are unpredictable.

### Testing

`./gradlew test`

### Known limitations

We'll need to port this to 1.x

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ subprojects {
     mavenLocal()
     mavenCentral()
     maven { url = uri("https://packages.confluent.io/maven") }
-    maven { url = uri("https://reposdeitory.mulesoft.org/nexus/content/repositories/public/") }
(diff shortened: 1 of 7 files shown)


lijamie98:
@JakeUrban Do you mind to add [ANCHOR-427] to the title? https://stellarorg.atlassian.net/jira/software/c/projects/ANCHOR/boards/47/backlog?view=detail&selectedIssue=ANCHOR-427&issueLimit=100&search=observer

lijamie98 (approved):
LGTM

PR #1091, opened 2023-09-05
[Anchor-428] Add refund memo to platform transaction

Ifropc:
### Description

Add missing refund memo to the platform trasaction 

### Known limitations

N/A


JakeUrban:
Can we add some tests that verify they're present where they should be?

Ifropc:
I don't think we have a test for GetTransaction at the moment

JakeUrban (approved):

PR #1103, opened 2023-09-08
Softserve develop sync

MazurakIhor:
### Description

Fix build error

### Context

Build fails with an error `e: file:///home/runner/work/java-stellar-anchor-sdk/java-stellar-anchor-sdk/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/EventService.kt:30:20 Unresolved reference: AnchorEvent`

### Testing

`./gradlew test`

### DIFF
--- a/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/EventService.kt
+++ b/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/EventService.kt
@@ -28,7 +28,7 @@ class EventService {
     if (txnId != null) {
       // filter events with txnId
       return receivedEvents.filter {
-        it.type != AnchorEvent.Type.QUOTE_CREATED.type && it.payload.transaction.id == txnId
+        it.type != "quote_created" && it.payload.transaction.id == txnId
       }
     }
     // return all events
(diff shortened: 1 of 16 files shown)


kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/EventService.kt
```diff
@@ -27,7 +27,7 @@ class EventService {
     if (txnId != null) {
       // filter events with txnId
       return receivedEvents.filter {
-        it.type != AnchorEvent.Type.QUOTE_CREATED.type && it.payload.transaction.id == txnId
+        "quote_created" != it.type && it.payload.transaction.id == txnId
```
JakeUrban, line 30: Hey @MazurakIhor, I think the unresolved reference error you got is better resolved by importing `AnchorEvent`, rather than hard-coding the event type here.

JakeUrban (changes requested):

JakeUrban:
I pushed the import fix to softserve-develop, so we don't need this PR anymore, closing.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/EventService.kt
MazurakIhor: You're right. It's weird that https://github.com/stellar/java-stellar-anchor-sdk/pull/1102 didn't catch this issue.
Thank you for fixing this issue!

PR #1104, opened 2023-09-11
[ANCHOR-440] Add @Param to the status function parameter

lijamie98:
### Description

Add @Param to the status function parameter to fix the Spring database error.

```
[java.lang.IllegalArgumentException:Either use @Param on all parameters except Pageable and Sort typed once, or none at all](https://stackoverflow.com/questions/43489490/java-lang-illegalargumentexceptioneither-use-param-on-all-parameters-except-pa)
```
### Testing

`./gradlew test`


JiahuiWho (approved):
LGTM

PR #1105, opened 2023-09-12
[ANCHOR-439] Remove transaction store from reference server

philipliu:
### Description

This removes the `ActiveTransactionStore` from the Kotlin reference server. Previously, this was used to track which transactions were pending KYC info updates. Now, we request the transactions from the platform.

### Context

This is part of a series of commits to clean up the reference server implementation.

### Testing

`./gradlew test`

### Known limitations

N/A


lijamie98:
Can we change the branch name to: `feature/anchor-439-remove-java-reference-server`?

philipliu:
@lijamie98 I would need to open a new PR. This branch won't be long-lived and it's in my personal fork, so I don't think the branch naming conventions need to apply here.

PR #1106, opened 2023-09-12
ANCHOR-401: Allow non-default destination account

Ifropc:
### Description

This change adds list of `destination_accounts` into the clients configuration (only for type `custodial`)

On SEP-24 deposit, we validate that account either matches SEP-10 (previous impl), or is in `destination_accounts` list for this client (wallet)

### Context

Currently, if wallet wants to use a different destination account for deposits (different from SEP-10 authenticated account), it will raise an error.

While it makes sense for non-custodial case (otherwise, we allow to transfer money to non-KYCd accounts), it’s a blocker for custodial case.

Custodial wallets may have 1 or more hot wallets, separated from their authentication account.

### Testing

`./gradlew test`

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -266,11 +272,23 @@ public InteractiveTransactionResponse deposit(Sep10Jwt token, Map<String, String
...
+              clientConfig.getName());
+          throw new SepValidationException("Provided 'account' is not allowed");
+        }
+      } else {
+        infoF(
+            "The request account:{} does not match the one in the token:{}",
+            destinationAccount,
+            token.getAccount());
+        throw new SepValidationException(ERR_TOKEN_ACCOUNT_MISMATCH);
+      }
```
JakeUrban, line 291: If I'm understanding this right, this logic will throw an error if a custodial wallet passes an account that isn't in the destination accounts list, even if that list is empty, is that correct?

My concern is that we're not supporting the case where the anchor doesn't care what the destination account is as long as the authenticated account is a trusted custodial wallet. I think we need to support this case.

core/src/main/java/org/stellar/anchor/util/ConfigHelper.java
```diff
@@ -0,0 +1,29 @@
...
+    if (isEmpty(clientDomain)) {
+      clientConfig = clientsConfig.getClientConfigBySigningKey(sep10Account);
+      if (clientConfig != null && clientConfig.getType() == ClientsConfig.ClientType.NONCUSTODIAL) {
+        throw new SepValidationException("Non-custodial clients must specify a client_domain");
+      }
+    } else {
+      clientConfig = clientsConfig.getClientConfigByDomain(sep10Account);
+    }
+    return clientConfig;
+  }
```
JakeUrban, line 23: I know we're just moving this function, but its a bit confusing to me. It seems like we're assuming the account requesting authentication (`sep10Account`) is the same as the `SIGNING_KEY` on a wallet's stellar.toml file. Is that correct? 

@philipliu can you explain how this works?

JakeUrban: Actually I think I understand, we decided to call accounts that custodial wallets use to authenticate "signing keys".

Ifropc: ^ I think so

philipliu: Yeah, that's correct. I think this method could benefit from some Javadoc. Do you mind adding it?

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
```diff
@@ -98,7 +100,7 @@ private void validateClient(ClientConfig clientConfig, Errors errors) {
     }
   }
 
-  void validateCustodialClient(ClientConfig clientConfig, Errors errors) {
+  public void validateCustodialClient(ClientConfig clientConfig, Errors errors) {
```
philipliu, line 103: WDYT about validating all `destination_account`s are valid Stellar accounts? I don't think Anchors should discover this when a client makes a deposit.

philipliu: Agreed. I was thinking more along the lines of validating the format of the account ID using `KeyPair.fromAccountId`.

Ifropc: If it's invalid account it wouldn't matter, as it's not going to pass account matching

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
JakeUrban: I think if the list is empty, we can assume the destination account is permitted. I don't think that complicates the implementation much

Ifropc: You mean any destination account is permitted?

Ifropc: It is there by default. They just need to omit list in config (it's optional)
So if they don't pass list in the config, it works as before (only auth account is allowed)

JakeUrban: > So if they don't pass list in the config, it works as before (only auth account is allowed)

But I'm saying thats not what we want. We should allow any destination account if the list is omitted (or empty?) and the authenticated account belongs to a custodial wallet.

Ifropc: But is there a use case for that? It's very easy to add this functionality, but I'm curious why exactly we would want to do it. 
I'm concerned that adding empty list by mistake would allow arbitrary account to make deposits (bypassing their KYC policy, etc)

JakeUrban: Well I think its a matter of the on/off ramp's security policy and whether or not they can realistically keep these allowlists accurately over time. Its also something the custodial wallets probably would prefer.

If a custodial wallet (or exchange) updates their hot wallets frequently, or has a large number of them, I think its unreasonable to assume they will notify their on/off ramp partners of changes or that those on/off ramps will be able to update their list in a timely fashion.

Or lets say a hot wallet is compromised (or they suspect it is) so they transfer funds to another address. If their on/off ramp partners require that new address to be added to their allowlist, the wallet's on/off ramp services are effectively down.

Ifropc: Let's just add a flag "allowAll" instead of using empty list for this.
So in config, they either:
1. Don't do anything (allowAll = false, list = null) -> old behavior  (only auth address is allowed) 
2. Set allowAll to true (and list will be just ignored) -> allow deposit from all sources
3. Set list to any value -> only addresses from the list are allowed
WDYT?

JakeUrban: Yea I think thats good too. What would the schema look like?

Ifropc: The config schema? 
```yaml
     - name:
       type: custodial
       signing_key:
       callback_url:
       allow_any_destination:
       destination_accounts:
```
Something like this

JakeUrban: Ok -- that seems ok. I think the default should be `true` but the important piece is that the feature is available.

Last thing, how do you specify the destination accounts list in environment variables? Are these equivalent?

```
destination_accounts:
  - G1
  - G2
```

```
destination_accounts: ["G1", "G2"]
```

```
DESTINATION_ACCOUNTS=G1,G2
```

Ifropc: They are not. It will be 
```
clients[1].destination_accounts[0]=G1
clients[1].destination_accounts[0]=G2
```
I don't mind to make config to be string and parse it later
So it will be
```yaml
destination_accounts = "G1, G2"
```
and env
```
DESTINATION_ACCOUNTS=G1,G2
```

JakeUrban: Any way we could allow the proper list form in the yaml file but a comma-separated list in the env vars? If not, I'm not sure which variation is better.

Ifropc: Not in the current implementation (we need to add support for this to our configuration handlers)

Ifropc: I think we should just go with lists for now. MGI is using .yaml files, and I don't think anybody else is interested in this feature at the moment.
We will re-visit client and list configurations in v3

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -184,6 +184,8 @@ event_processor:
 #                  is the same as the `signing_key`, the callback will be activated.
 #   - callback_url: (optional) the URL of the client's callback API endpoint. Optional due to some wallets may opt to
 #                  poll instead, or may use polling first before implementing callbacks at a later stage.
+#   - destination_accounts: (optional) list of accounts allowed to be used for the deposit. By default, only SEP-10
+#                  authenticated account can be used to deposit funds into.
```
JakeUrban, line 188: I think we're missing an entry for `allow_any_destination` here right? Its in the example below

JakeUrban (approved):
I think we need one more adjustment but otherwise LGTM

PR #1126, opened 2023-09-26
[ANCHOR-464] SEP-6: Enable client status callback

philipliu:
### Description

This enables client status callback for SEP-6.

### Context

Status callback should be supported by the SEP-6 implement.

### Testing

`./gradlew test`

### Known limitations

The SEP-24 end-to-end was previously not asserting on the wallet status callbacks. This was caused by the signature verification failing in the wallet reference server, causing the GET `/callback` to always return null. The test previously only asserted when the result was _not_ null. These tests have been disabled in this PR due to flakiness but will be re-enabled again when `develop` is merged in.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
@@ -2,6 +2,7 @@ package org.stellar.anchor.sep6;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.stellar.anchor.api.sep.sep6.Sep6TransactionResponse;
 import org.stellar.anchor.api.shared.RefundPayment;
 import org.stellar.anchor.api.shared.Refunds;
 
@@ -13,7 +14,7 @@ public class Sep6TransactionUtils {
    * @param txn the SEP-6 database transaction object
    * @return the SEP-6 API transaction object
    */
-  public static org.stellar.anchor.api.sep.sep6.Sep6Transaction fromTxn(Sep6Transaction txn) {
+  public static Sep6TransactionResponse fromTxn(Sep6Transaction txn) {
     Refunds refunds = null;
     if (txn.getRefunds() != null && txn.getRefunds().getPayments() != null) {
       List<RefundPayment> payments = new ArrayList<>();
@@ -33,8 +34,8 @@ public class Sep6TransactionUtils {
               .payments(payments.toArray(new RefundPayment[0]))
               .build();
     }
-    org.stellar.anchor.api.sep.sep6.Sep6Transaction.Sep6TransactionBuilder builder =
-        org.stellar.anchor.api.sep.sep6.Sep6Transaction.builder()
+    Sep6TransactionResponse.Sep6TransactionResponseBuilder builder =
+        Sep6TransactionResponse.builder()
             .id(txn.getId())
             .kind(txn.getKind())
             .status(txn.getStatus())
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
@@ -1,0 +1,74 @@
+package org.stellar.anchor.sep6;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.stellar.anchor.api.shared.RefundPayment;
+import org.stellar.anchor.api.shared.Refunds;
+
+public class Sep6TransactionUtils {
+
+  /**
+   * Converts a SEP-6 database transaction object to a SEP-6 API transaction object.
+   *
+   * @param txn the SEP-6 database transaction object
+   * @return the SEP-6 API transaction object
+   */
+  public static org.stellar.anchor.api.sep.sep6.Sep6Transaction fromTxn(Sep6Transaction txn) {
+    Refunds refunds = null;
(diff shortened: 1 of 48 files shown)


core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class Sep6TransactionUtils {
+
+  /**
+   * Converts a SEP-6 database transaction object to a SEP-6 API transaction object.
+   *
+   * @param txn the SEP-6 database transaction object
+   * @return the SEP-6 API transaction object
+   */
+  public static org.stellar.anchor.api.sep.sep6.Sep6Transaction fromTxn(Sep6Transaction txn) {
```
lijamie98, line 16:
```suggestion
  public static Sep6Transaction fromTxn(Sep6Transaction txn) {
```

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
```diff
@@ -100,27 +99,25 @@ class Sep24End2EndTest(config: TestConfig, val jwt: String) {
...
-        val expectedCallbacks: List<Sep24GetTransactionResponse> =
-          gson.fromJson(
-            expectedDepositCallbacksJson,
-            object : TypeToken<List<Sep24GetTransactionResponse>>() {}.type
-          )
-        compareAndAssertCallbacks(asset, expectedCallbacks, actualCallbacks)
-      }
+      val actualCallbacks =
+        walletServerClient.pollCallbacks(response.id, 4).map {
+          gson.fromJson(it, Sep24GetTransactionResponse::class.java)
```
lijamie98, line 113: Can you add some comments of why removing `assertEquals(4, it.size)` would fix the problem?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
```diff
@@ -11,7 +11,6 @@ import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.delay
```
lijamie98, line 11: 💡 Can we separate the fix to a different PR?

wallet-reference-server/src/main/kotlin/org/stellar/reference/wallet/WalletServerClient.kt
```diff
@@ -42,14 +43,25 @@ class WalletServerClient(val endpoint: Url = Url("http://localhost:8092")) {
...
-    )
+    return gson.fromJson(response.body<String>(), object : TypeToken<List<String>>() {}.type)
   }
 
-  suspend fun getLatestCallback(): Sep24GetTransactionResponse? {
+  suspend fun pollCallbacks(txnId: String?, expected: Int): List<String> {
+    var retries = 5
+    var callbacks: List<String> = listOf()
+    while (retries > 0) {
+      // TODO: remove when callbacks are de-duped
```
lijamie98, line 53: Where and when the callbacks should be de-duped?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
philipliu: It doesn't fix the problem. The assertion was required before because `waitForWalletServerCallbacks` returned a null if the response size is not 4. I changed the API of `waitForWalletServerCallbacks` / `pollCallbacks` so that it will always return a list, so the assertion can be removed.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
philipliu: I needed to add tests for SEP-6 client status callbacks which ended up requiring touching the SEP-24 tests.

wallet-reference-server/src/main/kotlin/org/stellar/reference/wallet/WalletServerClient.kt
philipliu: I'm not sure.... Apparently this was fixed in the `develop` branch. When I pull that in, I'll remove this TODO and see if everything works.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep24End2EndTests.kt
philipliu: These changes will be thrown away once I merge in `develop`, anyway. I've only modified these tests so the CI passes.

core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
philipliu: Unfortunately, we have two classes with the same name.

philipliu: Ah, it's because SEP-24/31 call the platform transaction class something else. https://github.com/stellar/java-stellar-anchor-sdk/blob/625e469bb7c55ffa77d7cf0906b1b926da967fa0/core/src/main/java/org/stellar/anchor/sep24/Sep24Helper.java#L81.

I will rename the class too.

lijamie98 (approved):
LGTM

PR #1130, opened 2023-09-27
[Anchor-418] Support auth with SEP10 in SEP38

JiahuiWho:
### Description

Add sep38.requires_sep10 configuration with default value to FALSE

### Context

Currently, according to the SEP-38 spec, some of the SEP38 endpoints have SEP-10 as optional. This task seeks ensure we support this optionality by add configurability to these endpoints' SEP-10 requirements.

### Testing
Tests were added to verify url pattern was added to filter if auth is required

### DIFF
--- a/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep38Tests.kt
+++ b/integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep38Tests.kt
@@ -5,7 +5,6 @@ import java.time.format.DateTimeFormatter
 import kotlin.test.assertEquals
 import org.junit.jupiter.api.assertThrows
 import org.stellar.anchor.api.exception.SepException
-import org.stellar.anchor.api.exception.SepNotAuthorizedException
 import org.stellar.anchor.api.sep.sep38.Sep38Context.SEP31
 import org.stellar.anchor.platform.Sep38Client
 import org.stellar.anchor.platform.TestConfig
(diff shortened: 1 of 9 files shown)


integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep38Tests.kt
```diff
@@ -84,9 +86,22 @@ class Sep38Tests(config: TestConfig, toml: TomlContent, jwt: String) {
...
+      "stellar:USDC:GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP",
+      SEP31
+    )
+  }
+
   fun testAll() {
     println("Performing SEP38 tests...")
     `test sep38 info, price and prices endpoints`()
     `test selling over asset limit throws an exception`()
+    `test endpoints does not required valid token when auth is disabled`()
```
philipliu, line 105: Can we also add tests for when authentication is enabled?

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -451,6 +451,8 @@ sep38:
   # Whether to enable SEP-38
   #
   enabled: false
+  # Whether to require SEP-10 authentication for SEP-38 requests.
+  requires_sep10: false
```
philipliu, line 455: SEP-38 has other endpoints that require authentication such as `POST /quote`. Should we change the name of this flag or schema (nesting it under `sep38.endpoints.info.authentication_required`, for example) so that this is clear?

lijamie98, line 455: I feel `requires_sep10` is vague because some of the endpoints already requires sep10.

How about:
```suggestion
  enforce_sep10: false
```

`sep38.endpoints.info.authentication_required` seems a bit over engineering.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep38Tests.kt
```diff
@@ -14,9 +14,11 @@ import org.stellar.anchor.util.Sep1Helper.TomlContent
 
 class Sep38Tests(config: TestConfig, toml: TomlContent, jwt: String) {
   private val sep38Client: Sep38Client
+  private val sep38ClientWithoutJwt: Sep38Client
 
   init {
     sep38Client = Sep38Client(toml.getString("ANCHOR_QUOTE_SERVER"), jwt)
+    sep38ClientWithoutJwt = Sep38Client(toml.getString("ANCHOR_QUOTE_SERVER"), "Invalid Token")
```
lijamie98, line 21: This is only used in `test endpoints does not required valid token when auth is disabled()`. It's better to use local var.

lijamie98 (approved):
Please feel free to merge after the comments are resolved.

PR #1139, opened 2023-10-03
SEP-6: Remove to account set by withdraw

philipliu:
### Description

Removes the `to` account set by the withdraw endpoint.

### Context

`to` is an optional field in the transactions object, and is meant to store the user's off-chain account. This is not known at the time of transaction creation.

### Testing

- `./gradlew test`

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionStore.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionStore.java
@@ -27,5 +27,6 @@ public interface Sep6TransactionStore {
 
   List<? extends Sep6Transaction> findTransactions(TransactionsParams params) throws SepException;
 
-  Sep6Transaction findOneByToAccountAndMemoAndStatus(String toAccount, String memo, String status);
+  Sep6Transaction findOneByWithdrawAnchorAccountAndMemoAndStatus(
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionStore.java
```diff
@@ -27,5 +27,6 @@ List<Sep6Transaction> findTransactions(
 
   List<? extends Sep6Transaction> findTransactions(TransactionsParams params) throws SepException;
 
-  Sep6Transaction findOneByToAccountAndMemoAndStatus(String toAccount, String memo, String status);
+  Sep6Transaction findOneByWithdrawAnchorAccountAndMemoAndStatus(
+      String toAccount, String memo, String status);
```
JiahuiWho, line 31: `String withdrawAnchorAccount` ?

JiahuiWho (approved):
LGTM

core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionStore.java
philipliu: Oops, thanks!

PR #1141, opened 2023-10-05
[Fix] SEP-6: Only patch transaction if request field is not null

philipliu:
### Description

SEP-6 fields are patched without using the `updateField` method, so it sets any fields that to null that aren't specified in the patch transaction request.

### Context

This is how fields are patched in SEP-24 and SEP-31.

### Testing

- `./gradlew test`

### Known limitations

The class is missing unit tests, but PATCH transactions will be deprecated soon in favor of RPC API, anyway.


JiahuiWho (approved):
LGTM

PR #1144, opened 2023-10-09
[ANCHOR-498] Restructure Sep10Service functions for extensibility and customizability

lijamie98:
### Description

- Restructure Sep1 service functions for extensibility and customizability.

### Context

Core SDK release.

### Testing

- `./gradlew test`

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep10/ISep10Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep10/ISep10Service.java
@@ -9,14 +9,30 @@ import org.stellar.anchor.api.sep.sep10.ValidationRequest;
 import org.stellar.sdk.Memo;
 
 public interface ISep10Service {
+
   /**
-   * Validate the challenge transaction. The default implementation is NOOP. This is used for the
+   * Validate the challenge request. The default implementation is NOOP. This is used for the
    * customization of the validation behavior.
    *
    * @param request The challenge request.
-   * @throws SepException If the challenge transaction is invalid.
    */
-  void postChallengeRequestValidation(ChallengeRequest request) throws SepException;
+  void preChallengeRequestValidation(ChallengeRequest request) throws SepException;
+
+  /**
+   * Validate the home domain of the challenge request.
+   *
+   * @param request The challenge request.
+   * @throws SepException If the home domain is invalid.
+   */
+  void validateChallengeRequestFormat(ChallengeRequest request) throws SepException;
+
+  /**
+   * Validate the client domain of the challenge request.
+   *
+   * @param request The challenge request.
+   * @throws SepException If the client domain is invalid.
+   */
+  void validateChallengeRequestClient(ChallengeRequest request) throws SepException;
 
   /**
    * Validate the memo of the challenge request and return the memo that is validated.
@@ -28,28 +44,13 @@ public interface ISep10Service {
   Memo validateChallengeRequestMemo(ChallengeRequest request) throws SepException;
 
   /**
-   * Validate the client domain of the challenge request.
-   *
-   * @param request The challenge request.
-   * @throws SepException If the client domain is invalid.
-   */
-  void validateChallengeRequestClient(ChallengeRequest request) throws SepException;
-
-  /**
-   * Validate the home domain of the challenge request.
-   *
-   * @param request The challenge request.
-   * @throws SepException If the home domain is invalid.
-   */
-  void validateChallengeRequestFormat(ChallengeRequest request) throws SepException;
-
-  /**
-   * Validate the challenge request. The default implementation is NOOP. This is used for the
+   * Validate the challenge transaction. The default implementation is NOOP. This is used for the
    * customization of the validation behavior.
    *
    * @param request The challenge request.
+   * @throws SepException If the challenge transaction is invalid.
    */
-  void preChallengeRequestValidation(ChallengeRequest request) throws SepException;
+  void postChallengeRequestValidation(ChallengeRequest request) throws SepException;
 
   /** Increment the metrics counter for the number of challenge requests created. */
   void incrementChallengeRequestCreatedCounter();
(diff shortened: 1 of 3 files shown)


JiahuiWho (approved):
LGTM

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -58,167 +57,248 @@ public Sep10Service(
...
+      warnEx(ex);
+      throw new SepException("Failed to create the sep-10 challenge.", ex);
+    }
+  }
+
+  Transaction newChallenge(ChallengeRequest request, String clientSigningKey, Memo memo)
+      throws InvalidSep10ChallengeException {
+
+    KeyPair signer = KeyPair.fromSecretSeed(secretConfig.getSep10SigningSeed());
+    long now = System.currentTimeMillis() / 1000L;
```
philipliu, line 133: We should probably avoid calling `System#currentTimeMillis` and `Instant#now` and replace them with a `Clock` call, so that we can test time-related code.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -58,167 +57,248 @@ public Sep10Service(
...
-                sep10Config.getWebAuthDomain());
+  String fetchSigningKeyFromClientDomain(String clientDomain) throws SepException {
+    return Sep10Helper.fetchSigningKeyFromClientDomain(clientDomain);
+  }
 
-    debugF(
-        "Challenge parsed. account={}, home_domain={}",
-        shorter(challenge.getClientAccountId()),
-        challenge.getMatchedHomeDomain());
+  void validateHomeDomain(ChallengeRequest request) throws SepValidationException {
```
philipliu, line 236: What do you think about moving some of the validations into their own class? We do something like this in [SEP-6](https://github.com/stellar/java-stellar-anchor-sdk/blob/sep-6/core/src/main/java/org/stellar/anchor/sep6/RequestValidator.java). I found that it made the `Sep6Service` easier to test.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -253,51 +333,60 @@ public String validateChallenge(String challengeXdr)
 
       // increment counter
       sep10ChallengeValidatedCounter.increment();
```
philipliu, line 335: Should the counter be incremented here?

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -58,167 +57,248 @@ public Sep10Service(
...
-      challengeRequest.setHomeDomain(sep10Config.getHomeDomain());
-    } else if (!challengeRequest.getHomeDomain().equalsIgnoreCase(sep10Config.getHomeDomain())) {
-      infoF("Bad home_domain: {}", challengeRequest.getHomeDomain());
-      throw new SepValidationException(
-          String.format("home_domain [%s] is not supported.", challengeRequest.getHomeDomain()));
+  public ChallengeResponse createChallenge(ChallengeRequest request) throws SepException {
+    info("Creating SEP-10 challenge.");
+
+    // pre validation to be defined by the anchor
+    preChallengeRequestValidation(request);
```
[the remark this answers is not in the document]
lijamie98: All the other functions has comments. I would make this consistent unless you felt really annoyed. :-)

core/src/main/java/org/stellar/anchor/sep10/ISep10Service.java
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @param request The validation request.
+   * @param challenge The parsed challenge transaction.
+   * @throws SepException If the validation request is invalid.
+   */
+  void preValidateRequestValidation(ValidationRequest request, ChallengeTransaction challenge)
+      throws SepException;
+
+  /** Increment the metrics counter for the number of validation requests validated. */
+  void incrementValidationRequestValidatedCounter();
```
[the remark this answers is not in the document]
lijamie98: This would make Sep10Service abstract and we will need to have another class in the `platform` to implement this function. I am trying to find the balance here.

The `Metrics` class is done by micrometer ana they did a good job in accommodating different metrics frameworks.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -58,167 +57,248 @@ public Sep10Service(
...
+      Transaction txn = newChallenge(request, clientSigningKey, memo);
+
+      // Convert the challenge to response
+      trace("SEP-10 challenge txn:", txn);
+      ChallengeResponse challengeResponse =
+          ChallengeResponse.of(txn.toEnvelopeXdrBase64(), appConfig.getStellarNetworkPassphrase());
+      trace("challengeResponse:", challengeResponse);
+      return challengeResponse;
+    } catch (InvalidSep10ChallengeException ex) {
+      warnEx(ex);
```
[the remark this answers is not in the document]
lijamie98: Yeah, this triggers warning logs.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
lijamie98: Replace with `Instant.now().getEpochSecond()`.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
lijamie98: Defiantly. Do you mind to extract this method in a SEP-6 PR? This way, the use case will be more solid.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
lijamie98: This is actually a bug. We should returned inside the try block and the end of the function.

PR #1146, opened 2023-10-10
[Fix] SEP-6: Verify KYC of authenticated account

philipliu:
### Description

SEP-6 will now check the KYC information of the SEP-10 account instead of the requested account.

### Context

The requested account is not the same as the SEP-10 account for custodial wallets.

### Testing

- `./gradlew test`

### Known limitations

N/A


lijamie98 (approved):
LGTM

JakeUrban (approved):
I would much rather always return a success response with a transaction ID and have the anchor provide the required fields asynchronously, but if we're going with this approach this update LGTM

PR #1136, opened 2023-10-02
[ANCHOR-353] SEP-6: Verify SEP-12 status before deposit/withdraw proceeds

philipliu:
### Description

This adds a check to verify that a customer is SEP-12 accepted as part of the deposit/withdraw request validation. This also removes the `VERIFICATION_REQUIRED` status as it is not a valid SEP-12 customer info status.

### Context

This was caught by a `stellar-anchor-tests` test case.

### Testing

- `./gradlew test`
- `stellar-anchor-tests`

### Known limitations

N/A

PR #1158, opened 2023-10-12
[Fix] SEP-6: Set customers field in platform transaction

philipliu:
### Description

This sets the `customers`(See `TransactionSEP31` [docs](https://developers.stellar.org/api/anchor-platform/resources/get-transaction)) field in the SEP-6 platform transaction.

### Context

When the business server receives a transaction event, it needs to verify if the customer has enough KYC information to proceed with the transaction. Currently, the platform transaction only has `source_account` and `destination_account` fields which is not enough to identify users using a custodial wallet.

### Testing

- `./gradlew test`

### Known limitations

This persists SEP-12 customer IDs in the SEP-6 transaction table. SEP-12 customers can be deleted at any time, therefore an anchor may lose the mapping from transactions to these customers. This should be fine though since transactions would likely not move back into `pending_customer_info_update` after accepting user funds and can be completed.

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/RequestValidator.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/RequestValidator.java
@@ -116,13 +116,15 @@ public class RequestValidator {
   }
 
   /**
-   * Validates that the authenticated account has been KYC'ed by the anchor.
+   * Validates that the authenticated account has been KYC'ed by the anchor and returns its SEP-12
+   * customer ID.
    *
    * @param sep10Account the authenticated account
    * @param sep10AccountMemo the authenticated account memo
    * @throws AnchorException if the account has not been KYC'ed
+   * @return the SEP-12 customer ID if the account has been KYC'ed
    */
-  public void validateKyc(String sep10Account, String sep10AccountMemo) throws AnchorException {
+  public String validateKyc(String sep10Account, String sep10AccountMemo) throws AnchorException {
     GetCustomerRequest request =
         sep10AccountMemo != null
             ? GetCustomerRequest.builder()
@@ -151,5 +153,7 @@ public class RequestValidator {
         throw new ServerErrorException(
             String.format("unknown customer status: %s", response.getStatus()));
     }
+
+    return response.getId();
   }
 }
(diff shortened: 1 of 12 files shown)


lijamie98:
Do we need[ `getDepositAsset()` ](https://github.com/stellar/java-stellar-anchor-sdk/pull/1158/files#diff-2a145c0174c66604c712cb3b9092186becced6bb2a07513983712300e16de126R29)and `getWithdrawAsset()` in the validator?

lijamie98:
When `response` isn't ACCEPTED in the [`validateKyc`](https://github.com/stellar/java-stellar-anchor-sdk/blob/98d3c35378486550f29ff25d5ed3202c4d67a934/core/src/main/java/org/stellar/anchor/sep6/RequestValidator.java#L142C5-L142C36)  , I think we need to discuss a little more when the status is `NEEDS_INFO`. An alternative is to save the transaction and return the status as `NEEDS_INFO`. We can discuss more in the final review meeting.

lijamie98 (approved):
I put some comments about handling the `validateKyc` call. That can be addressed in a different meeting and different PR. 

Otherwise, LGTM.

philipliu:
> Do we need[ getDepositAsset() ](https://github.com/stellar/java-stellar-anchor-sdk/pull/1158/files#diff-2a145c0174c66604c712cb3b9092186becced6bb2a07513983712300e16de126R29)and getWithdrawAsset() in the validator?
>
> Or should they be part of Sep6Helper? Or making them package protected?

These methods perform validation on the request asset, they don't necessarily need to return anything, but we need a reference to the `AssetInfo` if we pass the validation. This is more of an optimization to avoid making another call to `AssetService`, but it makes the interface of `RequestValidator` kind of ugly. I'm not fully settled on it yet, but I think we work to clean it up the implementation we launch.

philipliu:
> When response isn't ACCEPTED in the [validateKyc](https://github.com/stellar/java-stellar-anchor-sdk/blob/98d3c35378486550f29ff25d5ed3202c4d67a934/core/src/main/java/org/stellar/anchor/sep6/RequestValidator.java#L142C5-L142C36) , I think we need to discuss a little more when the status is NEEDS_INFO. An alternative is to save the transaction and return the status as NEEDS_INFO. We can discuss more in the final review meeting.

I think the decision ultimately comes down to whether we want Anchors to perform KYC at all before initiating SEP-6 transactions. This validation was added in a previous PR to satisfy the `stellar-anchor-tests` https://github.com/stellar/java-stellar-anchor-sdk/pull/1136. If want to go with your suggested approach (Jake also suggested this), we need to remove those KYC tests.

PR #1159, opened 2023-10-12
[ANCHOR-462] Enable platform and reference server tests concurrency

lijamie98:
### Description

Enable platform and reference server tests concurrency

### Context

Optimize the unit tests.

### Testing

- `./gradlew test`

### DIFF
--- a/platform/build.gradle.kts
+++ b/platform/build.gradle.kts
@@ -57,7 +57,28 @@ dependencies {
 
 tasks.test {
   // Enable parallel test execution
-  systemProperty("junit.jupiter.execution.parallel.enabled", false)
+  // Enable parallel test execution
+  systemProperty("junit.jupiter.execution.parallel.enabled", true)
+  // Enable parallel test execution
+  systemProperty("junit.jupiter.testinstance.lifecycle.default", "per_method")
+  // Allocate thread count based on available processors
+  systemProperty("junit.jupiter.execution.parallel.config.strategy", "dynamic")
+  // Set default parallel mode to same thread. All tests within a class are run in sequence.
+  systemProperty("junit.jupiter.execution.parallel.mode.default", "concurrent")
+  // Set default parallel mode for classes to concurrent. All test classes are run in parallel.
+  systemProperty("junit.jupiter.execution.parallel.mode.classes.default", "concurrent")
+
+  // Set default test class order to order annotation. All test classes are run in parallel.
+  // Some tests take longer to run. Enabling the order will execute long-running tests first to
+  // shorten the overall test time.
+  systemProperty(
+    "junit.jupiter.testclass.order.default",
+    "org.junit.jupiter.api.ClassOrderer\$OrderAnnotation"
+  )
+  maxParallelForks =
+    (Runtime.getRuntime().availableProcessors() / 2).coerceAtLeast(1).also {
+      println("junit5 ... setting maxParallelForks to $it")
+    }
   testLogging {
     exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
     events = setOf(org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED)
--- a/platform/src/main/java/org/stellar/anchor/metrics/MetricsService.java
+++ b/platform/src/main/java/org/stellar/anchor/metrics/MetricsService.java
@@ -1,0 +1,10 @@
+package org.stellar.anchor.metrics;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
+
+public class MetricsService {
+  public Counter counter(String name, String... tags) {
+    return Metrics.counter(name, tags);
+  }
+}
--- a/platform/src/main/java/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentService.java
@@ -12,6 +12,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import lombok.Getter;
+import lombok.Setter;
 import org.springframework.retry.annotation.Backoff;
 import org.springframework.retry.annotation.Retryable;
 import org.stellar.anchor.api.custody.CreateTransactionPaymentResponse;
@@ -30,6 +32,8 @@ import org.stellar.anchor.util.GsonUtils;
 import org.stellar.sdk.xdr.MemoType;
 
 /** Fireblocks implementation of payment service */
+@Setter
+@Getter
 public class FireblocksPaymentService implements CustodyPaymentService<TransactionDetails> {
 
   private static final Gson gson = GsonUtils.getInstance();
@@ -46,11 +50,12 @@ public class FireblocksPaymentService implements CustodyPaymentService<Transacti
   private static final String QUERY_PARAM_SORT = "sort";
   private static final String TRANSACTIONS_ORDER_BY = "createdAt";
   private static final String TRANSACTIONS_SORT = "ASC";
-  public static int TRANSACTIONS_LIMIT = 500;
+  private static final int TRANSACTION_LIMIT = 500;
 
   private final FireblocksApiClient fireblocksApiClient;
   private final FireblocksConfig fireblocksConfig;
   private final Type transactionDetailsListType;
+  private int transactionLimit = TRANSACTION_LIMIT;
 
   public FireblocksPaymentService(
       FireblocksApiClient fireblocksApiClient, FireblocksConfig fireblocksConfig) {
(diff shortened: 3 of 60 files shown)


platform/src/main/java/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentService.java
```diff
@@ -46,11 +50,12 @@ public class FireblocksPaymentService implements CustodyPaymentService<Transacti
   private static final String QUERY_PARAM_SORT = "sort";
   private static final String TRANSACTIONS_ORDER_BY = "createdAt";
   private static final String TRANSACTIONS_SORT = "ASC";
-  public static int TRANSACTIONS_LIMIT = 500;
+  private static final int TRANSACTION_LIMIT = 500;
 
   private final FireblocksApiClient fireblocksApiClient;
   private final FireblocksConfig fireblocksConfig;
   private final Type transactionDetailsListType;
+  private int transactionLimit = TRANSACTION_LIMIT;
```
philipliu, line 58: I think we are doing this because we want to configure this for tests. If so, can we make `transactionLimit` a constructor parameter?

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
```diff
@@ -9,17 +9,18 @@ import kotlin.test.assertEquals
...
+import org.junit.jupiter.api.parallel.ExecutionMode
 import org.skyscreamer.jsonassert.JSONAssert
 import org.skyscreamer.jsonassert.JSONCompareMode
 import org.stellar.anchor.api.exception.FireblocksException
 import org.stellar.anchor.platform.config.FireblocksConfig
 import org.stellar.anchor.platform.data.JdbcCustodyTransaction
 import org.stellar.anchor.util.GsonUtils
 
+@Execution(ExecutionMode.CONCURRENT)
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
```
philipliu, line 23: Are these the defaults for this project? Does it need to be set here?

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
```diff
@@ -173,15 +174,15 @@ class FireblocksPaymentServiceTest {
 
   @Test
   fun `getTransactionsByTimeRange select number of items equal to limit`() {
-    FireblocksPaymentService.TRANSACTIONS_LIMIT = 2
+    fireblocksPaymentService.transactionLimit = 2
```
philipliu, line 177: Instead of mutating the object under test? Can we initialize a new `FireblocksPaymentService` with a different `transactionLimit per test?

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
```diff
@@ -244,6 +245,8 @@ class FireblocksPaymentServiceTest {
 
     val ex =
       assertThrows<IllegalArgumentException> {
+        println("44444444444444444")
```
philipliu, line 248: Oops?

platform/build.gradle.kts
```diff
@@ -57,7 +57,28 @@ dependencies {
 
 tasks.test {
   // Enable parallel test execution
-  systemProperty("junit.jupiter.execution.parallel.enabled", false)
+  // Enable parallel test execution
+  systemProperty("junit.jupiter.execution.parallel.enabled", true)
+  // Enable parallel test execution
```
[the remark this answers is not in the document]
lijamie98: Yes, `per_method` is required to run in parallel unless we need to do major changes. I am writing a document that will address the reason.

platform/src/main/java/org/stellar/anchor/metrics/MetricsService.java
```diff
@@ -0,0 +1,10 @@
+package org.stellar.anchor.metrics;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
+
+public class MetricsService {
+  public Counter counter(String name, String... tags) {
```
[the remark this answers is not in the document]
lijamie98: It is not for anchor's metrics though... Don't feel it would benefit too much renaming to `AnchorMetrics`.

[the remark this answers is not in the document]
lijamie98: > We can avoid the explicit `toString` conversion when calling this method.

Can you elaborate a little? Where do we need to call the `toString` conversion?

platform/src/main/java/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentService.java
```diff
@@ -124,6 +129,8 @@ public TransactionDetails getTransactionById(String txnId) throws FireblocksExce
         TransactionDetails.class);
   }
 
+  int counter = 0;
```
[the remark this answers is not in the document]
lijamie98: Oops. My concurrency debug code. Removed.

platform/src/main/java/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentService.java
lijamie98: Yes, this was added to allow configuration of the tests. 

Add a constructor that accepts the `transactionLimit` as an argument.

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
lijamie98: Removed.

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
lijamie98: That would cause a lot of duplication calling the constructor though. 

We need to do: 
```
    fireblocksPaymentService = FireblocksPaymentService(fireblocksClient, fireblocksConfig, 2)
```
for every function even if we use the default `TRANSACTIONS_LIMIT`

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
lijamie98: Ouch....

Removed.

platform/src/main/java/org/stellar/anchor/metrics/MetricsService.java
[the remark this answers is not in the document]
philipliu: I'm not suggesting renaming the method, I'm asking if we could use a type besides String for the `name` parameter. 

[Here](https://github.com/stellar/java-stellar-anchor-sdk/pull/1159/files#diff-f202a2e3c43a99f71c9036808ee8c71a540fd22566211facb7f5b60c5cb95b1fR60-R61) is an example, where we call `toString` on an `AnchorMetric` argument.

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
philipliu: I think a bit of duplication in the test code is preferable to mutating the service's configuration. We should be avoiding mutation in general but especially for non-data types like `FireblocksPaymentService`.

platform/build.gradle.kts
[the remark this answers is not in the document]
philipliu: It would be great if there was also some explanation in this file.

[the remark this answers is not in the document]
lijamie98: Added

platform/src/main/java/org/stellar/anchor/metrics/MetricsService.java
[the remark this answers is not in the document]
lijamie98: Good idea. Also created [ANCHOR-512](https://stellarorg.atlassian.net/browse/ANCHOR-512) to deprecate MetricConstants.

platform/src/test/kotlin/org/stellar/anchor/platform/custody/fireblocks/FireblocksPaymentServiceTest.kt
lijamie98: I created a overloaded constructor and make the `transactionLimit` read-only.

JiahuiWho (approved):

PR #1161, opened 2023-10-13
[ANCHOR-362] SEP-6: Update reference server to set required_customer_info_updates

philipliu:
### Description

This updates the SEP-6 E2E test to use the `required_customer_info_updates` field when providing KYC updates.

### Context

Currently, this field is not set by the reference server implementation. This is needed so that the demo wallet could prompt the user to submit the missing KYC information.

### Testing

- `./gradlew test`

### Known limitations

N/A


api-schema/src/main/java/org/stellar/anchor/api/callback/PutCustomerRequest.java
```diff
@@ -113,10 +113,10 @@ public class PutCustomerRequest {
   String idCountryCode;
 
   @SerializedName("id_issue_date")
-  Instant idIssueDate;
+  String idIssueDate;
```
lijamie98, line 116: I think there is a reason that we use `Instant`. Can you explain a little the reason we want to use `String` instead?

philipliu: Oops, this is a mistake, it should be an `Instant`.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/test/Sep6End2EndTest.kt
```diff
@@ -99,8 +116,7 @@ class Sep6End2EndTest(val config: TestConfig, val jwt: String) {
       )
     assertEquals(completedDepositTxn.transaction.id, transactionByStellarId.transaction.id)
 
-    val expectedStatuses =
-      listOf("incomplete", "pending_anchor", "pending_customer_info_update", "completed")
+    val expectedStatuses = listOf("incomplete", "pending_customer_info_update", "completed")
```
lijamie98, line 119: nit: Can we use the `enum` here?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/customer/CustomerService.kt
```diff
@@ -55,11 +55,18 @@ class CustomerService(private val customerRepository: CustomerRepository) {
         customer.copy(
```
lijamie98, line 55: 💡 Is there anything like [BeanUtils ](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/BeanUtils.html#copyProperties(java.lang.Object,java.lang.Object,java.lang.Class))in Kotlin?

lijamie98: Or use `gson.toJson` and `gson.fromJson` for copying? 

I think this is for the reference and does not require critical performance but rather explanations.

lijamie98 (approved):
LGTM. 

kotlin-reference-server/src/main/kotlin/org/stellar/reference/callbacks/customer/CustomerService.kt
philipliu: I wrote it this way since I don't want `null` request fields to delete a previously set customer field. By default, `BeanUtils#copyProperties` does not ignore `null` fields, we would need to write a wrapper around it if we want this behavior. This is the same reason, I did not use `gson` here.

PR #1162, opened 2023-10-14
ANCHOR-511: Fix Missing asset in JWT

Ifropc:
### Description

- Fix bug of missing asset in JWT when no asset issuer is specified

### Context

- Instead of parsing asset again, pass asset info to interactive url constructor

### Testing

- `./gradlew test`
- Added test to reproduce this behavior

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/util/AssetHelper.java
+++ b/core/src/main/java/org/stellar/anchor/util/AssetHelper.java
@@ -51,28 +51,6 @@ public class AssetHelper {
   }
 
   /**
-   * Returns the asset id in the SEP-38 asset identification format, or null if the asset is not
-   * supported.
-   *
-   * @param assetCode The asset code
-   * @param assetIssuer The asset issuer
-   * @return The asset id in the SEP-38 asset identification format.
-   */
-  public static String getAssetId(String assetCode, String assetIssuer) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/stellar/anchor/util/AssetHelper.java
```diff
@@ -50,28 +50,6 @@ public static boolean isNativeAsset(String assetCode, String assetIssuer) {
...
 
-  /**
-   * Returns the asset id in the SEP-38 asset identification format, or null if the asset is not
-   * supported.
-   *
-   * @param assetCode The asset code
-   * @param assetIssuer The asset issuer
-   * @return The asset id in the SEP-38 asset identification format.
-   */
-  public static String getAssetId(String assetCode, String assetIssuer) {
```
lijamie98, line 61: Quick question: Is this function not used or moved? I remember it is used somewhere...

lijamie98 (approved):
LGTM.

core/src/main/java/org/stellar/anchor/util/AssetHelper.java
Ifropc: It was used in removed code

PR #1163, opened 2023-10-17
[ANCHOR-444] SEP-6: Set `client_id` in fee request

philipliu:
### Description

This sets the `client_id` in the `GET /fee` integration request. It also fixes a bug where the `senderId` and `receiverId`s were incorrectly set in the same request.

### Context

Client attribution should work for SEP-6.

### Testing

- `./gradlew test`

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep6/ExchangeAmountsCalculator.java
```diff
@@ -124,4 +132,42 @@ public static class Amounts {
     String amountFee;
     String amountFeeAsset;
   }
+
+  private String getClientId(Sep10Jwt token) throws BadRequestException {
```
lijamie98, line 136: 💡 This function should not belong to `ExchangeAmountsCalculator` class. Can we move it somewhere else?

core/src/main/java/org/stellar/anchor/sep6/ExchangeAmountsCalculator.java
```diff
@@ -124,4 +132,42 @@ public static class Amounts {
...
+        throw new BadRequestException("Client name not allowed");
+      }
+      throw new BadRequestException("Client domain not allowed");
+    }
+
+    if (sep10Config.getAllowedClientNames().contains(client.getName())
+        || sep10Config.getAllowedClientNames().isEmpty()) {
+      return client.getName();
+    }
+    throw new BadRequestException("Client name not allowed");
```
lijamie98, line 171: The logic seems complicated and not documented. Can we add some reference?

philipliu: Sure. I'll add a Javadoc and annotate the sections.

core/src/main/java/org/stellar/anchor/sep6/ExchangeAmountsCalculator.java
philipliu: Yeah, I'll move it into a separate class since it will be used by SEP-24 as well.

core/src/main/java/org/stellar/anchor/sep6/ExchangeAmountsCalculator.java
reecexlm: would it make sense to actually break this into smaller methods? would be easeier to read and be more testable?

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
```diff
@@ -0,0 +1,73 @@
...
+
+    // Check if the client domain and name are allowed
+    if (client == null) {
+      throw new BadRequestException("Client not found");
+    }
+
+    // Token has a client domain set
+    if (token.getClientDomain() != null) {
+      if (isDomainAllowed(client.getDomain()) && isClientNameAllowed(client.getName())) {
+        return client.getName();
```
lijamie98, line 43: nit: Can we move the return to the end of the function and throw exception with failed validation in the middle of the function?

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
```diff
@@ -0,0 +1,73 @@
...
+
+    // Token has only an account set
+    if (isClientNameAllowed(client.getName())) {
+      return client.getName();
+    }
+    throw new BadRequestException("Client name not allowed");
+  }
+
+  @Nullable
+  private ClientsConfig.ClientConfig getClient(Sep10Jwt token) {
```
lijamie98, line 56: nit: static import?

lijamie98 (approved):
LGTM.

Please feel free to merge after taking care of the comments.

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
philipliu: Nice!

PR #1166, opened 2023-10-17
[ANCHOR-388] SEP-6: Implement RPC actions

philipliu:
### Description

This updates the RPC actions for SEP-6 and updates the reference server implementation to use them. This PR is quite large as it updates the unit tests for most of the RPC actions but, implementation-wise, not much has changed. It might helpful to start by reviewing the `Sep6End2EndTest`  to get a sense of what changed with this PR.

### Context

SEP-6 should support RPC API.

### Testing

- `./gradlew test`

### Known limitations

- Platform API integration tests do not test the new flow but are covered by the SEP-6 e2e tests.
- Refunds and custody service integration have not been implemented.

Both of these will be addressed by ANCHOR-508

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -90,6 +90,14 @@ public class Sep6Service {
             .type(request.getType())
             .assetCode(request.getAssetCode())
             .assetIssuer(asset.getIssuer())
+            // NB: these are purposely set to incorrect values.
+            // amount_out and amount_fee assets cannot be determined when the
+            // platform creates the transaction, but the RPC API requires
+            // these to be set during a notify_amounts_updated call.
+            .amountOut(request.getAmount())
+            .amountOutAsset(asset.getSep38AssetName())
+            .amountFee("0")
+            .amountFeeAsset(asset.getSep38AssetName())
             .amountExpected(request.getAmount())
             .startedAt(Instant.now())
             .sep10Account(token.getAccount())
@@ -244,6 +252,16 @@ public class Sep6Service {
             .type(request.getType())
             .assetCode(request.getAssetCode())
             .assetIssuer(asset.getIssuer())
+            .amountIn(request.getAmount())
+            .amountInAsset(asset.getSep38AssetName())
+            // NB: these are purposely set to incorrect values.
+            // amount_out and amount_fee assets cannot be determined when the
+            // platform creates the transaction, but the RPC API requires
+            // these to be set during a notify_amounts_updated call.
+            .amountOut(request.getAmount())
+            .amountOutAsset(asset.getSep38AssetName())
+            .amountFee("0")
+            .amountFeeAsset(asset.getSep38AssetName())
             .amountExpected(request.getAmount())
             .startedAt(Instant.now())
             .sep10Account(token.getAccount())
--- a/platform/src/main/java/org/stellar/anchor/platform/component/platform/RpcActionBeans.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/component/platform/RpcActionBeans.java
@@ -39,6 +39,7 @@ import org.stellar.anchor.platform.validator.RequestValidator;
 import org.stellar.anchor.sep24.Sep24DepositInfoGenerator;
 import org.stellar.anchor.sep24.Sep24TransactionStore;
 import org.stellar.anchor.sep31.Sep31TransactionStore;
+import org.stellar.anchor.sep6.Sep6TransactionStore;
 
 @Configuration
 public class RpcActionBeans {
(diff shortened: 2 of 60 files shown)


lijamie98:
Can we add class comments of the RPC handler classes? Or some reference link to the RPC definition document?

lijamie98 (approved):
LGTM.

lijamie98:
At this moment, I am not familiar with the RPC enough to make too much meaningful reviews. But the PR looks fine with me.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -90,6 +90,14 @@ public StartDepositResponse deposit(Sep10Jwt token, StartDepositRequest request)
             .type(request.getType())
             .assetCode(request.getAssetCode())
             .assetIssuer(asset.getIssuer())
+            // NB: these are purposely set to incorrect values.
```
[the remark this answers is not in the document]
philipliu: It's an abbreviation for "nota bene", which basically means "note".

philipliu:
> Can we add class comments of the RPC handler classes? Or some reference link to the RPC definition document?

Sure, I can link to the docs, but I'll do this outside of this PR.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
Ifropc, line 93: Does it mean there will be a state of the transaction in DB when it has incorrect amounts?

platform/src/main/java/org/stellar/anchor/platform/component/platform/RpcActionBeans.java
```diff
@@ -165,18 +200,26 @@ NotifyOffchainFundsReceivedHandler notifyOffchainFundsReceivedHandler(
 
   @Bean
   NotifyOffchainFundsSentHandler notifyOffchainFundsSentHandler(
+      Sep6TransactionStore txn6Store,
```
Ifropc, line 203: We def should group this services together to simplify all factory methods for RPC handlers

platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyOffchainFundsReceivedHandler.java
```diff
@@ -157,9 +177,12 @@ protected void updateTransactionWithRpcRequest(
       txn.setAmountFee(request.getAmountFee().getAmount());
     }
 
-    JdbcSep24Transaction txn24 = (JdbcSep24Transaction) txn;
-    if (custodyConfig.isCustodyIntegrationEnabled()) {
-      custodyService.createTransaction(txn24);
+    // TODO: add support for SEP-6
```
Ifropc, line 180: Is it going to be resolved in another PR?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: Yeah, the amounts are technically incorrect anyway since the Anchor updates fees asynchronously. Deposit instructions and withdraw anchor accounts (to be done in the next PR) will not be provided until the business server makes a request for funds RPC so there is no way for the user to accidentally submit the wrong amount.

platform/src/main/java/org/stellar/anchor/platform/component/platform/RpcActionBeans.java
philipliu: Yeah, this was quite annoying...

platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyOffchainFundsReceivedHandler.java
philipliu: Yes, in the PR for ANCHOR-508.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
Ifropc: But if the user requests the transaction, won't we return incorrect data? Shouldn't amount just be null instead?

philipliu: Yes. the `NotifyAmountsUpdatedHandler` validation requires the assets to be set so it can't be null ([code](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyAmountsUpdatedHandler.java#L53-L67)). I guess we can remove the validation?

philipliu: I should probably think about this more, so I'll create a ticket to address this and merge it as is.

Ifropc: I'm pretty sure RPC validation should just validate only the request having amount to not be null, if we expect transaction to start with null amounts

philipliu: Yeah, that makes sense. Created ANCHOR-525 to fix it.

PR #1167, opened 2023-10-18
[ANCHOR-518] Add @LockAndMockStatic and @LockStatic annotations

lijamie98:
### Description

Add @LockAndMockStatic and @LockStatic annotations. Instead of 
```kotlin
class MyTest {
  @LockAndMockStatic([NetUtil::class]) 
  fun myTest(){
    lockAndMockStatic() {
      // my tests here
   }
  }
}
```
the test method can be annotated with `@LockAndMockStatic`

```kotlin
@ExtendWith(LockAndMockTest::class) 
class MyTest {
  @LockAndMockStatic([NetUtil::class]) 
  fun myTest(){
  }
}
```

### Context

To simplify the `lockAndMock` function.

### Testing

- `./gradlew test`

### DIFF
--- a/core/src/test/kotlin/org/stellar/anchor/auth/AuthHelperTest.kt
+++ b/core/src/test/kotlin/org/stellar/anchor/auth/AuthHelperTest.kt
@@ -22,8 +22,8 @@ class AuthHelperTest {
   @ParameterizedTest
   @EnumSource(AuthType::class)
   fun `test AuthHeader creation based on the AuthType`(authType: AuthType) {
-    val calendarSingleton = mockk<Calendar>(relaxed = true)
     lockAndMockStatic(Calendar::class) {
+      val calendarSingleton = mockk<Calendar>(relaxed = true)
       // Mock calendar to guarantee the jwt token format
       when (authType) {
         JWT -> {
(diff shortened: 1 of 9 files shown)


core/src/test/kotlin/org/stellar/anchor/auth/AuthHelperTest.kt
```diff
@@ -22,8 +22,8 @@ class AuthHelperTest {
   @ParameterizedTest
   @EnumSource(AuthType::class)
   fun `test AuthHeader creation based on the AuthType`(authType: AuthType) {
-    val calendarSingleton = mockk<Calendar>(relaxed = true)
     lockAndMockStatic(Calendar::class) {
```
philipliu, line 25: Should this be replaced by the annotation?

philipliu (approved):
LGTM

core/src/test/kotlin/org/stellar/anchor/auth/AuthHelperTest.kt
lijamie98: Probably not. I think ```val calendarSingleton = mockk<Calendar>(relaxed = true)``` should happen before the mock.

philipliu: Didn't notice that 👍

PR #1171, opened 2023-10-24
[Fix] SEP-6: Add `type` to platform transaction

philipliu:
### Description

This commit adds the `type` field to the SEP-6 platform transaction.

### Context

The `type` determines which fields need to be collected for KYC.

### Testing

- `./gradlew test`

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
@@ -30,6 +30,8 @@ public class PlatformTransactionData {
   Kind kind;
   SepTransactionStatus status;
 
+  String type;
+
   @SerializedName("amount_expected")
   Amount amountExpected;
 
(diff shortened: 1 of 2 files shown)


api-schema/src/main/java/org/stellar/anchor/api/platform/PlatformTransactionData.java
```diff
@@ -30,6 +30,8 @@ public class PlatformTransactionData {
   Kind kind;
   SepTransactionStatus status;
 
+  String type;
+
```
JakeUrban, line 34: Whats the reason for adding a type field? I thought we were moving away from SEP-12 customer types?

philipliu: This is the `withdraw` and `deposit` type, like `bank_account` or `SWIFT`.

JakeUrban (approved):

PR #1175, opened 2023-10-25
[Fix] Fix test failures due to static mock race

philipliu:
### Description

This fixes two test concurrency bugs due to static mocking.

1. `Sep10ServiceTest`:  the verify sees two calls to `Sep10Challenge#newChallenge`. My guess is that the `LockAndMockStatic` annotation does not know about `ParameterizedTest`s.
2. `Sep24ServiceTest`: the mock of `KeyPair` in `Sep10ServiceTest` to throw an exception causes the `Sep24Service` to throw it as well.

These two issues were only observed in the `sep-6` branch despite it using static mocks or touching existing tests.

### Context

Fixes test stability in `sep-6` branch.

### Testing

- Ran `./gradlew test` a few times

### Documentation

N/A

### Known limitations

This is only a temporary fix as it does not prevent us from writing similar tests in the future.

### DIFF
--- a/core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
+++ b/core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
@@ -4,8 +4,11 @@ package org.stellar.anchor.sep10
 
 import com.google.common.io.BaseEncoding
 import com.google.gson.annotations.SerializedName
-import io.mockk.*
+import io.mockk.MockKAnnotations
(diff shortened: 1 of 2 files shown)


lijamie98:
Should this PR merge to the `develop` instead of the `sep6` branch?

philipliu:
I want to get this fix into `sep-6` quicker and I plan to merge `sep-6` into `develop later this week, anyway.

lijamie98 (approved):
LGTM

PR #1177, opened 2023-10-25
[ANCHOR-515] SEP-6: Remove KYC verification

philipliu:
### Description

This removes KYC verification from the SEP-6 transaction initiation. Now that KYC is removed, the customer may not be known to the Anchor at the time of transaction initiation so the `customer` field cannot be set. To allow Anchors to associate SEP transactions with a customer, the SEP-10 account memo is now added as a field to the `StellarId` object. 

This means for the exchange endpoints, the platform can no longer make a request to the Fee integration as it requires setting SEP-12 sender and receiver fields. Business servers are expected to update the amounts asynchronously like they already do for regular fees.

### Context

KYC verification is done "interactively" in SEP-6.

### Testing

- `./gradlew test`

### Documentation

Stellar docs need to be updated so that the `StellarId` object includes an optional `memo` field

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -150,7 +150,7 @@ public class Sep6Service {
           exchangeAmountsCalculator.calculateFromQuote(
               request.getQuoteId(), sellAsset, request.getAmount());
     } else {
-      // If a quote is no provided set, amounts to 0.
+      // If a quote is not provided, set the fee and out amounts to 0.
       // The business server should use the notify_amounts_updated RPC to update the amounts.
       amounts =
           Amounts.builder()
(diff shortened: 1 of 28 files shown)


core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -144,17 +143,24 @@ public StartDepositResponse depositExchange(Sep10Jwt token, StartDepositExchange
...
-    String customerId = requestValidator.validateKyc(token.getAccount(), token.getAccountMemo());
 
-    ExchangeAmountsCalculator.Amounts amounts;
+    Amounts amounts;
     if (request.getQuoteId() != null) {
       amounts =
           exchangeAmountsCalculator.calculateFromQuote(
               request.getQuoteId(), sellAsset, request.getAmount());
     } else {
+      // If a quote is no provided set, amounts to 0.
```
JiahuiWho, line 153: Nit lol. Same below.

JiahuiWho (approved):
LGTM

Ifropc (approved):

PR #1178, opened 2023-10-26
[ANCHOR-476] add withdraw_anchor_account to PATCH /transactions

JiahuiWho:
### Description

 add `withdraw_anchor_account` to `PATCH /transactions` to support setting destination accounts for withdrawals dynamically

### Context

Today, the Anchor Platform sets `withdraw_anchor_account` to the value of the asset’s distribution account set in the config when the transaction is created.

Per partner request, support setting `withdraw_anchor_account` via `PATCH /transactions` – the address set can be different from the configured distribution account for the asset

### Testing

- `./gradlew test`
- Added test case to verify `withdraw_anchor_account` is updated via patch /transaction

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
@@ -250,17 +250,6 @@ public class TransactionService {
       case "24":
         JdbcSep24Transaction sep24Txn = (JdbcSep24Transaction) txn;
 
-        // add a memo for the transaction if the transaction is ready for user to send funds
-        if (sep24Txn.getMemo() == null
-            && Kind.WITHDRAWAL.getKind().equals(sep24Txn.getKind())
-            && sep24Txn.getStatus().equals(PENDING_USR_TRANSFER_START.toString())) {
-          SepDepositInfo sep24DepositInfo = sep24DepositInfoGenerator.generate(sep24Txn);
-          sep24Txn.setToAccount(sep24DepositInfo.getStellarAddress());
-          sep24Txn.setWithdrawAnchorAccount(sep24DepositInfo.getStellarAddress());
-          sep24Txn.setMemo(sep24DepositInfo.getMemo());
-          sep24Txn.setMemoType(sep24DepositInfo.getMemoType());
-        }
-
         if (custodyConfig.isCustodyIntegrationEnabled()
             && !lastStatus.equals(sep24Txn.getStatus())
             && ((Kind.DEPOSIT.getKind().equals(sep24Txn.getKind())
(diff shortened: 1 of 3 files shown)


platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -343,7 +332,17 @@ void updateSepTransaction(PlatformTransactionData patch, JdbcSepTransaction txn)
           txnUpdated = updateField(patch, sep24Txn, "memoType", txnUpdated);
         }
 
+        // add a memo for the transaction if the transaction is ready for user to send funds
+        if (sep24Txn.getMemo() == null
+            && Kind.WITHDRAWAL.getKind().equals(sep24Txn.getKind())
+            && sep24Txn.getStatus().equals(PENDING_USR_TRANSFER_START.toString())) {
+          SepDepositInfo sep24DepositInfo = sep24DepositInfoGenerator.generate(sep24Txn);
+          sep24Txn.setMemo(sep24DepositInfo.getMemo());
```
philipliu, line 340: I think we need to keep setting `withdraw_anchor_account` otherwise this is a breaking change.

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
```diff
@@ -250,17 +250,6 @@ private GetTransactionResponse patchTransaction(PatchTransactionRequest patch)
       case "24":
         JdbcSep24Transaction sep24Txn = (JdbcSep24Transaction) txn;
 
-        // add a memo for the transaction if the transaction is ready for user to send funds
-        if (sep24Txn.getMemo() == null
-            && Kind.WITHDRAWAL.getKind().equals(sep24Txn.getKind())
-            && sep24Txn.getStatus().equals(PENDING_USR_TRANSFER_START.toString())) {
-          SepDepositInfo sep24DepositInfo = sep24DepositInfoGenerator.generate(sep24Txn);
-          sep24Txn.setToAccount(sep24DepositInfo.getStellarAddress());
```
philipliu, line 258: `to_account` was set incorrectly before, but we should probably keep setting it as well, in case anyone was depending on it.

JiahuiWho: Hmmm can you point me to were it was set incorrectly, or is it changed anywhere? 


This `sep24DepositInfo.getStellarAddress()` is returning the same address as "sep24Txn.toAccount()"

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
JiahuiWho: Actually both `to_account` and `withdraw_anchor_account` were already set to asset's distribution account when the transaction is created ([here](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java#L216))

I don't think it make any sense to reset them to the same value here...

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
philipliu: `to` should be the off-chain account in case of `withdraw`, but here it's set to a Stellar address ([SEP link](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#fields-for-withdraw-transactions)).

That being said, I see that this is also set when the transaction is created, so please ignore my comment 😅

platform/src/main/java/org/stellar/anchor/platform/service/TransactionService.java
philipliu: Didn't see that, thanks!

philipliu (approved):
LGTM

PR #1183, opened 2023-11-01
[ANCHOR-526] Fix account not being forwarded in SEP-9's /customer callback

Ifropc:
### Description

Fix account not being forwarded in SEP-9's /customer callback

### Context

When receiving callback for SEP-9 field forwarding all JWT fields (account and memo) was missing. Turned out, they were not being forwarded properly 

### Testing

- `./gradlew test`
-  Adjusted test scenario 

### Documentation

There is a task for documenting this feature 

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
@@ -268,7 +268,8 @@ public class Sep24Service {
     InteractiveTransactionResponse response =
         new InteractiveTransactionResponse(
             "interactive_customer_info_needed",
-            interactiveUrlConstructor.construct(txn, withdrawRequest, asset, token.getHomeDomain()),
+            interactiveUrlConstructor.construct(
+                txn, withdrawRequest, asset, token.getHomeDomain(), token),
             txn.getTransactionId());
 
     // increment counter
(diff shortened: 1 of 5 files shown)


philipliu (approved):
LGTM

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -268,7 +268,8 @@ public InteractiveTransactionResponse withdraw(
     InteractiveTransactionResponse response =
         new InteractiveTransactionResponse(
             "interactive_customer_info_needed",
-            interactiveUrlConstructor.construct(txn, withdrawRequest, asset, token.getHomeDomain()),
+            interactiveUrlConstructor.construct(
+                txn, withdrawRequest, asset, token.getHomeDomain(), token),
```
lijamie98, line 272: Do we need to pass `token.getHomeDomain()` when `token` is also passed?

platform/src/main/java/org/stellar/anchor/platform/service/SimpleInteractiveUrlConstructor.java
```diff
@@ -104,7 +109,7 @@ String constructToken(
     return jwtService.encode(token);
   }
 
-  void forwardKycFields(Map<String, String> request) throws AnchorException {
+  void forwardKycFields(Map<String, String> request, Sep10Jwt jwt) throws AnchorException {
```
lijamie98, line 112: The `memo` and `account` are not KYC fields. Can we add them after line#58?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
Ifropc: Good point

PR #1185, opened 2023-11-02
[ANCHOR-459] Split SEP Tests of AnchorPlatformIntegrationTest

lijamie98:
### Description

- Split SEP Tests of AnchorPlatformIntegrationTest
- Replace `SepClient` classes with wallet sdk for some of the SEP tests.
- Disable the tests that requires non-`default` test configuration.
- Enable integration test concurrency.
- Add Mutex to reference server transaction submission function.

### Context

- Test improvements.

### Testing

- `./gradlew test`

### Known Limitaions
- Tests with non-`default` configurations will be moved to `extended-tests` subproject.


integration-tests/src/test/kotlin/org/stellar/anchor/platform/AnchorPlatformEnd2EndTest.kt
```diff
@@ -1,38 +0,0 @@
...
-
-  @Test
-  @Order(1)
-  fun runSep24Test() {
-    singleton.sep24E2eTests.testAll()
-  }
-
-  @Test
-  @Order(2)
-  fun runSep6Test() {
```
philipliu, line 33: I think the SEP-6 tests need to be migrated as well. It doesn't look like they're running.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
```diff
@@ -310,8 +331,10 @@ class Sep24End2EndTests(config: TestConfig, val jwt: String) {
         .sign(keypair)
         .sign(newAcc)
 
-    wallet.stellar().submitTransaction(tx)
-
+    submissionLock.withLock {
+      wallet.stellar().submitTransaction(tx)
+      delay(5000)
```
philipliu, line 336: This is fine but I think maybe adding a retry would be more robust. We don't have to do this now though.

lijamie98: The retry is already implemented: https://github.com/stellar/kotlin-wallet-sdk/blob/main/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/horizon/Stellar.kt#L142

philipliu (approved):

PR #1188, opened 2023-11-03
[ANCHOR-525] Introduce `NotifyAmountsAssetsUpdated` RPC

philipliu:
### Description

This introduces the `NotifyAmountsAssetsUpdated` RPC which allows SEP-6 implementations to update all amounts/assets when it calculates fees.

### Context

The current `NotifyAmountsUpdated` RPC was previously created for the SEP-24 withdrawal flow where only the amount values for `amount_out` and `amount_fee` need to be updated. In SEP-6, the asset values need to be updated as well since in the case of a deposit, the `amount_in` asset and amount are not known at transaction creation time. In the SEP-6 withdrawal flow, the `amount_out` asset and amount are unknown.

The current reference implementation works around this by updating the `amount_in` value and asset when it requests off-chain funds which really could be down when it updates the fees initially ([`Sep6EventProcessor`](https://github.com/stellar/java-stellar-anchor-sdk/blob/c0eb22b53acc17cf1840d0617f768ded53d8160d/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt#L203-L207)). 

Additionally, by not using this RPC, the SEP-6 implementation would not need to purposely set wrong amount values when creating the transaction ([`Sep6Service`](https://github.com/stellar/java-stellar-anchor-sdk/blob/c0eb22b53acc17cf1840d0617f768ded53d8160d/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java#L259-L266)).

### Testing

- `./gradlew test`

### Documentation

TODO: make the stellar-docs change if this is ok.

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/rpc/method/NotifyAmountsAssetsUpdatedRequest.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/rpc/method/NotifyAmountsAssetsUpdatedRequest.java
@@ -1,0 +1,27 @@
+package org.stellar.anchor.api.rpc.method;
+
+import com.google.gson.annotations.SerializedName;
+import javax.validation.constraints.NotNull;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@SuperBuilder
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class NotifyAmountsAssetsUpdatedRequest extends RpcMethodParamsRequest {
+
+  @SerializedName("amount_in")
+  @NotNull
+  private AmountAssetRequest amountIn;
+
+  @SerializedName("amount_out")
+  @NotNull
+  private AmountAssetRequest amountOut;
+
+  @SerializedName("amount_fee")
+  @NotNull
+  private AmountAssetRequest amountFee;
+}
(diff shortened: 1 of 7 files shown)


Ifropc (approved):

JakeUrban:
I'm not sure I fully understand the problem this is addressing and what the changes are.

The explanation of the `notify_amounts_updated` is as follows:

> If onchain funds were received, but for some reason the amount_in differs from specified in the interactive flow (amount_expected), you can update amount_out and amount_fee to make them correspond to the actual amount_in. The status of the transaction in this case won't be changed and will be equal to pending_anchor.

This method is really only for an edge case, which is when the amount the client said they would send is not what was actually sent.

For 99% of cases, amounts and assets should be set when requesting and receiving funds using the following methods:

- request_offchain_funds
- notify_offchain_funds_received
- request_onchain_funds
- notify_onchain_funds_received

Why can't SEP-6 also use these methods, or why aren't these methods sufficient?

philipliu:
@JakeUrban Ah, this was a misunderstanding on my part. This all stems from the fact that I was using `notify_amounts_updated` to update the fees before requesting KYC. This RPC requires `amount_fee` and `amount_out` to be previously set, which is not the case for SEP-6 when transactions are created. The SEP-6 reference implementation _does_ use those methods you listed so there's no need to introduce this new PR.

I'll update the docs accordingly and close this PR.

PR #1193, opened 2023-11-07
[ANCHOR-537] Re-enable SEP-6 e2e tests

philipliu:
### Description

This PR reintroduces the SEP-6 end-to-end tests. Previously, this test would get stuck in a bad state when the same dev stack was re-used between test runs. This was due to using the same SEP-10 account + memo for each test. Now, the tests will generate a random memo for each execution.

After reintroducing these tests, the end-to-end tests have become more flaky due to frequent concurrent Stellar transaction submissions. This commit replaces all instances of `submissionLock.withLock` with locking retry.

### Context

The test was removed.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


integration-tests/src/test/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
```diff
@@ -219,11 +214,9 @@ class Sep24End2EndTests : AbstractIntegrationTests(TestConfig(testProfileName =
         .build()
     transfer.sign(keypair)
 
-    submissionLock.withLock {
-      wallet.stellar().submitTransaction(transfer)
-      // Wait for the status to change to PENDING_USER_TRANSFER_END
-      waitForTxnStatus(withdrawTxn.id, COMPLETED, token)
-    }
+    transactionWithRetry { wallet.stellar().submitTransaction(transfer) }
```
lijamie98, line 217: I think we should keep the lock in addition to the lock
```suggestion
   submissionLock.withLock {
      transactionWithRetry { wallet.stellar().submitTransaction(transfer) }
   }
```

integration-tests/src/test/kotlin/org/stellar/anchor/platform/suite/AbstractIntegrationTests.kt
```diff
@@ -24,6 +32,19 @@ abstract class AbstractIntegrationTests(val config: TestConfig) {
   var walletKeyPair = SigningKeyPair.fromSecret(CLIENT_WALLET_SECRET)
   var anchor = wallet.anchor(config.env["anchor.domain"]!!)
   var token: AuthToken
+  private val submissionLock = Mutex()
+
+  suspend fun transactionWithRetry(transactionLogic: suspend () -> Unit) =
```
lijamie98, line 37: Nice function!

:nit: Can we add parameters to the function?

```suggestion
  suspend fun transactionWithRetry(retries=5, range=5, transactionLogic: suspend () -> Unit) =
```

kotlin-reference-server/src/main/kotlin/org/stellar/reference/Mutex.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.retryWhen
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+// This is to make sure transaction submission is mutual exclusive to avoid failures
+internal val submissionLock = Mutex()
+
+suspend fun transactionWithRetry(transactionLogic: suspend () -> Unit) =
```
lijamie98, line 13: Isn't this a duplicate?

integration-tests/src/test/kotlin/org/stellar/anchor/platform/suite/AbstractIntegrationTests.kt
lijamie98: Should we name the function with a more generic name?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/Mutex.kt
philipliu: Yeah, it's implemented twice because it needs to be used in the reference server as well as the tests.

integration-tests/src/test/kotlin/org/stellar/anchor/platform/suite/AbstractIntegrationTests.kt
philipliu: I think the naming makes sense, it's only meant to be used for submitting Stellar transactions because it also holds a lock. We should not be using it for anything else.

lijamie98: ok, we can rename when we have other use cases.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/Mutex.kt
lijamie98: Can we add to the `lib-util` sub-project?

philipliu: Can we delay this until we have more shared code? I'm not sure if we want to create a whole new project just to hold a function.

lijamie98 (approved):
LGTM

PR #1194, opened 2023-11-07
[ANCHOR-514] Populate sep38quote info into sep24txn

JiahuiWho:
### Description

If `quote_id` is provided upon receiving the sep24 request, fetch quote details and populate it into sep24txn object

Left out the quote validation since it's specified in doc 
> When `quote_id` is set, `asset_code`, `source_asset` and `amount` must be validated by the anchor, if present. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`.

### Context

Add `quote_id` and `source_asset_id` to sep24 txn so that quotes can be displayed to users while initiate deposit/withdraw sep24 transaction

### Testing

- `./gradlew test`

### Documentation

Updated in https://github.com/stellar/stellar-protocol/pull/1358


platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep24Transaction.java
```diff
@@ -135,4 +135,12 @@ public void setRefunds(Sep24Refunds refunds) {
   @SerializedName("refund_memo_type")
   @Column(name = "refund_memo_type")
   String refundMemoType;
+
```
lijamie98, line 138: Please also update the flyway schemas

JakeUrban:
Why we aren't performing validation against the quote's attributes the the requests attributes. If I initiate a SEP-24 deposit transaction and specify an `asset_code` that doesn't match the specified quote's buy asset, the anchor platform should return 400.

philipliu:
I think we would also need to update the SEP-38 implementation. Only SEP-6, and SEP-24 are supported currently https://github.com/stellar/java-stellar-anchor-sdk/blob/8e595eab7aa01147d610cb39313afa0c18d75630/api-schema/src/main/java/org/stellar/anchor/api/sep/sep38/Sep38Context.java#L6-L10.

core/src/main/java/org/stellar/anchor/sep24/Sep24Transaction.java
```diff
@@ -236,6 +236,18 @@ public interface Sep24Transaction extends SepTransaction {
 
   String getMessage();
 
+  String getQuoteId();
```
philipliu, line 239: We would also need to add this field to the platform transaction. It's being mapped here https://github.com/stellar/java-stellar-anchor-sdk/blob/8e595eab7aa01147d610cb39313afa0c18d75630/core/src/main/java/org/stellar/anchor/util/TransactionHelper.java#L172-L196

philipliu: The docs would need to be updated to include `quote_id` under the SEP-24 transaction as well. https://developers.stellar.org/api/anchor-platform/resources/get-transaction

core/src/test/kotlin/org/stellar/anchor/sep24/Sep24ServiceTest.kt
```diff
@@ -274,6 +328,8 @@ internal class Sep24ServiceTest {
     assertEquals(TEST_ASSET_ISSUER_ACCOUNT_ID, slotTxn.captured.requestAssetIssuer)
     assertEquals(TEST_ACCOUNT, slotTxn.captured.toAccount)
     assertEquals(TEST_CLIENT_DOMAIN, slotTxn.captured.clientDomain)
+    assertEquals(depositQuote.id, slotTxn.captured.quoteId)
+    assertEquals(depositQuote.sellAsset, slotTxn.captured.sourceAsset)
```
philipliu, line 332: Can we add more unit tests for the actual quote validation? I think you'll find it easier to test if the `validatedAndPopulateQuote` is in its own class. I'll leave that decision up to you, though.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -555,4 +572,74 @@ public InfoResponse getInfo() {
                 sep24Config.getFeatures().getClaimableBalances()))
         .build();
   }
+
+  private void validatedAndPopulateQuote(
```
philipliu, line 576: I wrote a utility class that does this validation for SEP-6. Do you think we can leverage it here? I think the logic should be the same. https://github.com/stellar/java-stellar-anchor-sdk/blob/8e595eab7aa01147d610cb39313afa0c18d75630/core/src/main/java/org/stellar/anchor/sep6/ExchangeAmountsCalculator.java#L31-L62

core/src/test/kotlin/org/stellar/anchor/sep24/Sep24ServiceTest.kt
```diff
@@ -53,6 +56,42 @@ internal class Sep24ServiceTest {
     const val TEST_SEP24_MORE_INFO_URL = "https://test-anchor.stellar.org/more_info_url"
     val TEST_STARTED_AT: Instant = Instant.now()
     val TEST_COMPLETED_AT: Instant = Instant.now().plusSeconds(100)
+    val DEPOSIT_QUOTE_JSON =
```
philipliu, line 59: Can we also add some integration tests?

lijamie98 (approved):
LGTM.

PR #1201, opened 2023-11-14
[ANCHOR-536] validate `SelfDepositInfoGenerator` in SEP6/24/31 Config

JiahuiWho:
### Description

Make distribution account optional

### Context

`distribution_account` is not a required field in asset, and this is validated by AssetsConfigTest and AssetsServiceTest that using test_assets.yaml, which includes no-distribution-account asset.

Also confirmed with @Ifropc with above to add tests for DepositInfoGenerator

### Testing

- `./gradlew test`


core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -212,11 +212,13 @@ public InteractiveTransactionResponse withdraw(
             .sep10Account(token.getAccount())
             .sep10AccountMemo(token.getAccountMemo())
             .fromAccount(sourceAccount)
-            // TODO - jamie to add unique address generator
-            .withdrawAnchorAccount(asset.getDistributionAccount())
             .toAccount(asset.getDistributionAccount())
             .clientDomain(token.getClientDomain());
 
+    if (asset.getDistributionAccount() != null) {
```
lijamie98, line 218: It is safer to use `isEmpty()`

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -217,12 +217,16 @@ public Sep31PostTransactionResponse postTransaction(
...
             .amountOut(null)
             .amountOutAsset(null)
-            // updateDepositInfo will update these ⬇️
             .stellarAccountId(assetInfo.getDistributionAccount())
             .stellarMemo(null)
             .stellarMemoType(null)
             .build();
 
+    // updateDepositInfo will update these ⬇️
+    if (assetInfo.getDistributionAccount() != null) {
```
lijamie98, line 226: nit: `isEmpty()`

lijamie98:
Please also add the validation of `self` deposit info generator for `SEP-31` and `SEP-6`.

If any of the asset does not have the `distribution_account` configured, the `self` deposit info generator should not be used.

Please also update the `anchor-config-default-values.yaml` file.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
Ifropc: Wouldn't it throw npe?

lijamie98: `StringHelper.isEmpty()` won't throw npe.

platform/src/main/java/org/stellar/anchor/platform/component/observer/PaymentObserverBeans.java
```diff
@@ -35,7 +37,10 @@ public StellarPaymentObserver stellarPaymentObserver(
     if (assetService == null || assetService.listAllAssets() == null) {
       throw new ServerErrorException("Asset service cannot be empty.");
     }
-    List<AssetInfo> stellarAssets = assetService.listStellarAssets();
+    List<AssetInfo> stellarAssets =
```
Ifropc: Isn't it equivalent to `assetService.listStellarAssets`?

Ifropc (approved):

lijamie98 (approved):
LGTM.

PR #1210, opened 2023-11-23
[ANCHOR-555] Add `client_domain` and `client_name` to Platform transactions

philipliu:
### Description

This adds the `client_domain` and `client_name` to the Platform transaction objects

### Context

`client_name` and `client_domain` are inaccessible to the business servers for SEP-6 and SEP-31 at the moment.

### Testing

- `./gradlew test`

### Documentation

The AP docs need to be updated with this new field.

### Known limitations

N/A


lijamie98 (commented):
LGTM

lijamie98 (approved):

core/src/main/java/org/stellar/anchor/sep6/Sep6Transaction.java
```diff
@@ -276,6 +276,15 @@ public interface Sep6Transaction extends SepTransaction {
 
   void setMemoType(String memoType);
 
+  /**
+   * The client domain of the wallet that initiated the transaction.
+   *
+   * @return the client domain.
+   */
+  String getClientDomain();
```
Ifropc, line 284: It's not a part of the standard, isn't it?
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#transaction-history
Why do we want to add client_domain ?
Should we use client_name instead? Derived from the configuration?

philipliu: > It's not a part of the standard, isn't it?

It will only be visible by querying the Platform API, it is not being mapped on the SEP side. https://github.com/stellar/java-stellar-anchor-sdk/blob/7e42b5ccc7cf82a19a29a84716a6141208b0cb7c/core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java#L37-L63

> Why do we want to add client_domain ?
Should we use client_name instead? Derived from the configuration?

I can add `client_name`, but should we keep `client_domain` as well? In SEP-24 both are included in the interactive JWT.

Ifropc: I think in general it's easier to work with client_name including both sounds like a good idea. But won't it require to add client_name to other SEPs too?

Ifropc (approved):

PR #1211, opened 2023-11-23
[ANCHOR-539 ]Map empty request parameter strings to null

philipliu:
### Description

This maps empty request parameters (`param=`) to `null`. There is a bug in Spring that maps these to "".

### Context

Since we don't explicitly check for empty parameters in SEP-6, an empty amount causes a `NumberFormatException` when converting it to a `BigDecimal`, which surfaces to the client as a 500 error.

### Testing

- `./gradlew test`
- Manual testing SEP-6 withdraw with`&amount=`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/controller/sep/SepControllerConfig.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/controller/sep/SepControllerConfig.java
@@ -1,0 +1,16 @@
+package org.stellar.anchor.platform.controller.sep;
+
+import org.springframework.beans.propertyeditors.StringTrimmerEditor;
+import org.springframework.web.bind.WebDataBinder;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.InitBinder;
+
+@ControllerAdvice
+public class SepControllerConfig {
+  @InitBinder
+  void initBinder(final WebDataBinder binder) {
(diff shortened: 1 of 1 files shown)


lijamie98 (commented):
LGTM

platform/src/main/java/org/stellar/anchor/platform/controller/sep/SepControllerConfig.java
```diff
@@ -0,0 +1,16 @@
+package org.stellar.anchor.platform.controller.sep;
+
+import org.springframework.beans.propertyeditors.StringTrimmerEditor;
+import org.springframework.web.bind.WebDataBinder;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.InitBinder;
+
+@ControllerAdvice
```
lijamie98, line 8: Does this happen to platform API as well?

lijamie98 (approved):

platform/src/main/java/org/stellar/anchor/platform/controller/sep/SepControllerConfig.java
philipliu: At the moment, no. I'll add the config anyway to the platform and custody controllers anyway to prevent future issues.

PR #1213, opened 2023-11-24
[ANCHOR-510] Fix duplicate client status callbacks in tests

philipliu:
### Description

This fixes the event duplication bug observed in the end-to-end tests.

### Context

There is no actual bug with the event processor, rather it's a bug within the client configuration. There are multiple clients with the same client callback URL. This causes each event to be sent to the wallet server once per client, causing duplication.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


lijamie98 (approved):
Good catch.

PR #1214, opened 2023-11-27
[ANCHOR-559] Remove duplicated and unused tests

lijamie98:
### Description

`sep24CustodyE2eTests`, `sep24CustodyRpcE2eTests` are the same as `sep24RpcE2eTests`.

`sep31CustodyRpcE2eTests` is the same as `sep31RpcE2eTests`. 

These duplicated tests should be removed.

### Context

Code clean up.

### Testing

- `./gradlew test`


extended-tests/src/test/kotlin/org/stellar/anchor/platform/AbstractIntegrationTest.kt
```diff
@@ -34,10 +34,7 @@ open class AbstractIntegrationTest(private val config: TestConfig) {
   lateinit var platformApiCustodyTests: PlatformApiCustodyTests
   lateinit var custodyApiTests: CustodyApiTests
   lateinit var sep24RpcE2eTests: Sep24BaseEnd2EndTest
-  lateinit var sep24CustodyE2eTests: Sep24BaseEnd2EndTest
```
philipliu, line 37: The classes are the same, but we should still run the custody end-to-end tests. The SEP-24 reference implementation has different behavior under the custody test profile ([code](https://github.com/stellar/java-stellar-anchor-sdk/blob/4c94a70cffd856682a993b738ebea98b872d43e6/kotlin-reference-server/src/main/kotlin/org/stellar/reference/sep24/DepositService.kt#L41-L49)). We should test this.

philipliu (changes requested):

philipliu (approved):

PR #1215, opened 2023-11-27
[ANCHOR-483] Add metrics for SEP-6

philipliu:
### Description

This adds metrics to SEP-6.

### Context

Observability

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


platform/src/main/java/org/stellar/anchor/platform/service/AnchorMetrics.java
```diff
@@ -1,10 +1,12 @@
 package org.stellar.anchor.platform.service;
 
 public enum AnchorMetrics {
+  SEP6_TRANSACTION_OBSERVED("sep6.transaction.observed"),
```
lijamie98, line 4: WDYT to merge `AnchorMetrics.java` with `MetricsConstants.java`?

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
```diff
@@ -257,5 +257,13 @@ void handleSep6Transaction(ObservedPayment payment, JdbcSep6Transaction txn)
         payment.getTransactionHash(),
         payment.getAmount(),
         rpcConfig.getCustomMessages().getIncomingPaymentReceived());
+
+    Metrics.counter(
+            AnchorMetrics.SEP6_TRANSACTION_OBSERVED.toString(),
+            "status",
+            SepTransactionStatus.PENDING_ANCHOR.toString())
+        .increment();
```
lijamie98, line 265: We have two metrics patterns. 

WDYT to use something like:
```suggestion
Counter sep6WithdrawalCounter =
      counter(
          MetricConstants.SEP6_TRANSACTION_OBSERVED
          MetricConstants.STATUS,
          MetricConstants.PENDING_ANCHOR);
sep6WithdrawalCounter.increment();
```

platform/src/main/java/org/stellar/anchor/platform/service/AnchorMetrics.java
philipliu: I think the metrics need to be refactored quite a bit. If it's all the same to you, I would like to do this in a separate PR.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
philipliu: I think we should probably differentiate tags and metrics either by type or by making them members of different classes.

platform/src/main/java/org/stellar/anchor/platform/service/AnchorMetrics.java
lijamie98: It's ok to be in a separate PR.

platform/src/main/java/org/stellar/anchor/platform/service/PaymentOperationToEventListener.java
lijamie98: Do it in a separate PR?

philipliu: Yep

PR #1216, opened 2023-11-27
[ANCHOR-550] Convert extended tests to JUnit

lijamie98:
### Description

- Create JUnit test `Sutie` to group the extended tests into 6 categories. `rpc`, `custody`, `auth-jwt-custody`, `auth-jwt-platform`, `auth-apikey-custody`, and `auth-apikey-platform`
- Fix the bug that the `custody` tests that can only run once. Create dynamic signatures when calling the fireblocks custody server.
- Modify the Github workflow files to run the extended tests with different test configurations.
- Updated `README` docs.

### Context

CI improvements.

### Testing

- `./gradlew test`

### Documentation
Updated the how to contribute documentation.


.github/workflows/sub_essential_tests.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   essential_tests:
     name: Run Essential Tests (Integration Tests, Faster End-2-End Tests, Unit Tests, and Stellar Validation Tools)
-    runs-on: ubuntu-latest-16-cores
+    runs-on: ubuntu-latest-8-cores
     #    runs-on: ubuntu-latest
```
philipliu, line 11: nit: can we remove the comment?

extended-tests/src/test/kotlin/org/stellar/anchor/platform/extendedtest/auth/apikey/custody/AuthApiKeyCustodyTests.kt
```diff
@@ -1,43 +1,18 @@
...
-      testProfileRunner.start()
-    }
-
-    @AfterAll
-    @JvmStatic
-    fun breakdown() {
-      testProfileRunner.shutdown()
-    }
-  }
+// use TEST_PROFILE_NAME = "auth-apikey-custody"
```
philipliu, line 14: This is an idea: is there a way to make the tests fail if they run against an unexpected profile?

platform/src/main/java/org/stellar/anchor/platform/config/FireblocksConfig.java
```diff
@@ -191,6 +191,7 @@ public void validateRetryDelay(Errors errors) {
    */
   public PublicKey getFireblocksPublicKey() throws InvalidConfigException {
     try {
+      System.out.println("publicKey=" + publicKey);
```
philipliu, line 194: Nit: remove println

philipliu (approved):
LGTM

extended-tests/src/test/kotlin/org/stellar/anchor/platform/extendedtest/auth/apikey/custody/AuthApiKeyCustodyTests.kt
lijamie98: The test will naturally fail if it is run against an unexpected profile. The old paradigm of running the docker compose with the test will be able to make sure of the consistency. However, we have switched to the new paradigm.

Another alternative is to have an end-point to expose the test-profile. However, this may cause other issues. WDYT?

platform/src/main/java/org/stellar/anchor/platform/config/FireblocksConfig.java
lijamie98: I was working on removing all the printlin statements too. :-)

extended-tests/src/test/kotlin/org/stellar/anchor/platform/extendedtest/auth/apikey/custody/AuthApiKeyCustodyTests.kt
philipliu: Yeah, on second thought, I don't think we should be doing this.

PR #1229, opened 2023-12-12
[ANCHOR-570] Remove Java Reference Server

lijamie98:
### Description

Remove Java Reference Server

### Context

Clean up the project.

### Testing

- `./gradlew test`


philipliu (approved):
We should update the README as well. I think there are still references to it.

PR #1232, opened 2023-12-13
[ANCHOR-544] Fix date fields in SEP-12 PUT /customer

philipliu:
### Description

This changes the `id_issue_date` and `id_expiration_date` type to a String instead of an Instant. Validation has been added to these fields to ensure they are in ISO-8601 date-only format before passing it to the business server.

### Context

SEP-9 date fields should be in ISO-8601 date-only format. 

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


Ifropc:
Shouldn't we accept ISO 8601 date time too for backward compatibility? Not sure if anybody uses it right not, but simple regex replacement of "T.*", "" could be sufficient

philipliu:
@Ifropc Good point. I think there is less room for error if we avoid using regex, so I'll parse instead.

Ifropc (approved):

Ifropc:
Ideally we would want to pass ISO-8601 date-only format to business server but I think date-time is acceptable

PR #1233, opened 2023-12-13
[PARTNER-218] Bump java SDK version in legacy 1.0

Ifropc:
### Description

Bump java SDK version in legacy 1.0

### Context

Using the latest Java SDK

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


JakeUrban (approved):
Note that 0.42 is compatible with testnet starting Dec 18th. They don't use the Stellar observer so they should be ok though.

PR #1245, opened 2024-01-03
[ANCHOR-577] Print Anchor Platform banner with version string

lijamie98:
### Description

Print Anchor Platform banner with version string

### Context

This is to improve the anchor platform developer expeirence.

### Testing

- `./gradlew test`


philipliu:
Is this also accessible from the docker image?

lijamie98:
> Is this also accessible from the docker image?

Yes it does. When running from the docker image, the banner is also printed.

philipliu (approved):

PR #1250, opened 2024-01-09
[ANCHOR-524] Change message db column type from VARCHAR to TEXT

JiahuiWho:
### Description

Change the type from VARCHAR to TEXT for column that stores generic message

### Context

Instead of validating the input length of each field on api level, this can be delegate to database to ensure the data integrity and cross-platform consistency. Thus changing the type from varchar to text for column that could contain generic messages.

### Testing

- `./gradlew test`


lijamie98:
How do we verify that the Flyway script does not break if there are existing transactions in the database?

JiahuiWho:
`ALTER TABLE` will not rewrite the table, since varchar and text have the same on-disk representation, both allow variable-length strings with virtually no difference in performance or storage.

db migration is only needed when the column has a index

philipliu (approved):
LGTM

PR #1252, opened 2024-01-10
[ANCHOR-566] Fix the assertion failure caused by SSEStream connection reset that caused the status to become YELLOW and remove event deletion

lijamie98:
### Description

When the test parallelization is turned on, the essential tests may fail from time to time.

- The SSEStream may cause IOException that causes the SSEStream to restart and set the observer status to YELLOW.
- Remove the reference server event deletion that causes instability of event assertion.
- Log the events to find the root cause of the event assertion failure

### Context

Improve test stability.

### Testing

- `./gradlew test`

### DIFF
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
@@ -31,6 +31,7 @@ import org.stellar.anchor.auth.Sep24InteractiveUrlJwt
 import org.stellar.anchor.platform.AbstractIntegrationTests
 import org.stellar.anchor.platform.CLIENT_WALLET_SECRET
 import org.stellar.anchor.platform.TestConfig
+import org.stellar.anchor.util.GsonUtils
 import org.stellar.anchor.util.Log.debug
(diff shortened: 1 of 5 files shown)


essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
```diff
@@ -83,7 +83,6 @@ open class Sep24End2EndTests : AbstractIntegrationTests(TestConfig()) {
   ) = runBlocking {
     val keypair = SigningKeyPair.fromSecret(walletSecretKey)
     walletServerClient.clearCallbacks()
-    anchorReferenceServerClient.clearEvents()
```
philipliu, line 86: Is the endpoint used anymore? Should we remove it?

philipliu (approved):

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
lijamie98: Yeah, I was debating... There might be future use case. So, I kept it.

I will remove from the client and still keep the end-point.

PR #1254, opened 2024-01-11
[ANCHOR-580] Disable ConfigManager health check

lijamie98:
### Description

Disable ConfigManager health check

### Context

The config manager health check prints config values which poses security vulnerability.

### Testing

- `./gradlew test`


essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
```diff
@@ -163,11 +163,9 @@ open class Sep24End2EndTests : AbstractIntegrationTests(TestConfig()) {
     actualEvents?.let {
       assertEquals(expectedStatuses.size, actualEvents.size)
 
-      GsonUtils.getInstance().toJson(expectedStatuses).let { json ->
-        println("expectedStatuses: $json")
-      }
+      println("expectedStatuses: " + GsonUtils.getInstance().toJson(expectedStatuses))
 
-      GsonUtils.getInstance().toJson(actualEvents).let { json -> println("actualEvents: $json") }
+      println("actualEvents: " + GsonUtils.getInstance().toJson(expectedStatuses))
```
philipliu, line 168: This isn't related to the healthcheck, right?

philipliu (commented):
Is this health check being used by anyone? Should we be fixing the exposure instead of disabling the health check?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep24End2EndTests.kt
lijamie98: That's good question. 

I think the ConfigMap contains the secret. The health check retrieves the details for the config which may lead to leak of secrets. I think it is safer to seal it for now so that we don't risk leaking it.

In the future when we implement the dynamic configuration or admin console, we should separate the secret from the ConfigManager.

PR #1256, opened 2024-01-11
[ANCHOR-579][Quick Fix] Fix Sep10 config home_domain coexist error

JiahuiWho:
### Description

Fix sep10 config reading `home_domain` and `home_domains` from different resource and causing conflicting

### Context

Currently Sep10Config will throw error if `home_domains` is set in env config, because it is also reading `home_domain` from default config, and these two field cannot coexist.

This is a quick fix. Will deprecate `home_domain` and move to use `home_domains` only in 3.0

### Testing

- `./gradlew test`
Tested with default config and multi-tenant config provided by SDP team, both work properly.


philipliu (approved):
LGTM

PR #1258, opened 2024-01-17
[ANCHOR-593] Fix extended tests post Testnet reset

philipliu:
### Description

The platform tests hardcode transaction hashes that get erased after a Testnet reset. This commit updates the hardcoded values with new transactions.

### Context

Testnet reset broke these tests.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

This will need to be repeated after the next Testnet reset. We need to look for a way around this.


JiahuiWho (approved):
Thanks for fixing this!

PR #1260, opened 2024-01-23
[ANCHOR-554] Implement SEP-12 binary fields

philipliu:
### Description

This adds for SEP-9 binary fields support in the SEP-12 PUT /customer endpoint. SEP-12 PUT /customer requests containing binary fields are expected to use the `multipart/form-data` content type. The Anchor Platform will act as a proxy, passing along the fields as is, to the business server.

I also added reference support for all SEP-9 fields since I'm already touching the code.

### Context

Previously, only string fields were supported.

### Testing

- `./gradlew test`

### Documentation

I will create a stellar-docs PR to add the new `multipart/form-data` callback endpoints.

### Known limitations

N/A


JakeUrban (approved):
This LGTM! Lets remember to update our documentation to notify businesses that they need to accept multipart/form-data requests on their `PUT /customer` business server endpoints if they want to accept binary fields.

PR #1261, opened 2024-01-26
[ANCHOR-590] Support wildcard `homeDomains` for sep10

JiahuiWho:
### Description

Introduce the capability to use a wildcard prefix in the SEP10_HOME_DOMAINS configuration. This enhancement will allow for more dynamic and flexible domain management, especially beneficial in multi-tenant settings.

### Context

1. Support wildcard match (e.g., *.stellar.org).

2. To Ensure backward compatibility, the existing configurations with specific domains (e.g. ap.stellar.org) continue to work seamlessly.

### Testing

- `./gradlew test`

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
+++ b/core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
@@ -51,6 +51,15 @@ public class Sep10Helper {
     }
   }
 
+  /**
+   * Checks if the given domain name matches any pattern or exact domain in the provided list.
+   *
+   * @param patternsAndDomains A list containing patterns and/or exact domain names to match
+   *     against.
+   * @param domainName The domain name to check for a match.
+   * @return true if the domain name matches any pattern or exact domain in the list, false
+   *     otherwise.
+   */
   public static Boolean isDomainNameMatch(List<String> patternsAndDomains, String domainName) {
     for (String patternOrDomain : patternsAndDomains) {
       if (patternOrDomain.contains("*")) {
@@ -72,6 +81,12 @@ public class Sep10Helper {
     return false;
   }
 
+  /**
+   * Retrieves the first non-wildcard domain name from the provided list of patterns and domains.
+   *
+   * @param patternsAndDomains A list containing patterns and/or exact domain names to search.
+   * @return The first exact domain name found in the list, or null if no exact domain is present.
+   */
   public static String getDefaultDomainName(List<String> patternsAndDomains) {
     for (String patternOrDomain : patternsAndDomains) {
       if (!patternOrDomain.contains("*")) {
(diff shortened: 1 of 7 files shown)


JakeUrban:
❓ should we support a wildcard prefix or just assume subdomains of domains provided are valid? For example, should anchor.stellar.org be valid if stellar.org was configured? Or should we require the business to explicitly configure a wildcard for subdomains?

core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
```diff
@@ -47,4 +50,34 @@ public static String fetchSigningKeyFromClientDomain(String clientDomain) throws
       throw new SepException(String.format("Invalid config: %s", e.getMessage()));
     }
   }
+
+  public static Boolean isDomainNameMatch(List<String> patternsAndDomains, String domainName) {
```
philipliu, line 54: Can we add Javadoc for these public methods?

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -137,7 +137,9 @@ void validateConfig(Errors errors) {
             "sep10-web-auth-domain-invalid",
             "The sep10.web_auth_domain does not have valid format.");
       }
-    } else if (homeDomains != null && homeDomains.size() > 1) {
+    } else if (homeDomains != null
+        && !homeDomains.isEmpty()
+        && (homeDomains.size() > 1 || homeDomains.get(0).contains("*"))) {
```
philipliu, line 142: Is the wildcard home domain always the first home domain in the list?

JiahuiWho: No this is to make sure `web_auth_domain` must be set if
1.  `homeDomains` has more than one value
or  2.  `homeDomains` has only one value and it is a wildcard domain

core/src/main/java/org/stellar/anchor/sep10/Sep10Helper.java
```diff
@@ -47,4 +50,34 @@ public static String fetchSigningKeyFromClientDomain(String clientDomain) throws
...
+        // Escape special characters in the pattern and replace '*' with '.*'
+        String regex = patternOrDomain.replace(".", "\\.").replace("*", ".*");
+        Pattern patternObject = Pattern.compile(regex);
+        Matcher matcher = patternObject.matcher(domainName);
+        if (matcher.matches()) {
+          return true;
+        }
+      } else {
+        // exact domain
+        if (patternOrDomain.equals(domainName)) {
```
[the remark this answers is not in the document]
JiahuiWho: Oh I separate the check for wildcard domain and exact domain .. so no pattern or matcher created for exact domain

philipliu (approved):
LGTM

PR #1262, opened 2024-01-30
[ANCHOR-563 ]Add reference server to `docker-compose.yaml`

philipliu:
### Description

This adds `reference-server` service to the existing `docker-compose.yaml`.  

### Context

Ease of wallet testing.

### Testing

- `./gradlew test`
- Testing with demo wallet revealed that the reference server is incorrectly configured when running from docker. The Kafka bootstrap server is modified to point to the `kafka` hostname instead.

### Documentation

Will update the README with instructions for running a pre-configured AP.

### Known limitations

This will be fully tested after the PR is merged the `edge` image is updated in Dockerhub

### DIFF
--- a/kotlin-reference-server/src/main/resources/default-config.yaml
+++ b/kotlin-reference-server/src/main/resources/default-config.yaml
@@ -44,7 +44,7 @@ event:
   # Enables the Kafka event processor.
   enabled: true
   # The Kafka boostrap server.
-  bootstrapServer: localhost:29092
+  bootstrapServer: kafka:29092
   # The AnchorEvent topic to subscribe to.
   topic: TRANSACTION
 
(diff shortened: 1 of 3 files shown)


Ifropc:
Does simply running `docker-compose up` spin up the whole stack which could be used to testing for the wallet implementation?

philipliu:
@Ifropc Yep, I was able to start the SEP-6 flows using the demo wallet.

Ifropc:
Do you mind updating README just adding a short notice about docker-compose? (And maybe cover how to run it with demo-wallet)

Ifropc (approved):

philipliu:
@Ifropc Will do after I confirm everything works as expected

PR #1266, opened 2024-01-31
[ANCHOR-587] Disable URL encoding of SEP-10 challenge response

philipliu:
### Description

This commit fixes the URL encoding of the challenge transaction returned by the SEP-10 endpoint. This is done by disabling the "HTML escaping" done by the Gson instance used throughout the project. It is safe to do so because Anchor Platform only returns HTTP responses with the `application/json` content type, which does not require URL encoding.

### Context

The challenge transaction returned by the `/auth` endpoint is not decodable.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/Sep10Tests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/Sep10Tests.kt
@@ -1,5 +1,9 @@
 package org.stellar.anchor.platform.integrationtest
 
+import com.google.gson.*
+import com.google.gson.stream.JsonReader
+import com.google.gson.stream.JsonWriter
+import java.io.IOException
 import java.util.*
 import kotlin.test.assertFailsWith
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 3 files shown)


essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/Sep10Tests.kt
```diff
@@ -30,12 +31,18 @@ class Sep10Tests : AbstractIntegrationTests(TestConfig()) {
...
-            CLIENT_WALLET_EXTRA_SIGNER_2_SECRET
-          )
+            CLIENT_WALLET_EXTRA_SIGNER_2_SECRET,
+          ),
         )
     }
   }
 
+  @Test
+  fun testChallengeSerialization() {
```
Ifropc, line 41: Does this test fail without this change?

Ifropc (approved):

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/Sep10Tests.kt
philipliu: It doesn't 🤦 . Good catch, let me fix this test.

PR #1268, opened 2024-02-08
[ANCHOR-610] Fix log level

Ifropc:
### Description

Remove hardcoded DEBUG level of Stellar and add support of dynamic logging configuration reload

### Context

Previously, it was impossible to set DEBUG level to Stellar level config

### Testing

Manually tested with different log levels

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/SpringFrameworkConfigurator.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/SpringFrameworkConfigurator.java
@@ -15,11 +15,12 @@ public class SpringFrameworkConfigurator extends AbstractConfigurator
     PropertiesPropertySource pps = createPrefixedPropertySource("spring");
     pps.getSource()
         .forEach(
-            (String k, Object v) -> {
-              if (k.startsWith("logging.level")) {
+            (String cfgName, Object cfgValue) -> {
+              if (cfgName.startsWith("logging.level") && cfgValue instanceof String) {
                 LoggingSystem system =
                     LoggingSystem.get(SpringFrameworkConfigurator.class.getClassLoader());
-                system.setLogLevel(k.replace("logging.level.", ""), LogLevel.valueOf((String) v));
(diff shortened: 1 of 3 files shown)


platform/src/main/java/org/stellar/anchor/platform/configurator/SpringFrameworkConfigurator.java
```diff
@@ -11,6 +13,16 @@ public class SpringFrameworkConfigurator extends AbstractConfigurator
   public void initialize(@NotNull ConfigurableApplicationContext applicationContext) {
     // Load and add the data access settings to Spring `Environment`
     PropertiesPropertySource pps = createPrefixedPropertySource("spring");
+    pps.getSource()
+        .forEach(
+            (String cfgName, Object cfgValue) -> {
+              if (cfgName.startsWith("logging.level") && cfgValue instanceof String) {
+                LoggingSystem system =
```
philipliu, line 20: nit: Do we need to call this for every source?

platform/src/main/resources/log4j2-spring.yaml
```diff
@@ -53,12 +53,4 @@ Configutation:
       level: info
       AppenderRef:
         - ref: ${env:LOG_APPENDER:-console_appender}
-    Logger:
```
philipliu, line 56: What are the log levels set to after removing this?

Ifropc: Default is always INFO

platform/src/main/java/org/stellar/anchor/platform/configurator/SpringFrameworkConfigurator.java
Ifropc: Actually not, but it's just for one partner in the legacy branch (and most likely they will only use it for debug logs), so I decided to keep it simple

philipliu (approved):

PR #1271, opened 2024-02-20
[ANCHOR-612] Fix Sep10 validate wildcard domain failure

JiahuiWho:
### Description

Decode the challenge XDR and extract the home domain

Verify the domain matches a configured home domain or wildcard

Call the SDK’s parse challenge method with the extracted home domain

### Testing

- `./gradlew test`


core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -288,6 +288,8 @@ void validateChallengeRequest(
     Set<Sep10Challenge.Signer> signers = fetchSigners(account);
     // the signatures must be greater than the medium threshold of the account.
     int threshold = account.getThresholds().getMedThreshold();
+    Network network = new Network(appConfig.getStellarNetworkPassphrase());
+    String homeDomain = extractHomeDomainFromChallengeXdr(request.getTransaction(), network);
```
philipliu, line 292: This method can throw `InvalidSep10ChallengeException` which is an SDK-specific error, we should catch it and throw an AP error here.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -430,6 +437,47 @@ String generateSep10Jwt(ChallengeTransaction challenge, String clientDomain, Str
...
+
+  /**
+   * Extracts the home domain from a Stellar SEP-10 challenge transaction represented in XDR format.
+   *
+   * @param challengeXdr SEP-0010 transaction challenge transaction in base64.
+   * @param network The network to connect to for verifying and retrieving.
+   * @return The extracted home domain from the Manage Data operation within the SEP-10 challenge
+   *     transaction.
+   * @throws IOException If read XDR string fails, the exception will be thrown.
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be
```
philipliu, line 449: This isn't really SEP-10 validation failing, rather it is the SEP-10 challenge transaction being malformed.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -430,6 +437,47 @@ String generateSep10Jwt(ChallengeTransaction challenge, String clientDomain, Str
     debug("jwtToken:", sep10Jwt);
     return jwtService.encode(sep10Jwt);
   }
+
+  /**
+   * Extracts the home domain from a Stellar SEP-10 challenge transaction represented in XDR format.
+   *
+   * @param challengeXdr SEP-0010 transaction challenge transaction in base64.
```
philipliu, line 444: nit: Let's use SEP-10 or SEP-0010 everywhere.

philipliu (commented):
I think the code changes look good overall, but can we add an integration test as well?

philipliu (approved):

PR #1278, opened 2024-03-01
[ANCHOR-439] Listen to webhooks in reference server

philipliu:
### Description

Listens to webhooks instead of consuming directly from the internal `TRANSACTION` topic in the reference server.

### Context

This is the recommended approach for businesses building on the Anchor Platform.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
+++ b/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
@@ -4,7 +4,6 @@ import java.time.Instant
 import java.util.*
 import kotlinx.coroutines.runBlocking
 import org.stellar.anchor.api.callback.GetCustomerRequest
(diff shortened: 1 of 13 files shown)


kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -75,71 +74,65 @@ class Sep6EventProcessor(
...
+                // transferred 1 USD to the Anchor's bank account
+                if (transaction.amountExpected.amount.equals("0")) {
+                  "1"
+                } else {
+                  transaction.amountExpected.amount
+                },
+              )
           }
+          onchainPayments[transaction.id] = stellarTxnId
+          patchTransaction(
```
Ifropc, line 110: Why do we still have patch transaction here?

Ifropc (approved):

PR #1282, opened 2024-03-05
[ANCHOR-625] Fix docker compose stack

philipliu:
### Description

This commit fixes the docker-compose stack. Also, this replaces the Anchor Platform image used by the stack with a locally built one.

### Context

Two things broke the stack:
- We are listening to webhook events instead of directly consuming Kafka, we need to start the `event-processor`.
- Docker ports were incorrectly mapped

### Testing

- `./gradlew test`
- Tested SEP-6 deposit and withdrawal using demo-wallet

### Documentation

N/A

### Known limitations

Future bugs like this will be difficult to catch without running automated tests against the docker-compose stack. Rather than directly running the `ServiceRunner` in GH actions, we can use docker-compose instead and run our tests against that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-## TODO: this should use alpine when the platform is not Apple Silicon
-FROM gradle:7.6.4-jdk11 AS build
+ARG BASE_IMAGE=gradle:7.6.4-jdk11-alpine
+
+FROM ${BASE_IMAGE} AS build
 WORKDIR /code
 COPY --chown=gradle:gradle . .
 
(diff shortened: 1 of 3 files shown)


Ifropc (approved):
LGTM
Question: do we not use docker-compose that is there in tests? Should we add a step in pipeline that just does basic sanity tests using this compose file? (Like pinging /health routes)

philipliu:
We don't. Our current CI setup runs the tests against Jar. Pinging the server is a good place to start, although it wouldn't catch issues like not starting up the event processor. That would only be caught by running the tests.

PR #1290, opened 2024-03-11
[ANCHOR-614] Config more_info_url in sep6

JiahuiWho:
### Description

Add "more_info_url" config to sep6

### Context

This was originally missing from sep6 config, also drop this column from db

### Testing

- `./gradlew test`


core/src/main/java/org/stellar/anchor/sep24/MoreInfoUrlConstructor.java
```diff
@@ -2,8 +2,9 @@
 
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
+import org.stellar.anchor.SepTransaction;
 
 public abstract class MoreInfoUrlConstructor {
-  public abstract String construct(Sep24Transaction txn)
+  public abstract String construct(SepTransaction txn)
```
philipliu, line 8: Should we move this out of the `sep24` package now that it's being used by SEP-6 and SEP-24?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -81,29 +85,31 @@ public String encode(Sep10Jwt token) {
...
 
   public String encode(Sep24MoreInfoUrlJwt token) throws InvalidConfigException {
     if (sep24MoreInfoUrlJwtSecret == null) {
       throw new InvalidConfigException(
           "Please provide the secret before encoding JWT for more_info_url");
     }
+    return encodeForSep6Sep24Url(token, sep24MoreInfoUrlJwtSecret);
+  }
+
+  private String encodeForSep6Sep24Url(AbstractJwt token, String secret) {
```
philipliu, line 112: nit: What do you think of calling this `encodeForMoreInfoUrl`?

core/src/main/java/org/stellar/anchor/auth/Sep6MoreInfoUrlJwt.java
```diff
@@ -0,0 +1,23 @@
+package org.stellar.anchor.auth;
+
+import static org.stellar.anchor.auth.JwtService.CLIENT_DOMAIN;
+import static org.stellar.anchor.auth.JwtService.CLIENT_NAME;
+
+import io.jsonwebtoken.Jwt;
+import org.stellar.anchor.api.exception.SepException;
+
+public class Sep6MoreInfoUrlJwt extends AbstractJwt {
```
philipliu, line 9: nit: There's nothing specific to SEP-6 about `more_info_url` JWT, can we merge this with `Sep24MoreInfoJwt` and call it `MoreInfoJwt`?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -67,7 +67,7 @@ public class Sep24Service {
   final Sep24TransactionStore txnStore;
   final EventService.Session eventSession;
   final InteractiveUrlConstructor interactiveUrlConstructor;
-  final MoreInfoUrlConstructor moreInfoUrlConstructor;
+  final MoreInfoUrlConstructor sep24moreInfoUrlConstructor;
```
philipliu, line 70: Just curious, why did we rename this field?

core/src/main/java/org/stellar/anchor/sep6/Sep6TransactionUtils.java
```diff
@@ -1,20 +1,26 @@
...
    * Converts a SEP-6 database transaction object to a SEP-6 API transaction object.
    *
    * @param txn the SEP-6 database transaction object
+   * @param moreInfoUrlConstructor the more_info_url constructor created from SEP-6 config
    * @return the SEP-6 API transaction object
    */
-  public static Sep6TransactionResponse fromTxn(Sep6Transaction txn) {
+  public static Sep6TransactionResponse fromTxn(
+      Sep6Transaction txn, MoreInfoUrlConstructor moreInfoUrlConstructor)
+      throws MalformedURLException, URISyntaxException {
```
philipliu, line 23: Since we never catch and handle the URL exceptions, I think we should catch them and throw a suitable `RuntimeException` instead. Otherwise, you have leaking `throws MalformedURLException, URISyntaxException` to the caller of this method.

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
```diff
@@ -0,0 +1,68 @@
...
+    this.jwtService = jwtService;
+  }
+
+  @SneakyThrows
+  @Override
+  public String construct(SepTransaction txn) {
+    return construct((Sep6Transaction) txn);
+  }
+
+  @SneakyThrows
```
philipliu, line 36: What checked exceptions is this catching?

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
```diff
@@ -0,0 +1,68 @@
...
+  private final JwtService jwtService;
+
+  public Sep6MoreInfoUrlConstructor(
+      PropertyClientsConfig clientsConfig, MoreInfoUrlConfig config, JwtService jwtService) {
+    this.clientsConfig = clientsConfig;
+    this.config = config;
+    this.jwtService = jwtService;
+  }
+
+  @SneakyThrows
```
philipliu, line 30: Do we need `@SneakyThrows` here if the called `construct` doesn't throw checked exceptions?

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
```diff
@@ -0,0 +1,68 @@
...
+import org.stellar.anchor.SepTransaction;
+import org.stellar.anchor.auth.JwtService;
+import org.stellar.anchor.auth.Sep6MoreInfoUrlJwt;
+import org.stellar.anchor.platform.config.MoreInfoUrlConfig;
+import org.stellar.anchor.platform.config.PropertyClientsConfig;
+import org.stellar.anchor.sep24.MoreInfoUrlConstructor;
+import org.stellar.anchor.sep6.Sep6Transaction;
+import org.stellar.anchor.util.ConfigHelper;
+
+public class Sep6MoreInfoUrlConstructor extends MoreInfoUrlConstructor {
```
philipliu, line 18: Do we need to split `SimpleMoreInfoUrlConstructor` into SEP-6 and SEP-24 specific implementations? It doesn't look like there is anything specific to the SEPs here.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -295,9 +295,21 @@ sep1:
...
+  # For details, please refer to https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#transaction-history
+  #   base_uri: is the base_uri used to construct the more_info_url
+  #   jwt_expiration: the timeout in seconds of the JWT returned with the embedded interactive url of the SEP-6 process.
+  #       The jwt secret must be set by the environment variable: secret.sep6.more_info_url.jwt_secret
+  #   txn_fields: is an optional list of transaction fields that will be appended to the jwt of the more_info url. If empty, no fields from transaction will be appended.
+  #   In addition to the txn_fields, the following fields are also added to the query parameters.
+  #       `transaction_id`: the transaction ID
+  #       `token`: the JWT token
+  more_info_url:
+    base_url: http://localhost:8080/sep6/transaction/more_info
```
philipliu, line 308: It's a bit weird to have a default base URL since the default points to an endpoint that doesn't exist on the SEP server. I guess this is for consistency with SEP-24 but I don't think there's any problem with removing it from both. What do you think?

core/src/main/java/org/stellar/anchor/auth/Sep6MoreInfoUrlJwt.java
JiahuiWho: We need to differentiate the JWT (sep6/sep24) so that we know which secret to use when encoding

platform/src/main/resources/config/anchor-config-default-values.yaml
JiahuiWho: IMO this acts more like a placeholder. `base_url` is a required field and validation error will be thrown if we left it empty.

Remove it would require us to add them as env var to all the other config under service-runner/src/main/resources/profiles

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
JiahuiWho: No there's nothing specific between SEP6 and SEP24, but this constructor is specific to SEP6 and SEP24, distinct from the rest SEPs.

1. The functions `construct()` is tied to Sep6 or Sep24 txn (`getClientDomain`, `getSep10Account`, `getTransactionId`, `getClientDomain` are sep6 and sep24 txn only). One way to do it is that we extract all these outside and make `construct()` a more general function, or we need to do type check and casting everywhere.

2. We need to differentiate these 2 `MoreInfoUrlConstructor` when creating the beans or it will leads to unexpected behavior or errors when other beans want to reference them (e.g `Sep6Service` and `Sep24Service` need corresponding `MoreInfoUrlConstructor`). These might be achieved with `@Qualifier` annotation if we can merge these 2 `MoreInfoUrlConstructor` but I haven't try that.

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
JiahuiWho: SepValidationException from the clientConfig, when non-custodial client doesn't specify client_domain, which I think should be caught already during the initialization phase

URIException from building the url

InvalidConfigException if no secret was provided for encoding the token

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
JiahuiWho: It doesn't have to... I might have changed it because I'm splitting the 2 url constructor.

core/src/main/java/org/stellar/anchor/MoreInfoUrlConstructor.java
```diff
@@ -0,0 +1,5 @@
+package org.stellar.anchor;
+
+public abstract class MoreInfoUrlConstructor {
```
lijamie98, line 3: There is no implementation here. I would suggest to use interface.

core/src/main/java/org/stellar/anchor/auth/JwtService.java
lijamie98: nit: I would prefer `encodeForMoreInfoUrl` though because the function has nothing specific with sep6 or sep24. It just encode and sign the URL

JiahuiWho: Ok  I will change it to `encodeForUr`l because it is also used to encode **interactive** url, so I won't put more_info in the naming

core/src/main/java/org/stellar/anchor/MoreInfoUrlConstructor.java
JiahuiWho: I don't think I get you comment. I have 2 separate `Sep6MoreInfoUrlConstructor` and `Se24MoreInfoUrlConstructor` extends it and implement the function. Can you elaborate more on the question?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -30,14 +32,12 @@ public class Sep6Service {
...
   private final RequestValidator requestValidator;
-
   private final ClientFinder clientFinder;
   private final Sep6TransactionStore txnStore;
   private final ExchangeAmountsCalculator exchangeAmountsCalculator;
   private final EventService.Session eventSession;
-
   private final InfoResponse infoResponse;
-
+  private final MoreInfoUrlConstructor sep6MoreInfoUrlConstructor;
```
lijamie98, line 40: nit: this is inside Sep6Service, can we call this `moreInfoUrlConstructor`?

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep24Config.java
```diff
@@ -140,7 +127,7 @@ void validateInteractiveUrlConfig(Errors errors) {
   void validateMoreInfoUrlConfig(Errors errors) {
     if (moreInfoUrl == null) {
       errors.rejectValue(
-          "moreInfoUrl", "sep24-moreinfo-url-invalid", "sep24.more-info-url is not defined.");
+          "moreInfoUrl", "sep24-more-info-url-invalid", "sep24.more-info-url is not defined.");
```
lijamie98, line 130: 💡 good catch

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
lijamie98: In this case, can we consider the factory pattern?
https://refactoring.guru/design-patterns/factory-method

philipliu: @JiahuiWho That makes sense, I'm ok with leaving it like this or using the factory pattern.

platform/src/main/resources/config/anchor-config-default-values.yaml
philipliu: Leaving it empty is almost the same thing as having a default URL since it points to an endpoint that doesn't exist. I honestly think we should make it optional and remove the validation. 

Thoughts @lijamie98 ?

platform/src/main/java/org/stellar/anchor/platform/service/Sep6MoreInfoUrlConstructor.java
JiahuiWho: I would prefer to do refactor in a separate PR and let this one serve it's purpose.

philipliu (approved):
LGTM

PR #1291, opened 2024-03-14
[ANCHOR-630] SEP-24 asset and amount related request parameters are not working correctly

lijamie98:
### Description

SEP-24 asset and amount related request parameters are not working correctly 

- Store `amount`, `source_asset`, and `destination_asset` in database as `amount_in_asset` and `amount_out_asset`
- Remove unused `source_asset` and `destination_asset` fields from Sep24Transaction
- Forward `customer_id` to JWT if exists in the request.

### Context
Bug fixes.

### Testing

- `./gradlew test`

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -252,10 +248,13 @@ public InteractiveTransactionResponse withdraw(
...
     String quoteId = withdrawRequest.get("quote_id");
+    AssetInfo buyAsset = assetService.getAssetByName(withdrawRequest.get("destination_asset"));
     if (quoteId != null) {
-      AssetInfo buyAsset = assetService.getAssetByName(withdrawRequest.get("destination_asset"));
-      this.validatedAndPopulateQuote(
+      validateAndPopulateQuote(
           quoteId, asset, buyAsset, strAmount, builder, WITHDRAWAL.toString(), txnId);
+    } else {
+      builder.amountExpected(strAmount);
+      if (buyAsset != null) builder.amountOutAsset(buyAsset.getSep38AssetName());
```
Ifropc, line 257: Should it not be `amountInAsset`?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
```diff
@@ -420,10 +419,13 @@ public InteractiveTransactionResponse deposit(Sep10Jwt token, Map<String, String
...
     String quoteId = depositRequest.get("quote_id");
+    AssetInfo sellAsset = assetService.getAssetByName(depositRequest.get("source_asset"));
     if (quoteId != null) {
-      AssetInfo sellAsset = assetService.getAssetByName(depositRequest.get("source_asset"));
-      this.validatedAndPopulateQuote(
+      validateAndPopulateQuote(
           quoteId, sellAsset, asset, strAmount, builder, DEPOSIT.toString(), txnId);
+    } else {
+      builder.amountExpected(strAmount);
+      if (sellAsset != null) builder.amountInAsset(sellAsset.getSep38AssetName());
```
Ifropc, line 428: Looks like a bug -> we not set amountInAsset anymore for when quote is provided. Should this be lifted outside of the parent "if" statement? (I.e. move it to the line below 422)

Ifropc, line 428: Should we also call `builder.amountIn("0")` for consistency with SEP-6 impl?

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
lijamie98: In `withdraw`, the buy should be `amountOutAsset`?

ex: sell USDC and buy USD. So, the `amountOutAsset` should be USD.

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
lijamie98: The `validateAndPopulateQuote` function sets `amountInAsset`.

lijamie98: oh yeah. Updated.

Ifropc: Ah you are right I saw deleted code and for some reason I thought it didn't do that

core/src/main/java/org/stellar/anchor/sep24/Sep24Service.java
Ifropc: I didn't see it was withdraw, you are right

Ifropc (approved):

PR #1294, opened 2024-03-19
[ANCHOR-484] Add 2.0 Helm Charts

philipliu:
### Description

This adds Anchor Platform 2.x Helm charts. To support this change, it required refactoring the Kafka and the reference server configurations. The Helm Charts added by this PR should be considered a work in progress. The values file is subject to change and needs to be documented before being released publicly.

To test this, you should deploy all charts along with the Postgres, and Kafka instances. You can modify the secret store to pass in credentials.

### Context

We are deploying the Anchor Platform to Kubernetes.

### Testing

- `./gradlew test`
- Testing with demo wallet.

### Documentation

This is missing READMEs and values file documentation. I will address this in a follow-up PR to limit the LoC changes.

### Known limitations

N/A


helm-charts/anchor-platform/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Anchor Platform 2.0 Helm Chart
+maintainers:
+  - name: Philip Liu
+    email: philip.liu@stellar.org
+sources:
+  - https://github.com/stellar/java-stellar-anchor-sdk
+name: anchor-platform
+version: 0.1.0
```
reecexlm, line 10: this verison number should be greater than the current existing anchor platform chart?

reecexlm:
this looks great.  reviewed, and overall looks good.  did you test with "all defaults" or "minimal values" file so that a user could get it up and running locally with little to no values  ? This can help in identifying the required configuration items that are essential for the application to run. It also makes it easier for someone new to the chart to understand the minimum setup needed.

reecexlm:
one crucial test will be to outline steps to migrate from 1.x helm chart to 2.x.  we could interate on that after this is merged and we get more testing with this chart.   some areas where some additional changes fixes might be needed because they are not present with this PR include: testing with installing from our cicd pipeline,  instructions to install chart, run tests against local k8s cluster.   once both of these are done it may lead to additional changes with goal in mind to make this as easy, flexible as possible

reecexlm (approved):

helm-charts/anchor-platform/Chart.yaml
philipliu: I was thinking of this as a completely separate chart so the versioning would start at 0.1.0. I can rename it though and bump the version to 1.0.

philipliu:
> this looks great.  reviewed, and overall looks good.  did you test with "all defaults" or "minimal values" file so that a user could get it up and running locally with little to no values  ? 

I guess minimal values is the most accurate. I haven't settled on what should be configured/have default values yet. Once we decide on that, I will update the values file so that it includes the minimum values to start the application.

PR #1303, opened 2024-03-20
ANCHOR-638: Auth header validation + Improved JWT secret validation

Ifropc:
### Description

- Implementation of the protocol change [https://github.com/stellar/stellar-protocol/pull/1470](1470: Add Authorization header for SEP-10 GET /Auth)
- Migration to the latest version of JJWT library
- Improvement of working with secrets in the code
- Improved JWT secret validation

### Context

- Protocol change
- Old JJWT library doesn't have a good support for the algorithm used by the protocol change
- Refactored how we work with JWT secrets in the code: use SecretKey instead of Sting
- Previously, weak keys were not validated properly and was allowed to use, even though it's against the guidelines. Now, this keys are properly validated using underlying JJWT library.

### Testing

- `./gradlew test`
- Add tests for SEP-10 auth header 

### Documentation

N/A

### Known limitations

N/A


core/build.gradle.kts
```diff
@@ -7,6 +7,8 @@ plugins {
   alias(libs.plugins.kotlin.jvm)
 }
 
+repositories { maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") } }
```
lijamie98, line 10: Isn't this available here?
https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -50,25 +60,25 @@ public JwtService(
...
-    this.sep10JwtSecret = toBase64OrNull(sep10JwtSecret);
-    this.sep24InteractiveUrlJwtSecret = toBase64OrNull(sep24InteractiveUrlJwtSecret);
-    this.sep24MoreInfoUrlJwtSecret = toBase64OrNull(sep24MoreInfoUrlJwtSecret);
-    this.callbackAuthSecret = toBase64OrNull(callbackAuthSecret);
-    this.platformAuthSecret = toBase64OrNull(platformAuthSecret);
-    this.custodyAuthSecret = toBase64OrNull(custodyAuthSecret);
+    this.sep10JwtSecret = toSecretKeySpecOrNull(sep10JwtSecret);
+    this.sep24InteractiveUrlJwtSecret = toSecretKeySpecOrNull(sep24InteractiveUrlJwtSecret);
+    this.sep24MoreInfoUrlJwtSecret = toSecretKeySpecOrNull(sep24MoreInfoUrlJwtSecret);
+    this.callbackAuthSecret = toSecretKeySpecOrNull(callbackAuthSecret);
```
lijamie98, line 66: Is it necessary to have the secret?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -193,7 +185,25 @@ public <T extends AbstractJwt> T decode(String cipher, Class<T> cls)
...
 
-  private String toBase64OrNull(String value) {
-    return value == null ? null : Base64.encodeBase64String(value.getBytes(StandardCharsets.UTF_8));
+  private SecretKey toSecretKeySpecOrNull(String secret) {
+    return isEmpty(secret) ? null : Keys.hmacShaKeyFor(((secret.getBytes(StandardCharsets.UTF_8))));
+  }
+
+  @SneakyThrows
+  public static Jws<Claims> getHeaderJwt(String signingKey, String cipher) {
+    var factory = KeyFactory.getInstance("Ed25519");
```
lijamie98, line 194: Are we switching to asym key?

core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
```diff
@@ -752,4 +748,45 @@ internal class Sep10ServiceTest {
     val validationRequest = ValidationRequest.of(transaction.toEnvelopeXdrBase64())
     assertDoesNotThrow { sep10Service.validateChallenge(validationRequest) }
   }
+
+  // ----------------------
+  // Signature header tests
+  //
+
+  //  val domainKp =
```
lijamie98, line 756: Remove?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
Ifropc: asym key is used for a new feature (auth header signing with Stellar private keys, which are Ed25519), I'm drafting protocol change in parallel. Wanted to write some code first to make sure it's a working solution (and testing that it can be implemented in Java, JS, Ruby and Go -> primary langs where this feature)

core/src/main/java/org/stellar/anchor/auth/JwtService.java
Ifropc: Sorry what do you mean?

core/build.gradle.kts
Ifropc: This is for SNAPSHOT wallet SDK. I'll merge Wallet SDK changes and release new version before merging this PR (so will remove snapshot repo)

core/src/test/kotlin/org/stellar/anchor/sep10/Sep10ServiceTest.kt
Ifropc: This part is still WIP :)

core/src/main/java/org/stellar/anchor/auth/JwtService.java
lijamie98: The way we use JWT for SEP10 and SEP24 do not seem to require a fixed secret. Is it possible to generate a random secret so that we can further simplify the secret configuration.

Ifropc: SEP-10 and SEP-24 do require fixed JWT. Otherwise, is we generate it randomly, tokens issued between the starts will not be valid. 
However, I like this idea and I think we can implement it in v3 (for anchors that are fine with tokens being invalid between restarts)

JakeUrban: I think we need the SEP-24 JWT secret to be configured by the business so that they can validate the signature before authenticating the user in their webview, but the SEP-10 JWT secret could be internalized since the business never needs to validate the SEP-10 signature itself.

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -47,20 +53,23 @@ public Sep10Service(
...
     this.horizon = horizon;
     this.jwtService = jwtService;
+    this.clientFinder = clientFinder;
     this.serverAccountId =
         KeyPair.fromSecretSeed(secretConfig.getSep10SigningSeed()).getAccountId();
     Log.info("Sep10Service initialized.");
   }
 
-  public ChallengeResponse createChallenge(ChallengeRequest request) throws SepException {
+  public ChallengeResponse createChallenge(ChallengeRequest request, String authorization)
```
lijamie98, line 71:
```suggestion
  public ChallengeResponse createChallenge(@NotNull ChallengeRequest request, @Nullable String authorization)
```

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -252,7 +262,85 @@ public void incrementValidationRequestValidatedCounter() {
...
+      }
+
+      return;
+    }
+
+    if (!authorization.startsWith("Bearer")) {
+      throw new SepValidationException("Invalid JWT token");
+    }
+
+    String token = authorization.replace("Bearer", "").stripLeading();
```
lijamie98, line 287: What if we have a the string `Bearer` in the token's payload?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -211,7 +206,21 @@ public <T extends AbstractJwt> T decode(String cipher, Class<T> cls)
     }
   }
 
-  private String toBase64OrNull(String value) {
-    return value == null ? null : Base64.encodeBase64String(value.getBytes(StandardCharsets.UTF_8));
+  @SneakyThrows
+  public Jws<Claims> getHeaderJwt(String signingKey, String cipher) {
+    var factory = KeyFactory.getInstance("Ed25519");
```
lijamie98, line 211: How does this relate to SEP10
[```{
  "alg": "EdDSA"
}```](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request-header)

Is this verified by the JWT library?

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -71,12 +80,20 @@ public ChallengeResponse createChallenge(ChallengeRequest request) throws SepExc
...
     // Validate the validity of the memo
     Memo memo = validateChallengeRequestMemo(request);
+    String clientDomainSigningKey = null;
+    if (!isEmpty(request.getClientDomain())) {
+      debugF("Fetching SIGNING_KEY from client_domain: {}", request.getClientDomain());
+      clientDomainSigningKey = fetchSigningKeyFromClientDomain(request.getClientDomain());
+      debugF("SIGNING_KEY from client_domain fetched: {}", clientDomainSigningKey);
+    }
+    // Check authorization
+    validateAuthorization(request, authorization, clientDomainSigningKey);
```
lijamie98, line 90: This is confusing because the signing key may not be the client domain signing key. 

The code does not clearly separate the case of the `client_domain` vs the `account` verification and can be confusing.

service-runner/src/main/resources/profiles/default/test.env
```diff
@@ -5,7 +5,8 @@ reference.server.url=http://localhost:8091
 observer.server.url=http://localhost:8083
 platform.server.url=http://localhost:8085
 custody.server.url=http://localhost:8086
-wallet.server.url=http://localhost:8092
+# TODO
```
lijamie98, line 8: What's the purpose of `TODO` here?

service-runner/src/main/resources/profiles/default/config.env
```diff
@@ -1,14 +1,14 @@
 # secrets
 secret.data.username=postgres
 secret.data.password=password
-secret.platform_api.auth_secret=myAnchorToPlatformSecret
-secret.callback_api.auth_secret=myPlatformToAnchorSecret
+secret.platform_api.auth_secret=myAnchorToPlatformSecret1234567890
```
lijamie98, line 4: What is the purpose of the trailing `1234567890`?

service-runner/src/main/resources/config/reference-config.yaml
```diff
@@ -33,10 +33,10 @@ auth:
   # CallbackAPI requests (`Platform->Anchor`) will contain an authentication header whose token was built using this
   # secret. The Anchor Backend will also store this same secret and use it to decode the incoming token to verify it
   # came from the Platform.
-  platformToAnchorSecret: myPlatformToAnchorSecret
+  platformToAnchorSecret: myPlatformToAnchorSecret1234567890
```
lijamie98, line 36: What's the purpose of the trailing `1234567890`?

wallet-reference-server/src/main/resources/assets/stellar.toml
```diff
@@ -0,0 +1,5 @@
+ACCOUNTS = [ "GAJ3JPQH3ISAZBMRUPKQ6ZQLRXIROCKJKBIYMGPFUEETQEDGFAW4IQ5H" ]
```
lijamie98, line 1: The location is confusing because this is not part of an asssets file.

.github/workflows/sub_essential_tests.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           SEP1_TOML_VALUE: /home/runner/java-stellar-anchor-sdk/service-runner/src/main/resources/config/stellar.host.docker.internal.toml
           SEP10_WEB_AUTH_DOMAIN: host.docker.internal:8080
           SEP10_HOME_DOMAINS: host.docker.internal:8080,*.stellar.org
+          TOML_PATH: /home/runner/java-stellar-anchor-sdk/wallet-reference-server/src/main/resources/assets
```
lijamie98, line 72: Is this necessary? I don't see other `stellar.toml` for the wallet reference server?

core/src/main/java/org/stellar/anchor/auth/AuthConfig.java
```diff
@@ -8,7 +10,9 @@
 @NoArgsConstructor
 public class AuthConfig {
   AuthType type;
-  String secret;
+  String secretString;
```
lijamie98, line 13: What's the reason to change to `secretString`?

The `String` of the `String secretString` seems redudant.

core/src/main/java/org/stellar/anchor/auth/AuthConfig.java
```diff
@@ -8,7 +10,9 @@
...
+  @Nullable // Only not-null when type is JWT
+  SecretKey secretJwt;
```
lijamie98, line 15: This should map to the auth configuration. We should move this to a helper function to avoid confusion.

core/src/main/java/org/stellar/anchor/auth/JwtService.java
```diff
@@ -42,39 +51,42 @@ public JwtService(SecretConfig secretConfig, CustodySecretConfig custodySecretCo
         secretConfig.getSep10JwtSecretKey(),
         secretConfig.getSep24InteractiveUrlJwtSecret(),
         secretConfig.getSep24MoreInfoUrlJwtSecret(),
-        secretConfig.getCallbackAuthSecret(),
-        secretConfig.getPlatformAuthSecret(),
-        custodySecretConfig.getCustodyAuthSecret());
+        secretConfig.getCallbackAuthSecretKey(),
```
lijamie98, line 54: Can we use helper function to avoid mixing the `secretKey` in the config class?

core/src/main/java/org/stellar/anchor/config/CustodySecretConfig.java
```diff
@@ -1,9 +1,13 @@
...
+import javax.crypto.SecretKey;
+
 public interface CustodySecretConfig {
   String getFireblocksApiKey();
 
   String getFireblocksSecretKey();
 
   String getCustodyAuthSecret();
+
+  SecretKey getCustodyAuthSecretKey();
```
lijamie98, line 12: The `SecretKey` does not belong to the config class. We should move it out to avoid confusion.

core/src/main/java/org/stellar/anchor/config/SecretConfig.java
```diff
@@ -1,20 +1,26 @@
 package org.stellar.anchor.config;
 
+import javax.crypto.SecretKey;
+
 public interface SecretConfig {
-  String getSep6MoreInfoUrlJwtSecret();
+  SecretKey getSep6MoreInfoUrlJwtSecret();
```
lijamie98, line 6: This is a config interfaces and should be mapped to the configuration. Please use a class or helper function to avoid confusion.

core/src/test/kotlin/org/stellar/anchor/filter/Sep10JwtFilterTest.kt
```diff
@@ -21,6 +21,7 @@ import org.stellar.anchor.config.CustodySecretConfig
 import org.stellar.anchor.config.SecretConfig
 import org.stellar.anchor.filter.Sep10JwtFilter.APPLICATION_JSON_VALUE
 import org.stellar.anchor.filter.Sep10JwtFilter.JWT_TOKEN
+import org.stellar.anchor.toSecretKey
```
lijamie98, line 25: Can we add test cases for custodial challenge with bad signatures?

docs/resources/deployment-examples/example-k8s/anchor-platform-dev/reference-server-config-map.yaml
```diff
@@ -23,8 +23,8 @@ data:
 
     integration-auth:
       authType: none
-      platformToAnchorSecret: myPlatformToAnchorSecret
-      anchorToPlatformSecret: myAnchorToPlatformSecret
+      platformToAnchorSecret: myPlatformToAnchorSecret1234567890
```
lijamie98, line 26: Trailing `1234567890`?

essential-tests/build.gradle.kts
```diff
@@ -32,6 +30,8 @@ dependencies {
   testFixturesImplementation(project(":service-runner"))
   testFixturesImplementation(project(":lib-util"))
 
+  testFixturesImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test")
```
lijamie98, line 33: Is this necessary? If yes, can we move it to the `libs.version.toml`?

extended-tests/build.gradle.kts
```diff
@@ -7,15 +7,14 @@ plugins {
...
-repositories { maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") } }
-
 dependencies {
   testImplementation(libs.stellar.wallet.sdk)
   testImplementation(libs.okhttp3.mockserver)
   testImplementation(libs.assertj.core)
   testImplementation("org.springframework.boot:spring-boot-starter-web")
   testImplementation(libs.kotlin.serialization.json)
   testImplementation(libs.ktor.client.json)
+  testImplementation(libs.google.gson)
```
lijamie98, line 17: Is this necessary?

platform/src/test/kotlin/org/stellar/anchor/platform/callback/PlatformIntegrationHelperTest.kt
```diff
@@ -42,7 +43,16 @@ class PlatformIntegrationHelperTest {
...
           )
 
-        val jwtService = JwtService(null, null, null, null, "secret", "secret", "secret")
+        val jwtService =
+          JwtService(
+            null,
+            null,
+            null,
+            null,
+            "ssecretsecretsecretsecretecretsecretsecretsecret".toSecretKey(),
```
lijamie98, line 52: 💡 Is `ssecretsecretsecretsecretecretsecretsecretsecret` necessary? 😄

platform/src/test/kotlin/org/stellar/anchor/platform/config/Sep6ConfigTest.kt
```diff
@@ -28,7 +29,8 @@ class Sep6ConfigTest {
     MockKAnnotations.init(this, relaxUnitFun = true)
     assetService = DefaultAssetService.fromJsonResource("test_assets.json")
     every { custodyConfig.isCustodyIntegrationEnabled } returns true
-    every { secretConfig.sep6MoreInfoUrlJwtSecret } returns "more_info url jwt secret"
+    every { secretConfig.sep6MoreInfoUrlJwtSecret } returns
+      "more_info url jwt secret more_info url jwt secret 12345".toSecretKey()
```
lijamie98, line 33: trailing `12345` necessary?

platform/src/test/kotlin/org/stellar/anchor/platform/service/Sep24MoreInfoUrlConstructorTest.kt
```diff
@@ -41,7 +42,8 @@ class Sep24MoreInfoUrlConstructorTest {
   @BeforeEach
   fun setup() {
     MockKAnnotations.init(this, relaxUnitFun = true)
-    every { secretConfig.sep24MoreInfoUrlJwtSecret } returns "sep24_jwt_secret"
+    every { secretConfig.sep24MoreInfoUrlJwtSecret } returns
+      "sep24_jwt_secret_sep24_jwt_secretsep24_jwt_secret_sep24_jwt_secret".toSecretKey()
```
lijamie98, line 46: Seems excessive. 😸

lijamie98 (changes requested):

platform/src/test/kotlin/org/stellar/anchor/platform/service/Sep24MoreInfoUrlConstructorTest.kt
Ifropc: Key needs to be long

platform/src/test/kotlin/org/stellar/anchor/platform/config/Sep6ConfigTest.kt
Ifropc: Yes, the key needs to be long

platform/src/test/kotlin/org/stellar/anchor/platform/callback/PlatformIntegrationHelperTest.kt
Ifropc: Yes, it needs to be 256 bit

extended-tests/build.gradle.kts
Ifropc: It should already be there, but for some reason the code doesn't compile without this library, do you have any idea why?

essential-tests/build.gradle.kts
Ifropc: No, I forgot that I added it :)

docs/resources/deployment-examples/example-k8s/anchor-platform-dev/reference-server-config-map.yaml
Ifropc: It needs to be 256 bit, all key changes are necessary

core/src/test/kotlin/org/stellar/anchor/filter/Sep10JwtFilterTest.kt
Ifropc: It is in Sep10ServiceTests

.github/workflows/sub_essential_tests.yml
Ifropc: Yes it is. Otherwise it can't read toml the way we package the jar right now

wallet-reference-server/src/main/resources/assets/stellar.toml
Ifropc: What do you mean?

service-runner/src/main/resources/config/reference-config.yaml
Ifropc: Make key seize 256 bit

service-runner/src/main/resources/profiles/default/config.env
Ifropc: Make key seize 256 bit

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
Ifropc: It's always either client domain signing key or null, in which case does it seem to be not a client domain signing key?

core/src/main/java/org/stellar/anchor/auth/JwtService.java
Ifropc: Yes

core/src/main/java/org/stellar/anchor/sep10/Sep10Service.java
```diff
@@ -71,12 +80,20 @@ public ChallengeResponse createChallenge(ChallengeRequest request) throws SepExc
...
     // post validations to be defined by the anchor
     postChallengeRequestValidation(request);
     // increment counter
     incrementChallengeRequestCreatedCounter();
     // Create the challenge
-    return createChallengeResponse(request, memo);
+    return createChallengeResponse(request, memo, clientDomainSigningKey);
```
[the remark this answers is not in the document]
Ifropc: This is the signing key fetched from the client domain, do we want to rename it?

core/src/main/java/org/stellar/anchor/config/CustodySecretConfig.java
Ifropc: The reason I added it here is to validate it the same time it's created. Otherwise it can be easily missed. Plus, it's easier to work with it when it's stored in the config. Do you think we should remove it? Where should we keep validation?

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
```diff
@@ -45,7 +45,15 @@ public FilterRegistrationBean<Filter> platformTokenFilter(PlatformServerConfig c
     String authSecret = config.getSecretConfig().getPlatformAuthSecret();
     switch (config.getAuth().getType()) {
       case JWT:
-        JwtService jwtService = new JwtService(null, null, null, null, null, authSecret, null);
+        JwtService jwtService =
+            new JwtService(
```
philipliu, line 49: Maybe we should start using a builder for `JwtService`. Verifying the arguments are passed in correctly without looking at the constructor is difficult.

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -105,7 +105,7 @@ public void validateNonCustodialClient(ClientConfig clientConfig, Errors errors)
     if (!isEmpty(clientConfig.getSigningKey())) {
       try {
         String clientSigningKey =
-            Sep10Helper.fetchSigningKeyFromClientDomain(clientConfig.getDomain());
+            Sep10Helper.fetchSigningKeyFromClientDomain(clientConfig.getDomain(), false);
```
philipliu, line 108: Question: why is this hardcoded to false? I thought it depended on whether or not we were using Testnet.

extended-tests/build.gradle.kts
Ifropc: https://github.com/stellar/java-stellar-anchor-sdk/actions/runs/8622916329/job/23635221721
Doesn't compile without it, IMO we should brush out all dependencies in 3.0

service-runner/src/main/resources/profiles/host-docker-internal/config.env
```diff
@@ -1,3 +1,4 @@
 sep10.home_domain=host.docker.internal:8080
 sep1.toml.value=/config/stellar.host.docker.internal.toml
-sep24.more_info_url.base_url=http://host.docker.internal:8091/sep24/transaction/more_info
\ No newline at end of file
+sep24.more_info_url.base_url=http://host.docker.internal:8091/sep24/transaction/more_info
+wallet.server.url=http://host.docker.internal:8092
```
philipliu, line 4: Do we still need this? I believe you added this initially to fix the wallet server.

extended-tests/build.gradle.kts
JakeUrban: Lets create a ticket for that, I agree we should do that for 3.0

core/src/main/java/org/stellar/anchor/auth/AuthConfig.java
philipliu: I think it's because we use `SecretKey`s now and need a way to differentiate them.

.github/workflows/sub_essential_tests.yml
```diff
@@ -82,6 +83,14 @@ jobs:
           timeout: "300"
           interval: "1"
 
+      - name: Wait for the wallet server to start and get ready
```
philipliu, line 86: We should probably add this for all servers in another PR.

core/src/main/java/org/stellar/anchor/config/SecretConfig.java
```diff
@@ -1,20 +1,27 @@
...
 
-  String getSep10JwtSecretKey();
+  @Deprecated // TODO: ANCHOR-667
+  SecretKey getSep10JwtSecretKey();
 
   String getSep10SigningSeed();
 
-  String getSep24InteractiveUrlJwtSecret();
+  SecretKey getSep24InteractiveUrlJwtSecret();
```
philipliu, line 13: Question: why do some of these `SecretKey`s end with `key` but others do not?

core/src/test/kotlin/org/stellar/anchor/auth/JwtServiceTest.kt
```diff
@@ -37,13 +34,20 @@ internal class JwtServiceTest {
...
     custodySecretConfig = mockk()
-    every { secretConfig.sep6MoreInfoUrlJwtSecret } returns "jwt_secret"
-    every { secretConfig.sep10JwtSecretKey } returns "jwt_secret"
-    every { secretConfig.sep24InteractiveUrlJwtSecret } returns "jwt_secret"
-    every { secretConfig.sep24MoreInfoUrlJwtSecret } returns "jwt_secret"
-    every { secretConfig.callbackAuthSecret } returns "jwt_secret"
-    every { secretConfig.platformAuthSecret } returns "jwt_secret"
-    every { custodySecretConfig.custodyAuthSecret } returns "jwt_secret"
+    every { secretConfig.sep6MoreInfoUrlJwtSecret } returns
+      "jwt_secrejwt_secrejwt_secrejwt_secrettttjwt_secrejwt_secrejwt_secrejwt_secrejwt_secrettttjwt_secrettjwt_secrejwt_secrejwt_secrejwt_secrettttjwt_secret".toSecretKey()
```
philipliu, line 38: nit: we can move this into a test constants file so we don't have to copy these long strings everywhere.

extended-tests/build.gradle.kts
Ifropc: https://stellarorg.atlassian.net/browse/ANCHOR-668

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
Ifropc: I didn't want to complicate the code for the only use case (where anchor uses optional key verification configuration for non-custodial AND the client server uses http)
I don't think there should be a wallet server that uses http anywhere. This flag was more of a hack to enable wallet server.

docs/resources/deployment-examples/example-k8s/anchor-platform-dev/reference-server-config-map.yaml
```diff
@@ -23,8 +23,8 @@ data:
```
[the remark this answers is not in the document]
Ifropc: If we are not using them then I think we can remove it 🤔 
I just grep-ed the string and replaced it everywhere

service-runner/src/main/resources/profiles/host-docker-internal/config.env
Ifropc: I don't think so

.github/workflows/sub_essential_tests.yml
Ifropc: https://stellarorg.atlassian.net/browse/ANCHOR-669

core/src/main/java/org/stellar/anchor/config/SecretConfig.java
Ifropc: I will remove all added fields and rename old ones back

philipliu: I thought we decided to remove the `SecretKey`s?

Ifropc: Sorry I forgot to push 🤦🏼

philipliu (approved):
LGTM, just a minor comment.

platform/src/main/java/org/stellar/anchor/platform/config/PropertySep10Config.java
```diff
@@ -100,6 +102,8 @@ void validateConfig(Errors errors) {
           "Please set the secret.sep10.jwt_secret or SECRET_SEP10_JWT_SECRET environment variable");
     }
 
+    KeyUtil.validateJWTSecret(secretConfig.getSep10JwtSecretKey());
```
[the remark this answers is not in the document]
Ifropc: Good idea. Added validation and tests.

lijamie98 (approved):
LGTM

PR #1305, opened 2024-03-26
Fix event processing test after requiring security_protocol config

philipliu:
### Description

This adds the missing default config to the test.

### Context

This was logging a `NullPointerException` due to new default configs being added.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

I added a TODO here because an exception was logged but the test didn't fail.


lijamie98 (approved):
LGTM

PR #1306, opened 2024-03-26
[ANCHOR-622] Redefine clients configuration with a schema and a separate file

lijamie98:
### Description

Redefine clients configuration with a schema and a separate file

### Context

For version 3.0.0

### Testing

- `./gradlew test`

### Documentation

TODO: If this pull request adds a feature, describe changes to documentation that has been made

- Attach stellar-docs pull request, documenting new feature
- If it's an urgent feature request, please create a ticket and attach ticket number to this PR


platform/src/main/resources/config/clients-schema.yaml
Ifropc, line 1: Can we add an example of client configuration in this PR?

lijamie98: Is this example ok? We can always add more. 

https://github.com/stellar/java-stellar-anchor-sdk/pull/1306/files#diff-b3b440548352b4a8f629415a7f8acc8a499b5c5a0597746975fe30dc082398eeR1

core/src/main/java/org/stellar/anchor/config/ClientsConfigV3.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    String callbackUrl;
+
+    /**
+     * The public key used to verify the client's identity against the domain's TOML file. If
+     * specified, the service will validate this key against the `SIGNING_KEY` published in the
+     * domain's TOML file. A mismatch or absence of this key in the TOML file results in a
+     * validation error, ensuring security and integrity.
+     */
+    String signingKey;
```
JiahuiWho, line 87: There is also a discussion about deprecating `signing_key` and `domain` for non-custodial client, and adding `signing_key**s**` and `domains`. Do you think we can incorporate the change here? 

Here's the ticket https://stellarorg.atlassian.net/browse/ANCHOR-626

@Ifropc Can you help confirm?

platform/src/main/resources/config/clients-schema.yaml
```diff
@@ -0,0 +1,87 @@
...
+paths: {}
+components:
+  schemas:
+    ClientsConfig:
+      type: object
+      description: |-
+        Configuration for Stellar anchor clients, distinguishing between custodial and non-custodial clients.
+      properties:
+        custodial_clients:
+          type: array
```
JiahuiWho, line 32: nit: List? Let's unify the terminology :)

core/src/main/java/org/stellar/anchor/config/ClientsConfigV3.java
Ifropc: Right, I think the change is for 2.x and then we merge it into 3.x with removing deprecating signing_key and domain in 2.x and removing it in 3.x

lijamie98: I have modified the yaml file.

platform/src/main/resources/config/clients-schema.yaml
lijamie98: We usually use plural for list. Can you help pointing out the inconsistencies?

JiahuiWho: My bad. I mean since it is described as `List` (List of custodial clients. Defaults to an empty list), shall we just use type `List` instead of `Array`?

lijamie98: I do not see there is a type list in OpenAPI spec though. https://swagger.io/docs/specification/data-models/data-types/

Can you help to give me the reference?

PR #1309, opened 2024-03-28
[ANCHOR-605] Parameterize secrets by namespace

philipliu:
### Description

This parameterizes the secret keys by namespace.

### Context

Secrets are prefixed by namespace in SDF's Vault cluster.

### Testing

- `./gradlew test`
- Testing with demo wallet.

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/helm-charts/reference-server/values.yaml
+++ b/helm-charts/reference-server/values.yaml
@@ -1,4 +1,5 @@
 fullName: reference-server
+namespace: default
 
 secretStoreRefName: fake-secret-store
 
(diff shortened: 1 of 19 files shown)


JiahuiWho (approved):

helm-charts/reference-server/values.yaml
```diff
@@ -1,4 +1,5 @@
 fullName: reference-server
+namespace: default
```
lijamie98, line 2: Is this a bug?

helm-charts/sep-service/values.yaml
```diff
@@ -1,4 +1,5 @@
 fullName: anchor-platform
+namespace: default
```
lijamie98, line 2: Is this a bug?

PR #1310, opened 2024-03-29
[ANCHOR-582] Add verification fields to sep12

JiahuiWho:
### Description

- hardcode the verification fields
- fix some tests

### Testing

- `./gradlew test`


philipliu (approved):
LGTM but I think this should be merged into `develop` instead of `v3` right? We should also update the Customer callback API docs to include these fields.

PR #1327, opened 2024-04-16
[Fix] Update fake-secret-store secrets

philipliu:
### Description

This updates the `fake-secret-store` and charts so that Anchor Platform is deployable again after the recent secret changes.

### Context

Currently AP is undeployable since the chart does not configure `SECRET_SEP6_MORE_INFO_URL_JWT_SECRET` and uses weak keys for JWT secrets.

### Testing

- `./gradlew test`
- Testing with demo wallet.

### Documentation

N/A

### Known limitations

N/A


helm-charts/secret-store/templates/secretstore.yaml
```diff
@@ -12,9 +12,10 @@ spec:
             {
               "POSTGRES_USER": "postgres",
               "POSTGRES_PASSWORD": "cdb1ajkMih",
-              "SEP24_INTERACTIVE_URL_JWT_SECRET": "secret",
-              "SEP24_MORE_INFO_URL_JWT_SECRET": "secret",
-              "SEP10_JWT_SECRET": "secret",
+              "SEP24_INTERACTIVE_URL_JWT_SECRET": "c5457e3a349df9002117543efa7e316dd89e666a5ce6f33a0deb13e90f3f1e9d",
```
Ifropc, line 15: How did you generate this secrets? I think it's a great idea to update our docs saying something along the lines "you can generate secrets using XYZ utility" (preferably if it's something UNIX standard so you can use it from CLI on any machine)

Ifropc (approved):

helm-charts/secret-store/templates/secretstore.yaml
philipliu: I used the node `crypto` library but I think we can do something like `openssl rand -base64 32` instead. I'll create a Jira to add this to the docs.

PR #1345, opened 2024-04-24
[ANCHOR-672] Flatten anchor config fields in value file

philipliu:
### Description

All fields are configurable by the `config` struct in the value file. Previously, only the callback API was configurable using this method. 

### Context

We need to be able to configure using a separate Horizon instance. Rather than hardcoding the value mapping like the callback API, this refactors how config management is done.

### Testing

- `./gradlew test`
- Testing with the demo wallet

### Documentation

N/A

### Known limitations

This is a breaking change so the deployed environments will need to be updated.


helm-charts/reference-server/values.yaml
```diff
@@ -9,6 +9,7 @@ container:
 
 config:
   app:
+    horizonEndpoint: https://horizon-testnet.stellar.org
```
lijamie98, line 12: Where is the `passphrase` set for `public` or `testnet`? 

It would be nice to have the consistent pattern with https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/resources/config/anchor-config-default-values.yaml#L34

lijamie98:
Why do we need to keep the empty `sep-service.yaml` file?

helm-charts/reference-server/values.yaml
lijamie98: Oh, I see. It is the way the reference server is configured. I will approve this PR.

lijamie98 (approved):
LGTM

PR #1350, opened 2024-05-06
[ANCHOR-604] Add deployment annotations

philipliu:
### Description

This allows pod-level annotations to be set.

### Context

We need to tell which port and endpoint Prometheus should scrape the metrics from using annotations.

### Testing

- `./gradlew test`
- Verified the manifest is correctly generated using `helm template`

### Documentation

N/A

### Known limitations

N/A


lijamie98 (approved):
LGTM

PR #1353, opened 2024-05-07
Add withdrawal support for notify_refund_pending RPC method

Ifropc:
### Description

Add support for withdrawals for existing `notify_refund_pending` RPC method

### Context

I noticed refunds are not available for `pending_user_transfer_complete` and `pending_external statuses`.

Generally speaking, in the regular flow transaction goes from `pending_user_transfer_start` -> `pending_anchor` -> `completed`
However, when the anchor takes times to process a transaction and it's still in `pending_anchor` , user wants to refund it and can do so.
In the cases where status is `pending_user_transfer_complete` (funds are available for pickup) or `pending_external` (funds are processed by external entity) we want to fist move it to `pending_anchor` to indicate that funds are not available for pickup anymore, (or that bank transfer was cancelled), i.e. indicate that refund has been initiated.
In this case, if we have issues with Stellar transfer or Circle, etc. transaction won't be in status that shows user wrong information. (Especially in `pending_user_transfer_complete` case, where user funds is technically no longer available for pickup anymore, but status doesn't show that)
After that, we can use existing `notify_refund_send` RPC.

### Testing

- `./gradlew test`
- Added new test cases 

### Documentation

Flow charts should be updated (https://stellarorg.atlassian.net/browse/ANCHOR-688)

### Known limitations

N/A

### DIFF
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
@@ -27,7 +27,8 @@ import org.stellar.anchor.platform.AbstractIntegrationTests
 import org.stellar.anchor.platform.TestConfig
 import org.stellar.anchor.util.GsonUtils
 
-@Disabled
+@Disabled // TODO: https://stellarorg.atlassian.net/browse/ANCHOR-693
+// TODO: https://stellarorg.atlassian.net/browse/ANCHOR-694
 class PlatformApiTests : AbstractIntegrationTests(TestConfig()) {
   private val gson = GsonUtils.getInstance()
 
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
@@ -27,8 +27,8 @@ import org.stellar.anchor.platform.AbstractIntegrationTests
 import org.stellar.anchor.platform.TestConfig
 import org.stellar.anchor.util.GsonUtils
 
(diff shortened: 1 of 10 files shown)


platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyRefundPendingHandler.java
```diff
@@ -185,6 +188,10 @@ protected void updateTransactionWithRpcRequest(
       case SEP_6:
         JdbcSep6Transaction txn6 = (JdbcSep6Transaction) txn;
 
+        if (ImmutableSet.of(WITHDRAWAL, WITHDRAWAL_EXCHANGE).contains(Kind.from(txn6.getKind()))) {
```
philipliu, line 191: Can we add a comment here explaining why we don't update the transaction for withdrawals?

philipliu (approved):
The code looks good although we should update the Platform API tests to cover the new flows.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
```diff
@@ -27,7 +27,8 @@ import org.stellar.anchor.platform.AbstractIntegrationTests
 import org.stellar.anchor.platform.TestConfig
 import org.stellar.anchor.util.GsonUtils
 
-@Disabled
+@Disabled // TODO: https://stellarorg.atlassian.net/browse/ANCHOR-693
```
lijamie98, line 30: Can we put a little details here so that we don't need to go back to the ticket?

lijamie98 (approved):
LGTM with some minor comments.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PlatformApiTests.kt
Ifropc: Agree, let's do this little TODOs later on to not forget things that were put away!

PR #1355, opened 2024-05-08
[ANCNHOR-626] Update ClientsConfig with SigningKeys and Domains

JiahuiWho:
### Description

- Deprecate `signing_key` and `domain`, replace with `signing_keys` and `domains`
- Remove `signing_key` field for noncustodial client (usage only found in validation)
- Replace the use of `signing_key` and `domain` with `signing_keys` and `domains` in related function and test cases

### Testing

- `./gradlew test`

### Documentation

Will update in https://developers.stellar.org/network/anchor-platform/admin-guide/sep10

### DIFF
--- a/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
+++ b/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
@@ -1,5 +1,6 @@
 package org.stellar.anchor.client;
 
+import java.util.HashSet;
 import javax.annotation.Nullable;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
@@ -42,7 +43,7 @@ public class ClientFinder {
       throw new SepNotAuthorizedException("Client not found");
     }
 
-    if (clientDomain != null && !isDomainAllowed(client.getDomain())) {
+    if (clientDomain != null && !isClientDomainsAllowed(client)) {
       throw new SepNotAuthorizedException("Client domain not allowed");
     }
     if (!isClientNameAllowed(client.getName())) {
@@ -62,6 +63,11 @@ public class ClientFinder {
     ClientConfig clientByDomain = clientsConfig.getClientConfigByDomain(clientDomain);
     ClientConfig clientByAccount = clientsConfig.getClientConfigBySigningKey(account);
     return clientByDomain != null ? clientByDomain : clientByAccount;
+  }
+
+  private boolean isClientDomainsAllowed(ClientConfig client) {
+    return new HashSet<>(sep10Config.getAllowedClientDomains()).containsAll(client.getDomains())
+        || sep10Config.getAllowedClientDomains().isEmpty();
   }
 
   private boolean isDomainAllowed(String domain) {
--- a/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
+++ b/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
@@ -1,6 +1,5 @@
 package org.stellar.anchor.client;
 
-import java.util.HashSet;
 import javax.annotation.Nullable;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
@@ -38,15 +37,12 @@ public class ClientFinder {
       return client != null ? client.getName() : null;
     }
 
-    // Check if the client domain and name are allowed
+    // Check if the client is allowed
     if (client == null) {
       throw new SepNotAuthorizedException("Client not found");
     }
 
-    if (clientDomain != null && !isClientAllowed(client)) {
-      throw new SepNotAuthorizedException("Client domain not allowed");
-    }
-    if (!isClientNameAllowed(client.getName())) {
+    if (!sep10Config.getAllowedClientNames().contains(client.getName())) {
       throw new SepNotAuthorizedException("Client name not allowed");
     }
 
@@ -63,15 +59,5 @@ public class ClientFinder {
     ClientConfig clientByDomain = clientsConfig.getClientConfigByDomain(clientDomain);
     ClientConfig clientByAccount = clientsConfig.getClientConfigBySigningKey(account);
     return clientByDomain != null ? clientByDomain : clientByAccount;
-  }
-
-  private boolean isClientAllowed(ClientConfig client) {
-    return new HashSet<>(sep10Config.getAllowedClientDomains()).containsAll(client.getDomains())
-        || sep10Config.getAllowedClientDomains().isEmpty();
-  }
-
-  private boolean isClientNameAllowed(String name) {
-    return sep10Config.getAllowedClientNames().contains(name)
-        || sep10Config.getAllowedClientNames().isEmpty();
   }
 }
--- a/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
+++ b/core/src/main/java/org/stellar/anchor/client/ClientFinder.java
@@ -43,7 +43,7 @@ public class ClientFinder {
       throw new SepNotAuthorizedException("Client not found");
     }
 
-    if (clientDomain != null && !isClientDomainsAllowed(client)) {
+    if (clientDomain != null && !isClientAllowed(client)) {
       throw new SepNotAuthorizedException("Client domain not allowed");
     }
     if (!isClientNameAllowed(client.getName())) {
@@ -65,13 +65,8 @@ public class ClientFinder {
     return clientByDomain != null ? clientByDomain : clientByAccount;
   }
 
-  private boolean isClientDomainsAllowed(ClientConfig client) {
+  private boolean isClientAllowed(ClientConfig client) {
     return new HashSet<>(sep10Config.getAllowedClientDomains()).containsAll(client.getDomains())
-        || sep10Config.getAllowedClientDomains().isEmpty();
-  }
-
-  private boolean isDomainAllowed(String domain) {
-    return sep10Config.getAllowedClientDomains().contains(domain)
         || sep10Config.getAllowedClientDomains().isEmpty();
   }
 
--- a/core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
+++ b/core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
@@ -12,9 +12,9 @@ public interface ClientsConfig {
   class ClientConfig {
     String name;
     ClientType type;
-    @Deprecated String signingKey;
+    @Deprecated String signingKey; // ANCHOR-696
     Set<String> signingKeys;
-    @Deprecated String domain;
+    @Deprecated String domain; // ANCHOR-696
     Set<String> domains;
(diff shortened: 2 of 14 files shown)


platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -84,9 +104,11 @@ private void validateClient(ClientConfig clientConfig, Errors errors) {
   }
 
   public void validateCustodialClient(ClientConfig clientConfig, Errors errors) {
-    if (isEmpty(clientConfig.getSigningKey())) {
+    if (isEmpty(clientConfig.getSigningKey())
```
philipliu, line 107: Should we keep the validation until we remove the deprecated fields?

platform/src/test/kotlin/org/stellar/anchor/platform/config/PropertyClientsConfigTest.kt
```diff
@@ -18,12 +18,31 @@ class PropertyClientsConfigTest {
...
+  fun `test postConstruct`() {
+    val config = ClientConfig()
+    config.name = "sampleName"
+    config.signingKey = "sampleSigningKey"
+    config.domain = "sampleDomain"
+
+    configs.clients.add(config)
+    configs.postConstruct()
+
+    Assertions.assertEquals(config.signingKey, config.signingKeys.first())
```
philipliu, line 32: nit: Can we check that there are no other values in `signingKeys` besides the `signingKey`?

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
```diff
@@ -64,6 +65,11 @@ private ClientConfig getClient(String clientDomain, String account) {
     return clientByDomain != null ? clientByDomain : clientByAccount;
   }
 
+  private boolean isClientDomainsAllowed(ClientConfig client) {
```
lijamie98, line 68:
```suggestion
  private boolean isClientAllowed(ClientConfig client) {
```

platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
```diff
@@ -93,11 +93,13 @@ public void start() {
         switch (clientConfig.getType()) {
           case CUSTODIAL:
             processorName =
-                CLIENT_STATUS_CALLBACK_EVENT_PROCESSOR_NAME_PREFIX + clientConfig.getSigningKey();
+                CLIENT_STATUS_CALLBACK_EVENT_PROCESSOR_NAME_PREFIX
+                    + clientConfig.getSigningKeys().stream().findFirst();
```
lijamie98, line 97: If there are multiple signing keys, will this break? Should we use string concatenation instead of the first key?

lijamie98:
Can we add tests that has multiple `signing_keys` and multiple `domains`?

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
JiahuiWho: It's intended to remove `SigningKey` for non-custodial client. Only custodial wallets will need the public_key attribute per discussion.

platform/src/main/java/org/stellar/anchor/platform/event/EventProcessorManager.java
JiahuiWho: This won't break, it will just find the first element. 
But it could returns any element from the Stream since SigningKeys and domains are sets (i.e no encounter order), we'd better just stream all the values

JiahuiWho:
> Can we add tests that has multiple `signing_keys` and multiple `domains`?

Yes I have changed PropertyClientsConfigTest.kt to use `signingKeys` and `domains` for testing, in this way we can test with multiple values

I will add tests for `getClientConfigBySigningKey` and `getClientConfigByDomain` to validate it

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
```diff
@@ -42,7 +43,7 @@ public String getClientName(String clientDomain, String account)
       throw new SepNotAuthorizedException("Client not found");
     }
 
-    if (clientDomain != null && !isDomainAllowed(client.getDomain())) {
+    if (clientDomain != null && !isClientDomainsAllowed(client)) {
```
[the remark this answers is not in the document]
JiahuiWho: You are right, but only if we know the `client` is in `clientAllowList` for sure. I saw the only call site is the Sep10Service, where the account and domain were extracted from request and passed down here, but does it guarantee the client find is allowed?

If this is the case we should also either move or rename the function indicating this is for Sep10 use only, `ClientFinder` is just too general and the validation should be there for other cases.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -596,8 +591,13 @@ String getClientName(String account) throws BadRequestException {
     if (sep10Config.isClientAttributionRequired() && client == null) {
       throw new BadRequestException("Client not found");
     }
-    if (client != null && !sep10Config.getAllowedClientDomains().contains(client.getDomain())) {
-      client = null;
+
+    if (client != null && client.getDomains() != null) {
```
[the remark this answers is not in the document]
JiahuiWho: Not really... if custodial client didn't specify any domain, the following check will return true and client will stay the same
```
boolean hasDomainIntersection =
          new HashSet<>(sep10Config.getAllowedClientDomains()).containsAll(client.getDomains());
```

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -24,14 +24,32 @@ public class PropertyClientsConfig implements ClientsConfig, Validator {
...
+
   @Override
   public ClientsConfig.ClientConfig getClientConfigBySigningKey(String signingKey) {
     if (signingKeyToClientNameMap == null) {
       signingKeyToClientNameMap = Maps.newHashMap();
       clients.forEach(
           clientConfig -> {
-            if (clientConfig.getSigningKey() != null) {
-              signingKeyToClientNameMap.put(clientConfig.getSigningKey(), clientConfig.getName());
+            if (clientConfig.getSigningKeys() != null && !clientConfig.getSigningKeys().isEmpty()) {
```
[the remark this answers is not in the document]
JiahuiWho: I was hesitate to do that since previously we allow non-custodial client to specify signing key and custodial client to have domain, we might want to keep this until complete deprecation of `domain` and `signing key`? Although it shouldn't have any difference functional wise

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -98,24 +120,10 @@ public void validateCustodialClient(ClientConfig clientConfig, Errors errors) {
...
-              "The client.signingKey does not matched any valid registered keys");
-        }
-      } catch (SepException e) {
-        errors.reject(
-            "client-signing-key-toml-read-failure",
-            "SIGNING_KEY not present in 'client_domain' TOML or TOML file does not exist");
-      }
+    if (isEmpty(clientConfig.getDomain())
+        && (clientConfig.getDomains() == null || clientConfig.getDomains().isEmpty())) {
+      errors.reject(
```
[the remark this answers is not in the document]
JiahuiWho: I meant to prioritize `domains` over `domain` if both are specified. Sure we can make it more strict

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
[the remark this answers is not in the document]
Ifropc: Custodial client never specify a domain, so client.getDomains is always null
For a non-custodial domain, I don't understand why we are using getAllowedClientDomains instead of getAllowedClientNames, this is very confusing.
If client name is allowed, it's client domains must be allowed. And if the name is not allowed, then the domains are not allowed as well. This relates to all of my comments.
IMO, if we know the client name, we should use the check if client name is allowed or not.
I assume it's just for backward compatibility, but I don't remember for sure. If it is, we should create a task to clean it up, and annotate org.stellar.anchor.platform.config.PropertySep10Config#clientAllowList as deprecated

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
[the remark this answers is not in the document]
Ifropc: Let's do that

core/src/main/java/org/stellar/anchor/client/ClientFinder.java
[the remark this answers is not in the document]
JiahuiWho: Right because this function is trying to check if the `client` (found through `account` and `domain` extracted from the sep10 request) is allowed, if not then the sep10 request should be reject.

philipliu (approved):

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
```diff
@@ -12,8 +12,10 @@ public interface ClientsConfig {
   class ClientConfig {
     String name;
     ClientType type;
-    String signingKey;
-    String domain;
+    @Deprecated String signingKey; // ANCHOR-696
```
Ifropc, line 15: Nice!!

Ifropc (commented):
LGTM!
If you could, do you mind adding tests to check backward compatibility? (i.e. checking that when `domain` is defined, `domains` is automatically set to set of one domain and same for custodial key)
Just to ensure backward compatibility is covered by the test

Ifropc (approved):

lijamie98 (approved):
LGTM

JiahuiWho:
> LGTM! If you could, do you mind adding tests to check backward compatibility? (i.e. checking that when `domain` is defined, `domains` is automatically set to set of one domain and same for custodial key) Just to ensure backward compatibility is covered by the test

Yeah test for `postConstruct()` is added to `PropertyClientsConfigTest`

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -596,9 +591,9 @@ String getClientName(String account) throws BadRequestException {
     if (sep10Config.isClientAttributionRequired() && client == null) {
       throw new BadRequestException("Client not found");
     }
-    if (client != null && !sep10Config.getAllowedClientDomains().contains(client.getDomain())) {
+    System.out.println(Arrays.toString(sep10Config.getAllowedClientNames().toArray()));
```
Ifropc, line 594: sout should be removed @JiahuiWho

PR #1366, opened 2024-05-16
[ANCHOR-698] Enable PaymentObservingAccountsManager for SEP-31 only

JiahuiWho:
### Description

This PR add check to `ObservingAccountsBeans ` to check if SEP-31 is enabled. In this case the bean is always created, but only activate if SEP-31 is enabled.

This is because this bean is required in the presence for the creation of `PaymentObserverBeans`. We can't simply use conditional annotation that disable the bean's creation

### Context

The PaymentObseringAccountManager should only be restarted when needed. Currently only SEP-31 requires the service.

### Testing

- `./gradlew test`

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/component/share/ObservingAccountsBeans.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/component/share/ObservingAccountsBeans.java
@@ -2,14 +2,28 @@ package org.stellar.anchor.platform.component.share;
 
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
 import org.stellar.anchor.platform.observer.stellar.PaymentObservingAccountStore;
 import org.stellar.anchor.platform.observer.stellar.PaymentObservingAccountsManager;
 
 @Configuration
 public class ObservingAccountsBeans {
+  private final Environment env;
+
+  public ObservingAccountsBeans(Environment env) {
+    this.env = env;
+  }
+
   @Bean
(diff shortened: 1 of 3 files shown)


platform/src/main/java/org/stellar/anchor/platform/component/share/ObservingAccountsBeans.java
```diff
@@ -2,14 +2,28 @@
...
 
 @Configuration
 public class ObservingAccountsBeans {
+  private final Environment env;
+
+  public ObservingAccountsBeans(Environment env) {
+    this.env = env;
+  }
+
   @Bean
```
philipliu, line 17: Does `@ConditionalOnAnySepsEnabled(seps = {"sep31"})` work? We use this annotation to conditionally create other dependencies.

JiahuiWho: The issue is we need this bean even when sep31 is disabled because `PaymentObserverBeans ` needs it. So that leads to the choice that we always create this bean but not activate it

philipliu: When SEP-31 is disabled, the account the payment observer is monitoring is the distribution account. That is why the eviction task does not need to be enabled when SEP-31 is disabled because it will never be evicted. Is this correct?

JiahuiWho: That's right

lijamie98 (approved):
LGTM. 

This works great!

philipliu (approved):

PR #1367, opened 2024-05-16
[ANCHOR-687] Upgrade to JDK-17

lijamie98:
### Description

- Upgrade JDK from 11 to 17.
- Fix Jwt Json serialization

### Context

- JDK 11 is expected to sunset in 10/2024. 

### Testing

- `./gradlew test`

### Documentation

Update `How to Contribute` doc.

### Known limitations

N/A


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=gradle:7.6.4-jdk11-alpine
+ARG BASE_IMAGE=gradle:7.6.4-jdk17-alpine
```
Ifropc, line 1: We should upgrade gradle to latest in 3.0 if we haven't already

core/src/main/java/org/stellar/anchor/util/Log.java
```diff
@@ -81,7 +83,19 @@ public static void error(final String msg) {
    * @param detail The additional object to be logged.
    */
   public static void error(final String message, final Object detail) {
-    logMessageWithJson(message, detail, getLogger()::error);
+    if (detail instanceof Exception) {
+      Exception ex = (Exception) detail;
+
+      logMessageWithJson(
```
Ifropc, line 89: Why is this required now? Does the message get collapsed?

Ifropc (approved):
LGTM

core/src/main/java/org/stellar/anchor/util/Log.java
lijamie98: This is because `logMessageWithJson(message, detail, getLogger()::error);` failed serializing the `message` field if `detail` is an exception. 

That's the reason that we have expand it manually.

Dockerfile
lijamie98: I will try to make least impact for `2.x`. We can do major upgrades in v3.

Ifropc: Agree!

philipliu (approved):

PR #1384, opened 2024-06-11
[ANCHOR-622][1/n] Create new clients config with service

JiahuiWho:
### Description
The goal is to migrate current ClientsConfig (defined inline) to a separate file like assets do.

###  This PR does:
1. Mark all existing ClientsConfig usage with suffix "_DEPRECATED" for distinguish
2. Define the new `ClientsConfig`, including data schema and validation 
3. Implement `ClientService`, support reading from yaml and json file, as well as in-line config

Take a look at code change under core/src/main/java/org/stellar/anchor/client :)

###  What's next:
Migrate the usage of original ClientsConfig SEP by SEP, it will be done in separate PRs

### Testing
- `./gradlew test`


core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+  public static DefaultClientService fromFile(String filePath) throws InvalidConfigException {
+    Map<String, Object> map = parseFileToMap(filePath);
+    return createDCSFromMap(map);
+  }
+
+  private static Map<String, Object> parseFileToMap(String filePath) throws InvalidConfigException {
+    try {
+      String fileContent = FileUtil.read(Path.of(filePath));
+      return switch (FilenameUtils.getExtension(filePath).toLowerCase()) {
+        case "json" -> gson.fromJson(fileContent, Map.class);
```
JakeUrban, line 41: Why are we supporting JSON?

core/src/main/java/org/stellar/anchor/client/CustodialClientConfig.java
```diff
@@ -0,0 +1,47 @@
...
+   * The unique name identifying the client. This should be a recognizable name that represents the
+   * client entity.
+   */
+  @NonNull String name;
+
+  /**
+   * The public key used for SEP-10 authentication. If a SEP-10 request is signed with this key, the
+   * service will activate the callback if the callback_url is provided.
+   */
+  String signingKey; // ANCHOR-696
```
philipliu, line 23: nit: Can we add `@Deprecated`?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  private static DefaultClientService createDCSFromMap(Map<String, Object> map) {
+    DefaultClientService clientService = new DefaultClientService();
+    clientService.custodialClients =
+        gson.fromJson(gson.toJson(map.get("custodial_clients")), List.class);
+    clientService.nonCustodialClients =
+        gson.fromJson(gson.toJson(map.get("noncustodial_clients")), List.class);
+    return clientService;
+  }
```
JakeUrban, line 59: So is the file expected to have `custodial_clients` and `noncustodial_clients` attributes? This is not the formal of our current clients config. Why are we changing the format? If we want to use a different interface internally that is fine, but I don't see a reason to change the file schema

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+      throws InvalidConfigException {
+    return switch (clientsConfig.getType()) {
+      case FILE -> fromFile(clientsConfig.getValue());
+      case URL -> throw new InvalidConfigException("URL is not supported yet.");
+      default -> throw new InvalidConfigException(
+          String.format("clients.type:%s is not supported.", clientsConfig.getType()));
+    };
+  }
+
+  public static DefaultClientService fromFile(String filePath) throws InvalidConfigException {
```
philipliu, line 32: nit: This can be private.

core/src/main/java/org/stellar/anchor/client/CustodialClientConfig.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /** The public keys used for SEP-10 authentication. */
+  Set<String> signingKeys;
```
JakeUrban, line 26: I don't think we need both of these

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+      String fileContent = FileUtil.read(Path.of(filePath));
+      return switch (FilenameUtils.getExtension(filePath).toLowerCase()) {
+        case "json" -> gson.fromJson(fileContent, Map.class);
+        case "yaml", "yml" -> new Yaml().load(fileContent);
+        default -> throw new InvalidConfigException(
+            String.format("%s is not a supported file format", filePath));
+      };
+    } catch (Exception ex) {
+      throw new InvalidConfigException(
+          List.of(String.format("Cannot read from clients file: %s", filePath)), ex);
```
philipliu, line 48: Should the message be in a list?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+import org.stellar.anchor.util.FileUtil;
+import org.stellar.anchor.util.GsonUtils;
+import org.yaml.snakeyaml.Yaml;
+
+public class DefaultClientService implements ClientService {
+  static final Gson gson = GsonUtils.getInstance();
+  List<CustodialClientConfig> custodialClients;
+  List<NonCustodialClientConfig> nonCustodialClients;
+
+  public static DefaultClientService fromClientsConfig(ClientsConfig clientsConfig)
```
philipliu, line 22: Can we add Javadocs for public methods?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -0,0 +1,86 @@
...
+      throws InvalidConfigException {
+    return switch (clientsConfig.getType()) {
+      case FILE -> fromFile(clientsConfig.getValue());
+      case URL -> throw new InvalidConfigException("URL is not supported yet.");
+      default -> throw new InvalidConfigException(
+          String.format("clients.type:%s is not supported.", clientsConfig.getType()));
+    };
```
JakeUrban, line 29: We need to continue supporting in-line clients config. Ideally businesses wouldn't have to change their config files at all and continue using the `clients:` top-level attribute

core/src/main/java/org/stellar/anchor/client/NonCustodialClientConfig.java
```diff
@@ -0,0 +1,33 @@
...
+  @Nonnull String name;
+
+  /**
+   * The domain associated with the client, used for verifying the client's identity. If a SEP-10
+   * request's domain matches, the callback will be activated if callback_url is defined.
+   */
+  String domain; // ANCHOR-696
+
+  /** The domains associated with the client, used for verifying the client's identity. */
+  Set<String> domains;
```
JakeUrban, line 26: Do we need both of these?

core/src/main/java/org/stellar/anchor/client/NonCustodialClientConfig.java
```diff
@@ -0,0 +1,33 @@
...
+   * The unique name identifying the client. This should be a recognizable name that represents the
+   * client entity.
+   */
+  @Nonnull String name;
+
+  /**
+   * The domain associated with the client, used for verifying the client's identity. If a SEP-10
+   * request's domain matches, the callback will be activated if callback_url is defined.
+   */
+  String domain; // ANCHOR-696
```
philipliu, line 23: nit: Can we add `@Deprecated`?

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
```diff
@@ -1,32 +1,20 @@
...
-  ClientConfig getClientConfigByDomain(String domain);
+  enum ClientsConfigType {
+    @SerializedName("json")
+    JSON,
+    @SerializedName("yaml")
+    YAML,
+    @SerializedName("file")
+    FILE,
+    @SerializedName("url")
+    URL;
```
JakeUrban, line 18: Has anyone ever asked for fetching config via URL?

core/src/test/kotlin/org/stellar/anchor/client/DefaultClientServiceTest.kt
```diff
@@ -0,0 +1,25 @@
...
+  fun `test clients loading and listing`(filename: String) {
+    lateinit var dcs: DefaultClientService
+    when (FilenameUtils.getExtension(filename)) {
+      "yaml" -> dcs = DefaultClientService.fromYamlResourceFile(filename)
+      "json" -> dcs = DefaultClientService.fromJsonResourceFile(filename)
+    }
+
+    // check listing function.
+    val clients = dcs.listAllClients()
+    Assertions.assertEquals(3, clients.size)
```
philipliu, line 23: Can we assert the contents too?

platform/src/main/java/org/stellar/anchor/platform/component/share/ClientsBeans.java
```diff
@@ -3,13 +3,29 @@
...
-  PropertyClientsConfig clientsConfig() {
+  PropertyClientsConfig_DEPRECATED clientsConfigDeprecated() {
+    return new PropertyClientsConfig_DEPRECATED();
+  }
+
+  @Bean
+  @ConfigurationProperties(prefix = "clients")
+  ClientsConfig clientsConfig() {
     return new PropertyClientsConfig();
   }
```
JakeUrban, line 25: One thing that annoys me about the asset config is that specifying it in-line looks redundant: 

```yaml
assets:
  type: yaml
  value:
    assets:
       - ...
```

Ideally, we make `type` optional and if it is not present, assume its yaml. So it would look like this:

```yaml
assets:
  - ...
```

or

```yaml
assets:
  type: file
  value: /path/to/clients.yaml
```

platform/src/main/java/org/stellar/anchor/platform/component/share/ClientsBeans.java
```diff
@@ -3,13 +3,29 @@
...
 public class ClientsBeans {
   @Bean
   @ConfigurationProperties(prefix = "")
-  PropertyClientsConfig clientsConfig() {
+  PropertyClientsConfig_DEPRECATED clientsConfigDeprecated() {
+    return new PropertyClientsConfig_DEPRECATED();
+  }
+
+  @Bean
+  @ConfigurationProperties(prefix = "clients")
```
philipliu, line 22: What happens if the new client config is not provided? Does Anchor Platform still start?

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -1,156 +1,60 @@
...
   public void validate(@NotNull Object target, @NotNull Errors errors) {
-    PropertyClientsConfig configs = (PropertyClientsConfig) target;
-    configs.clients.forEach(clientConfig -> validateClient(clientConfig, errors));
-  }
-
-  private void validateClient(ClientConfig clientConfig, Errors errors) {
-    debugF("Validating client {}", clientConfig);
-    if (isEmpty(clientConfig.getName())) {
-      errors.reject("empty-client-name", "The client.name cannot be empty and must be defined");
+    // TODO(jiahui): ANCHOR-622 Support more types of clients config
```
philipliu, line 25: What other types do we plan to support beyond `json` and `yaml`?

JakeUrban (commented):
I see the value in enabling businesses to specify clients in a separate file, particularly when they're using environment variables for all other config.

However, its not clear to me why we're changing the file format. We can use a different interface internally if we want without impacting the business by mapping the file format to the java object schema.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -716,6 +671,19 @@ assets:
...
 
+##############################
+# Anchor Platform Clients Configuration
+##############################
+# The list of clients for the Anchor server to safely communicate with the outside wallet servers or clients.
+# The list is empty if not defined.
+clients:
+  # `type` can be one of the following.
+  #     `file`: value field contains the path to the assets file. The file name can be *.json or *.yaml.
+  #     `url`: value contains the url to the assets file. The file name can be *.json or *.yaml.
```
philipliu, line 682: Also inline `yaml` and `json`?

service-runner/src/main/resources/config/clients.yaml
```diff
@@ -0,0 +1,9 @@
+custodial_clients:
```
philipliu, line 1: nit: `_clients` is kinda redundant here. What do you think?

core/src/main/java/org/stellar/anchor/client/CustodialClientConfig.java
philipliu: Since this is being merged into `3.0`, we can remove it, right?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: This was based on Jamie's initial change https://github.com/stellar/java-stellar-anchor-sdk/pull/1306, and assume we want the same level of support for both clients and assets? 

If it's not necessary we can remove it.

core/src/main/java/org/stellar/anchor/client/NonCustodialClientConfig.java
JiahuiWho: Will remove the domain, currently it's only deprecated, we haven't get rid of it yet

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: I thought this is why we are doing it for 3.0 because it's a breaking change? We do want to replace the in-line config with the separate file

platform/src/main/java/org/stellar/anchor/platform/component/share/ClientsBeans.java
JiahuiWho: I think this is the new setup, you only need to define like this:
```
clients.type=file
clients.value=/config/clients.yaml
```

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
JiahuiWho: This was a reminder for myself during implementation, will remove

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: What should we do if businesses happened to have ClientsConfig specified both in-line and in a separate file? Should we prioritize one over the other?

Assets previously had an in-line definition feature, but it is not being utilized. Since ClientsConfig is being adopted by the business, shall we encourage them to use the separate file instead?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: Yeah this InvalidConfigException implementation takes a list of messages ..

core/src/main/java/org/stellar/anchor/client/NonCustodialClientConfig.java
JiahuiWho: will remove

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JakeUrban: I personally think we should remove json in 3.0, but I don't feel strongly about it.

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JakeUrban: We want to support both in-line and a separate file. In-line still makes sense if you're not using environment variables, since you can have all your config in one place.

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: The new format is more clear in terms of specifying which client should have what fields defined and is easier to validate. Since the ClientsConfig is new and I don't think businesses have started using it yet, this is an opportunity to establish the right configuration from the outset.

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: Just to confirm with you that by in-line you mean it can be defined in the default config as 
```
clients:
  - type:
  - value:
```

If so that's already included in this PR

JakeUrban: When I say inline, I mean specifying the clients list in the same file, not a separate one. This is supported for assets and we should support it for clients.

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JakeUrban: Ok, I don't feel too strongly about this, but can we update format to avoid 2 top-level keys?

Maybe:

```yaml
clients:
  noncustodial:
    - ...
  custodial:
    - ...
```

platform/src/main/java/org/stellar/anchor/platform/component/share/ClientsBeans.java
JiahuiWho: It should since ClientsConfig is not a required config.
If it's not provided we will just assume it's an empty config

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: I don't think assets supports the exact format you described. For assets it's either defined through file reference or yaml/json string 
```
assets:
  - type: file
  - value: file:assets.yaml
```
or
```
assets
  - type: yaml
  - value: "assets in yaml string not yaml format"
```

JakeUrban: I just tested and you're right, but this is a surprise to me. I thought the `yaml` type meant you could use yaml in the same file. I don't think anyone is going to provide a yaml-formatted or json-formatted string.

I think this is something we should support. It makes sense to support providing a file path to a yaml file when you're configuring everything else via envvars, but if you are using yaml-based config already, it doesn't make sense for the anchor platform to force you to put your assets (or clients) config in a different file. 

@philipliu @reecexlm @Ifropc what are your thoughts on this?

JiahuiWho: I agree. It seems counterintuitive to force user to place your assets (or clients) configuration in a different file. 

> I don't think anyone is going to provide a yaml-formatted or json-formatted string
That's what I thought. This in-line support for assets is not being utilized at all. We might just drop it.

core/src/main/java/org/stellar/anchor/client/CustodialClientConfig.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  @SerializedName("allow_any_destination")
+  boolean allowAnyDestination = false;
+
+  /**
+   * A list of destination accounts that the client is allowed to send assets to. If
+   * allow_any_destination is set to false, the destination account must be in this list.
+   */
+  @SerializedName("destination_accounts")
+  Set<String> destinationAccounts;
```
JakeUrban, line 45: I think the comments here are incorrect. These config attributes control which accounts the clients can use to _receive_ deposits from the anchor platform, right?

Do we need `allowAnySource` and `sourceAccounts` attributes as well? So that the anchor platform only acknowledges payments made from whitelisted accounts?

I'm not familiar with the requirements here. @Ifropc do you have more context on why we added these attributes?

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -716,6 +671,45 @@ assets:
...
+# You can either use a file to provide the configuration or define it inline.
+# Use only one of the following combinations:
+clients:
+  # Option 1: Use a file to provide the configuration.
+  # Specify the type as 'file' and provide the file path in the value field.
+  # The file can only be in .yaml or .json format.
+  # The file content should follow the format used in Option 2.
+  #
+  # type: file
+  # value: path/to/file.yaml
```
JakeUrban, line 688: Any reason not to remove `type` and `value` and replace it with `path`? Since `file` is the only valid `type`?

JakeUrban: This would be a breaking change I guess, so maybe we make it in 3.0?

core/src/main/java/org/stellar/anchor/client/CustodialClientConfig.java
JiahuiWho: You are right it's used for the deposit.

platform/src/main/resources/config/anchor-config-default-values.yaml
JiahuiWho: > Any reason not to remove `type` and `value` and replace it with `path`? Since `file` is the only valid `type`?
I thought about it, but taking a look at the whole config, `type` and `value` are always used in pair to reference files(e.g. sep1.toml, assets). I simply prefer to keep it consistent


> This would be a breaking change I guess, so maybe we make it in 3.0?
Yes this is for 3.0

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
philipliu: Agree that we should do this. I think if we are going to allow inline client configuration, we should also allow inline asset configuration at the same time. This seems like an additive change, so let's implement it later to not block this PR.

philipliu (approved):

core/src/main/java/org/stellar/anchor/config/ClientsConfig_DEPRECATED.java
```diff
@@ -0,0 +1,32 @@
+package org.stellar.anchor.config;
+
+import java.util.Set;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+public interface ClientsConfig_DEPRECATED {
```
Ifropc, line 8: Why do we keep deprecated config? This change goes to v3 we can just remove it

PR #1390, opened 2024-06-17
Add on_hold status support

Ifropc:
### Description

Added new RPC request (`notify_transaction_on_hold`) and handler for it

### Context

Add on_hold status support (introduced in https://github.com/stellar/stellar-protocol/pull/1479)

### Testing

- `./gradlew test`
- Add test for new handler

### Documentation

TODO: add documentation for this request + update diagrams (https://stellarorg.atlassian.net/browse/ANCHOR-679 and https://stellarorg.atlassian.net/browse/ANCHOR-688)

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyTransactionOnHoldHandler.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyTransactionOnHoldHandler.java
@@ -1,17 +1,18 @@
 package org.stellar.anchor.platform.rpc;
 
 import static java.util.Collections.emptySet;
-import static org.stellar.anchor.api.platform.PlatformTransactionData.Kind.WITHDRAWAL;
-import static org.stellar.anchor.api.platform.PlatformTransactionData.Kind.WITHDRAWAL_EXCHANGE;
-import static org.stellar.anchor.api.rpc.method.RpcMethod.NOTIFY_OFFCHAIN_FUNDS_AVAILABLE;
+import static org.stellar.anchor.api.platform.PlatformTransactionData.Kind.*;
+import static org.stellar.anchor.api.platform.PlatformTransactionData.Sep.SEP_24;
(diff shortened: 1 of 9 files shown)


platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyTransactionOnHoldHandler.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  protected Set<SepTransactionStatus> getSupportedStatuses(JdbcSepTransaction txn) {
+    if (Set.of(SEP_6, SEP_24).contains(Sep.from(txn.getProtocol()))) {
+      Set<SepTransactionStatus> supportedStatuses = new HashSet<>();
+      supportedStatuses.add(PENDING_USR_TRANSFER_START);
+      Sep from = Sep.from(txn.getProtocol());
+      if (from == SEP_6) {
+        JdbcSep6Transaction txn6 = (JdbcSep6Transaction) txn;
+        if (ImmutableSet.of(WITHDRAWAL, WITHDRAWAL_EXCHANGE)
+            .contains(PlatformTransactionData.Kind.from((txn6).getKind()))) {
```
philipliu, line 71: Do we also need to check `areFundsReceived`? Same for SEP-24.

Ifropc: Good catch, I think it was just copy-paste, we don't need it

Ifropc: I actually don't think so, I think in MGI flow it can actually go straight from pending_user_transfer_start
I updated notify_offchain_fuds_availble instead, so if transferReceivedAt is not updated yet, it would be there

philipliu (approved):

PR #1409, opened 2024-07-03
[ANCHOR-731] Only check send limits for SEP-31

philipliu:
### Description

This disables send limit checks when the SEP-38 context is not SEP-31.

### Context

This is to unblock the SEP-6 exchange flows.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java
@@ -201,17 +201,19 @@ public class Sep38Service {
     }
 
     // Check SEP31 send limits
-    if (sellAsset == null) {
-      throw new BadRequestException("Unsupported sell asset.");
-    }
-    String[] assetCode = sellAsset.getAsset().split(":");
-    AssetInfo asset = assetService.getAsset(assetCode[1]);
-    Long sendMinLimit = asset.getSend().getMinAmount();
-    Long sendMaxLimit = asset.getSend().getMaxAmount();
+    if (context == SEP31) {
+      if (sellAsset == null) {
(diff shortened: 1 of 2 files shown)


JakeUrban (commented):
We have deposit and withdraw min/max limits that should be checked here if `context` is `sep6`, can we implement that?

JakeUrban:
Alternatively, we can remove limit checks for both SEP-31 and SEP-6 in the SEP-38 service and enforce those limits in the transaction initiation requests. That seems more straightforward.

philipliu:
Checking deposit and withdrawal limits in SEP-38 is difficult since only the SEP is passed through context. Given a SEP-6 quote, we wouldn't know whether it's for deposit or withdrawal. Limits are already enforced at transaction creation time in both SEP-31 and SEP-6, so I'm in favor of removing the check from SEP-31 in version 3.0 since this is a breaking change.

- SEP-31: https://github.com/stellar/java-stellar-anchor-sdk/blob/f46d75aeea8b69e3f02c709d145f9cc5b46ed28f/core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java#L149-L153
- SEP-6 Deposit: https://github.com/stellar/java-stellar-anchor-sdk/blob/9b3503f200f659cfa818f2e0560ba547b8d6be2e/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java#L188-L193
- SEP-6 Withdrawal: https://github.com/stellar/java-stellar-anchor-sdk/blob/9b3503f200f659cfa818f2e0560ba547b8d6be2e/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java#L358-L363

JakeUrban (approved):

PR #1435, opened 2024-07-23
[ANCHOR-451] Merge `send` into `sep31` in assets config

JiahuiWho:
### Description

- remove the `send` field in assets config 
- Merge all fields under `send` into `sep31`

### Context

There is no reason to have separate object for sep31 and send, as they describe the same service.

### Testing

- `./gradlew test` All tests should pass

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java
@@ -50,7 +50,7 @@ public class AssetInfo {
 
   DepositOperation deposit;
   WithdrawOperation withdraw;
-  SendOperation send;
+  @Deprecated SendOperation send;
   Sep31Operation sep31;
   Sep38Operation sep38;
 
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java
@@ -50,7 +50,6 @@ public class AssetInfo {
 
   DepositOperation deposit;
   WithdrawOperation withdraw;
-  @Deprecated SendOperation send;
   Sep31Operation sep31;
   Sep38Operation sep38;
 
@@ -106,21 +105,6 @@ public class AssetInfo {
   @Data
   public static class WithdrawOperation extends AssetOperation {
     List<String> methods;
-  }
-
-  @Data
-  public static class SendOperation {
-    @SerializedName("fee_fixed")
-    Integer feeFixed;
-
-    @SerializedName("fee_percent")
-    Integer feePercent;
-
-    @SerializedName("min_amount")
-    Long minAmount;
-
-    @SerializedName("max_amount")
-    Long maxAmount;
   }
 
   @Data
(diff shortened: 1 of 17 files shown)


service-runner/src/main/resources/config/assets.yaml
```diff
@@ -19,11 +19,14 @@ assets:
       methods:
         - bank_account
         - cash
-    send:
+    send: # (Deprecated) All fields should be incorporated into `sep31`.
       fee_fixed: 0
       fee_percent: 0
       max_amount: 1000000
     sep31:
```
philipliu, line 26: Is it better to merge `sep31` into `send` instead? That way, we have `deposit/withdraw/send` instead of `deposit/withdraw/sep31`.

api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java
```diff
@@ -50,7 +50,7 @@ public static String makeSep11AssetName(String assetCode, String assetIssuer) {
 
   DepositOperation deposit;
   WithdrawOperation withdraw;
-  SendOperation send;
+  @Deprecated SendOperation send;
```
philipliu, line 53: Since we are merging this change into 3.0, should we be removing it instead of deprecating it?

core/src/main/java/org/stellar/anchor/util/AssetHelper.java
```diff
@@ -69,4 +70,32 @@ public static String getAssetCode(String asset) {
   public static String getAssetIssuer(String asset) {
     return asset.split(":").length >= 3 ? asset.split(":")[2] : null;
   }
+
+  /**
+   * The following functions are for merging SendOperation and Sep31Operation only. They will be
+   * removed once the migration is complete.
+   */
+  public static Integer getSep31SendFeeFixed(AssetInfo asset) {
```
philipliu, line 78: Nice

service-runner/src/main/resources/config/assets.yaml
JiahuiWho: I think `sep31` is a more intuitive choice since I have been working on SEP-12 and SEP-31 doc for while ( :stuck_out_tongue: ) and `sep31` is a more commonly referenced term in out doc.

philipliu: Hmm... I feel like we should be updating the docs to use `send` instead. What are your thoughts on this @Ifropc?

Ifropc: Hm I agree that SEP-31 is for me personally more clear than `send`, but I like the consistency of deposit/withdraw. Maybe we should support both?

Ifropc: I think `sep31` is a bit more clear, including the code. I'd also add `sep24` in withdrawal/deposit too, so it's `sep24.deposit` `sep24.withdrawal` for consistency. WDYT?

JiahuiWho: Isn't deposit and withdraw currently shared by sep6 and sep24? Do we need to separate them?

Ifropc:
```
sep31:
   ...
sep24:
   withdraw:
      ...
    deposit:
      ...
sep6:
   withdraw:
      ...
    deposit:
      ...
```

JiahuiWho:
Discussed offline to use a structure such as "SEP -> deposit/withdraw/send ->" to maintain clear and consistent configuration.

Ifropc (approved):

philipliu (approved):

Issue #1444, opened 2024-08-01
[ANCHOR-746] Add SEP-12 customer callback support


Ifropc:
Background: 

Following SEP-6 flow update, the KYC gathering logic is now the following:

Once SEP-6 transaction is in the `pending_customer_info_update `, wallet should call SEP-12 GET `/customer?transaction_id=<id>` to gather all KYC fields that is expected to be provided. 

Wallet then calls PUT `/customer?transaction_id=<id>` to update KYC. GET/PUT are called in sequence until KYC reaches ACCEPTED/REJECTED status. This works well until PROCESSING is added to the flow. In that case wallet needs to continuously pull GET /customer to check if status has been updated. This is not ideal considering that this call would need to be made from the user device for non-custodial wallets.   

Instead, we should add SEP-12 customer callback that notifies wallet backend about necessary change. We should update [notify_customer_info_updated](https://developers.stellar.org/platforms/anchor-platform/api-reference/rpc/methods/notify_customer_info_updated) RPC. When called, it should fetch latest KYC using callback to business server ([get /customer](https://developers.stellar.org/platforms/anchor-platform/api-reference/callbacks/get-customer)) and:

Update transaction status to `pending_customer_info_update` (`NEEDS_INFO`); `pending_anchor` (`PROCESSING`); `pending_user_transfer_start` (`ACCEPTED`); `error` (`REJECTED`)

We should add a new callback to send customer information. We can use <root>/sep12 path in v2 and organize it better in v3      

[Slack discussion thread](https://stellarfoundation.slack.com/archives/C036T11AY84/p1722442800768409)
Internal issue: [ANCHOR-746](https://stellarorg.atlassian.net/browse/ANCHOR-746)

[ANCHOR-746]: https://stellarorg.atlassian.net/browse/ANCHOR-746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

philipliu:
Rather than calling the customer callback in NotifyCustomerInfoUpdated. What do you think of introducing a new parameter called status (SEP-12 status)? The business server should have access to the customer status and we can avoid new failure mode in the RPC this way.

Ifropc:
We could do that, but the idea is that in a callback we need to pass the the formed KYC response in the callback, back to the wallet. So we need to make a call to BS either way

[...]

PR #1445, opened 2024-08-01
[ANCHOR-764] Update NotifyCustomerInfoUpdated RPC

philipliu:
### Description

Addresses https://github.com/stellar/java-stellar-anchor-sdk/issues/1444.

This change:

- Updates the `NotifyCustomerInfoUpdated` RPC so that it updates the transaction based on the requested Customer's status and sends a Customer updated event. `customerId` and `type` are new parameters needed to distinguish whether a customer is a SEP-31 receiver, sender, or a SEP-6 customer.
- Updates SEP-12 service such that it passes the full SEP-12 customer into the `AnchorEvent`. Previously only the customer ID was present in the event. This is not a breaking change since the SEP-12 customer object is serializer backward compatible since it also contains an ID.
- Sends a SEP-12 status callback to all clients.

Callback configuration and end-to-end tests will be added in another PR.

### Context

N/A

### Testing

- `./gradlew test`

### Documentation

stellar-docs PR will be created

### Known limitations

N/A


platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyCustomerInfoUpdatedHandler.java
```diff
@@ -59,15 +60,29 @@ public RpcMethod getRpcMethod() {
...
+      }
+    }
     return PENDING_RECEIVER;
   }
 
   @Override
   protected Set<SepTransactionStatus> getSupportedStatuses(JdbcSepTransaction txn) {
-    if (SEP_31 == Sep.from(txn.getProtocol())) {
-      return Set.of(PENDING_CUSTOMER_INFO_UPDATE);
+    switch (Sep.from(txn.getProtocol())) {
```
Ifropc, line 78: In which status we start before we call the notify_customer_info_updated first time? Is it PENDING_CUSTOMER_INFO_UPDATE?

philipliu: It should be according to the SEP-31 status transition diagram.

Ifropc: Oh sorry asking for SEP-6

philipliu: For SEP-6, it's also PENDING_CUSTOMER_INFO_UPDATE.

JakeUrban:
> Right. The new failure mode I was trying to avoid was more along the lines of calling a callback from an RPC. We don't do this today. The customer callback will be made in the callback event processor.

I think we need to make the request from RPC while processing the `notify_customer_info_updated` call. If we don't, the customer status could change again before we make the callback from the event processor.

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
```diff
@@ -64,13 +70,14 @@ public ClientStatusCallbackHandler(
...
     this.sep6TransactionStore = sep6TransactionStore;
+    this.customerIntegration = customerIntegration;
     this.sep6MoreInfoUrlConstructor = sep6MoreInfoUrlConstructor;
     this.sep24MoreInfoUrlConstructor = sep24MoreInfoUrlConstructor;
   }
 
   @Override
   boolean handleEvent(AnchorEvent event) throws IOException {
-    if (event.getTransaction() != null) {
+    if (event.getTransaction() != null || event.getCustomer() != null) {
```
Ifropc, line 80: Where is this event fired?

Ifropc (commented):
Could you add an integration test for this feature? Just checking that you can freely move statuses is fine
NEEDS_INFO -> PROCESSING -> NEEDS_INFO -> NEEDS_INFO -> ACCEPTED 
(covers all non-error transitions ^)
Ideally we should also check that we receive callback on the wallet side

philipliu:
@Ifropc I'm planning to do this in another PR along with the callback config if that's okay. I was originally going to add it in this PR but it's already quite large with the unit tests.

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
philipliu: https://github.com/stellar/java-stellar-anchor-sdk/blob/3e6dc69754556e40400c03eac3992e48022c5fb8/core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java#L112-L119

Ifropc: From my understanding it's only called if SEP12 was called. I think it's no longer a correct place to fire an event, we should do it from the RPC call. 
Because otherwise events won't be fired when anchor updates from PROCESSING to other status

api-schema/src/main/java/org/stellar/anchor/api/rpc/method/NotifyCustomerInfoUpdatedRequest.java
```diff
@@ -9,4 +9,6 @@
 @SuperBuilder
 @AllArgsConstructor
 @EqualsAndHashCode(callSuper = false)
-public class NotifyCustomerInfoUpdatedRequest extends RpcMethodParamsRequest {}
+public class NotifyCustomerInfoUpdatedRequest extends RpcMethodParamsRequest {
+  String status;
+}
```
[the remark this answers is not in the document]
philipliu: Should SEP-31 transactions also follow the same logic? Currently, calling this RPC always changes the transaction status to `PENDING_RECEIVER`. Adding a status field would allow us to preserve the original behavior.

Anyway, I believe the answer is yes. We want SEP-31 to call `GET /customer` and update the transaction status based on the response.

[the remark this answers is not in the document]
JakeUrban: Yes I think thats correct.

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
Ifropc: > only expected to make notify_customer_info_updated RPC calls when the customer status updates asynchronously outside of a request/response cycle

No, I don't think that's the case. If you don't call the RPC, Platform doesn't know about status change. In the file Philip linked above we do the callback, but we only receive id in response. We simply fire an event and that's it. There's no check for kyc status, and there will be no update done to the transaction status.

philipliu: That makes sense. The wallet doesn't need to know about the updates it initiates; it only needs to know when the business server has updated its evaluation of the customer status, which is made known to the AP through this RPC.

philipliu: We need to keep firing the event here in case anyone depends on it. We should remove it in 3.0.

JakeUrban: I see. So if a business wants to provide SEP-12 callbacks, they need to call `notify_customer_info_updated` for every customer status change, even if that status change happened synchronously as a result of a wallet's `PUT /customer` request.

I feel like there is an opportunity to improve this.

If the anchor platform is going to call `GET /customer` as a result of receiving a `notify_customer_info_updated` request, should it also make a `GET /customer` as a result of a wallet's `PUT /customer` request as well? 

That way the business only uses `notify_customer_info_updated` when there is an update outside of a request/response cycle. The only time that happens is when a customer goes from `PROCESSING` to another status.

Or, we could update the SEP-12 protocol to return the full customer response body from `PUT /customer` requests, negating the need to send a callback.

Ifropc: Business should just return full body customer response in `PUT /customer` IMO

philipliu: > If the anchor platform is going to call GET /customer as a result of receiving a notify_customer_info_updated request, should it also make a GET /customer as a result of a wallet's PUT /customer request as well?
>
> That way the business only uses notify_customer_info_updated when there is an update outside of a request/response cycle. The only time that happens is when a customer goes from PROCESSING to another status.

It's a bit awkward making a GET after the PUT but that's my plan.

platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyCustomerInfoUpdatedHandler.java
```diff
@@ -59,13 +70,31 @@ public RpcMethod getRpcMethod() {
 
   @Override
   protected SepTransactionStatus getNextStatus(
-      JdbcSepTransaction txn, NotifyCustomerInfoUpdatedRequest request)
-      throws InvalidRequestException {
+      JdbcSepTransaction txn, NotifyCustomerInfoUpdatedRequest request) throws AnchorException {
+    String status = null;
+    if (request.getCustomerId() != null) {
```
JakeUrban, line 75: We should probably require the business to provide the customer ID and type in 3.0 right?

JakeUrban (approved):

platform/src/main/java/org/stellar/anchor/platform/rpc/NotifyCustomerInfoUpdatedHandler.java
philipliu: Yeah

platform/src/main/java/org/stellar/anchor/platform/event/ClientStatusCallbackHandler.java
Ifropc: @philipliu do you mind creating a ticket to implement this functionality in 3.x?

Ifropc (approved):

PR #1447, opened 2024-08-03
Fix RequestOnchainFunds incorrect hash memo

philipliu:
### Description

This sets the `memo` value to the request `memo` value.

### Context

 `memo.toString()` returns the hex value of the memo but the `memo` in the Payment object is in Base64 format, so the observer will never find the transaction in the database.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


Ifropc (approved):

Ifropc:
This is a bug, so I think we should add a Jira ticket for that too

PR #1451, opened 2024-08-07
[ANCHOR-752] Remove amount_out requirement in RequestOnchainFunds

philipliu:
### Description

`amount_out` is not set if a transaction has an associated indicative quote but has not yet received funds.

### Context

N/A

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/rpc/RequestOnchainFundsHandler.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/rpc/RequestOnchainFundsHandler.java
@@ -14,6 +14,7 @@ import com.google.common.collect.ImmutableSet;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
 import org.stellar.anchor.api.exception.AnchorException;
 import org.stellar.anchor.api.exception.BadRequestException;
 import org.stellar.anchor.api.exception.SepException;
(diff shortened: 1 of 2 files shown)


Ifropc (approved):

platform/src/main/java/org/stellar/anchor/platform/rpc/RequestOnchainFundsHandler.java
```diff
@@ -140,9 +140,6 @@ protected void validate(JdbcSepTransaction txn, RequestOnchainFundsRequest reque
     if (request.getAmountIn() == null && txn.getAmountIn() == null) {
       throw new InvalidParamsException("amount_in is required");
     }
-    if (request.getAmountOut() == null && txn.getAmountOut() == null) {
-      throw new InvalidParamsException("amount_out is required");
-    }
```
JakeUrban, line 145: Shouldn't we require `amount_out` when a quote was used or when `amount_in_asset == amount_out_asset`?

Ifropc: Good point. I think we should require amount_out if there was no quote as well

JakeUrban: I don't think we can require `amount_out` if there was no quote, since the business doesn't have to commit to a payout amount until its about to send funds. The exception is if `amount_out_asset = amount_in_asset`, since its always 1:1 minus fees.

Ifropc: Sorry we need to require either amount_out to be present either in request or in transaction

Ifropc (approved):
LGTM thanks @philipliu !

PR #1453, opened 2024-08-08
[ANCHOR-425][1/n] Deprecate platform REST API

JiahuiWho:
### Description
1. Add rpc method for `/transactions/id` and `/transactions` instead
2. Annotate deprecated endpoints and services used for these endpoints.

### Testing

- `./gradlew test`

### Documentation

1. Add rpc method`/transactions/id` and `/transactions` to stellar doc
2. Remove https://github.com/stellar/stellar-docs/blob/main/openapi/anchor-platform/Platform%20API.yml (No longer available on the website)


JakeUrban:
Can we update the PR title to clarify that we're only deprecating the REST/non-RPC endpoints of the platform API?

core/src/main/java/org/stellar/anchor/apiclient/PlatformApiClient.java
```diff
@@ -49,17 +41,33 @@ public PlatformApiClient(AuthHelper authHelper, String endpoint) {
...
    * @throws AnchorException if the response is not successful.
    */
+  @Deprecated // ANCHOR-641 Use getTransactionByRpc instead
   public GetTransactionResponse getTransaction(String id) throws IOException, AnchorException {
     Request request = getRequestBuilder().url(endpoint + "/transactions/" + id).get().build();
     String responseBody = handleResponse(client.newCall(request).execute());
     return gson.fromJson(responseBody, GetTransactionResponse.class);
   }
 
+  public GetTransactionResponse getTransactionByRpc(String id) throws IOException, AnchorException {
```
lijamie98, line 51: Is there a use case that we have to keep the deprecated API in the `PlatformApiClient` class or interface? 

I think if the `PlatformApiClient` do not have to support the deprecated endpoint. We may change the `getTransaction()` implementation as RPC. 

Consider the pattern where the `PlatformClientApi` is an interface, it would be better to replace the implementation instead of adding additional functions.

JiahuiWho: I will complete this ticket in separate PRs as rpc methods require some different logic like error handling. 

This is the 1st one that implement new endpoint. 
Next step is migrate existing usage of old REST endpoints to new one and fix tests.
The last step is to replace/rename the rpc endpoint and get rid of the old one.

Let me know if this PR looks good to you and I can move to next step

platform/src/main/java/org/stellar/anchor/platform/rpc/GetTransactionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+import org.stellar.anchor.metrics.MetricsService;
+import org.stellar.anchor.platform.data.JdbcSepTransaction;
+import org.stellar.anchor.platform.validator.RequestValidator;
+import org.stellar.anchor.sep24.Sep24TransactionStore;
+import org.stellar.anchor.sep31.Sep31TransactionStore;
+import org.stellar.anchor.sep6.Sep6TransactionStore;
+import org.stellar.anchor.util.Log;
+import org.stellar.anchor.util.SepHelper;
+
+public class GetTransactionHandler extends RpcMethodHandler<GetTransactionRpcRequest> {
```
philipliu, line 26: What do you think of making `RpcMethodHandler` more generic so that it only has the `handle` and `getRpcMethod`? We can create a subclass of that called something like `TransactionRpcMethodHandler` with `getNextStatus`, `getSupportedStatuses`, `getSupportedStatuses`, and `updateTransactionWithRpcRequest` that the existing RPCs can extend.

philipliu (commented):
The changes generally look good. I have a suggestion for making the code a little nicer.

platform/src/main/java/org/stellar/anchor/platform/rpc/GetTransactionHandler.java
JiahuiWho: It will make the structure a little bit more clear bit not much since functions related to status update (`getNextStatus`, `getSupportedStatuses`, `getSupportedStatuses`, `updateTransactionWithRpcRequest`, `isErrorStatus` and `isFinalStatus`) are embedded in the `handle` function in `RpcMethodHandler` class. 

Eventually everything in `RpcMethodHandler` will be moved to the subclass and leave `RpcMethodHandler` an empty parent class.

PR #1454, opened 2024-08-08
[ANCHOR-749] Add missing validations to the RateIntegration response

lijamie98:
### Description

Add missing validations to the RateIntegration response

### Context

Some of the anchors return invalidate rate responses. We should catch and reject. 

### Testing

- `./gradlew test`


platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+              "'sell_amount' is not equal to price * buy_amount + (fee ?: 0) to  in the GET /rate response",
+              ServerErrorException.class);
+        }
+      }
+    } else {
+      // when fee is not present, check that sell_amount is equal to price * buy_amount
+      BigDecimal expected =
+          new BigDecimal(rate.getPrice())
+              .multiply(new BigDecimal(rate.getBuyAmount()))
+              .setScale(sellAsset.getSignificantDecimals(), HALF_UP);
```
Ifropc, line 205: We use FLOOR for total price, can we also use it here for consistency?
https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java#L531

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+          logErrorAndThrow(
+              "'sell_amount' is not equal to price * (buy_amount + (fee ?: 0)) in the GET /rate response",
+              ServerErrorException.class);
+        }
+      }
+
+      if (fee.getDetails() != null) {
+        BigDecimal totalFee =
+            new BigDecimal(0)
+                .setScale(Objects.requireNonNull(feeAsset).getSignificantDecimals(), HALF_UP);
```
Ifropc, line 178: I don't think you need to set scale here. 
1. When adding fees from the details it should already be counted on (scale would be <= significantDecimals and if it's not, the total check on line 194 would fail) 
2. compareTo doesn't care about the scale. 0 with scale 20 == 0 with scale 0 when used compareTo

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+      }
+
+      if (fee.getAsset().equals(request.getSellAsset())) {
+        // when fee is in sell_asset,
+        // check that sell_amount is equal to price * buy_amount + (fee ?: 0)
+        BigDecimal expected =
+            new BigDecimal(rate.getPrice())
+                .multiply(new BigDecimal(rate.getBuyAmount()))
+                .add(new BigDecimal(fee.getTotal()))
+                .setScale(sellAsset.getSignificantDecimals(), HALF_UP);
```
Ifropc, line 154: We use FLOOR for total price, can we also use it here for consistency?
https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java#L531

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+              "'sell_amount' is not equal to price * buy_amount + (fee?:0) in the GET /rate response",
+              ServerErrorException.class);
+        }
+      } else {
+        // when fee is in buy_asset,
+        // check that sell_amount is equal to price * (buy_amount + (fee ?: 0))
+        BigDecimal expected =
+            new BigDecimal(rate.getPrice())
+                .multiply(new BigDecimal(rate.getBuyAmount()).add(new BigDecimal(fee.getTotal())))
+                .setScale(sellAsset.getSignificantDecimals(), HALF_UP);
```
Ifropc, line 167: We use FLOOR for total price, can we also use it here for consistency?
https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/sep38/Sep38Service.java#L531

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
-        Log.error("'id' and/or 'expires_at' are missing in the GET /rate response");
-        throw new ServerErrorException("internal server error");
+        logErrorAndThrow(
+            "'id' and/or 'expires_at' are missing in the GET /rate response",
+            ServerErrorException.class);
+      }
+    }
+
+    // sell_amount is present and positive number
+    if (!isPositiveNumber(Objects.requireNonNull(rate).getSellAmount())) {
```
Ifropc, line 117: We should also validate that sell, buy and fee amounts have the scale corresponding to their asset's scale

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
```diff
@@ -0,0 +1,231 @@
...
+import org.stellar.anchor.api.callback.GetRateRequest.Type.from
+import org.stellar.anchor.api.callback.GetRateResponse
+import org.stellar.anchor.api.exception.ServerErrorException
+import org.stellar.anchor.api.sep.AssetInfo
+import org.stellar.anchor.api.sep.AssetInfo.Schema.iso4217
+import org.stellar.anchor.api.sep.AssetInfo.Schema.stellar
+import org.stellar.anchor.asset.AssetService
+import org.stellar.anchor.util.GsonUtils
+
+class RestRateIntegrationTest {
```
Ifropc, line 24: Can you please add tests here for assets that have different scale?
The best example is (fiat) with scale 2 <-> USDC with scale 7 conversion

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -4,7 +4,7 @@ assets:
     code: USDC
     issuer: GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
     distribution_account: GBN4NNCDGJO4XW4KQU3CBIESUJWFVBUZPOKUZHT7W7WRB7CWOA7BXVQF
-    significant_decimals: 2
+    significant_decimals: 4
```
Ifropc, line 7: Should be 7 FYI

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -190,7 +190,7 @@ assets:
   # Fiat
   - schema: iso4217
     code: USD
-    significant_decimals: 7
+    significant_decimals: 4
```
Ifropc, line 193: And fiat should be 2

core/src/main/java/org/stellar/anchor/util/ErrorHelper.java
```diff
@@ -0,0 +1,18 @@
+package org.stellar.anchor.util;
+
+import static org.stellar.anchor.util.Log.*;
+
+public class ErrorHelper {
```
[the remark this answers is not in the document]
philipliu: Generally, I would avoid doing log and throw. We should let the code catching the exception decide if they want to log the error.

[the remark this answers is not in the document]
lijamie98: I remember a discussion that we would like to log error and throw and this logic is prevalent in the code. I am ok with either logging at the catcher or both. But we should not go back and forth.

The logic was in the original code. If we decide to avoid log and throw, I can create a ticket for that purpose.

Ex: [RestRateIntegration.java](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java#L71)
[JwtService#L235](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/auth/JwtService.java#L234), [Sep1Service#62](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/sep1/Sep1Service.java#L62)

And some other places.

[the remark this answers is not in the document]
lijamie98: https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java#L71

`sout` removed. 

I am not a fan of using reflection either because it impacts the performance. However, in this case, performance is not a concern because we are throwing exception. But if we decide not to use log and throw pattern, this function can be removed. 

We may take care of the log and throw pattern in a different ticket.

In this particular use case, only `ServerErrorException` is thrown. We remove the function and just throw the exception

lib-util/src/main/java/org/stellar/anchor/util/NumberHelper.java
```diff
@@ -0,0 +1,16 @@
+package org.stellar.anchor.util;
+
+public class NumberHelper {
+  public static boolean isPositiveNumber(String str) {
```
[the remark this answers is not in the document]
lijamie98: Good suggestion. Modified and added tests.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: That's right, we don't need to set scale to validate the amount. If the response has floating errors caused by inconsistent decimals in the fee details, we should throw the error. 

I will remove the `setScale`.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: See above.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: See above.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: Good suggestions. Tests added.

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
lijamie98: Changed to `usdcAssetInfo.significantDecimals = 7`

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+  }
+
+  void validateRateResponse(GetRateRequest request, GetRateResponse getRateResponse)
+      throws ServerErrorException {
+    AssetInfo sellAsset = assetService.getAssetByName(request.getSellAsset());
     GetRateResponse.Rate rate = getRateResponse.getRate();
     if (rate == null || rate.getPrice() == null) {
-      Log.error("missing 'price' in the GET /rate response");
-      throw new ServerErrorException("internal server error");
+      logErrorAndThrow("missing 'price' in the GET /rate response", ServerErrorException.class);
```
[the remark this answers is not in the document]
Ifropc: Right it's originally called from the user request, sounds good to keep it 500

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
Ifropc: I think it comes from Bitso's request originally. Now I think we should just stick to it, and (better) declare a constant somewhere that is currently set to FLOOR. We can allow it to be configured in the future, if requested by some anchor

lib-util/src/main/java/org/stellar/anchor/util/NumberHelper.java
```diff
@@ -14,4 +16,17 @@ public static boolean isPositiveNumber(String str) {
       return false;
     }
   }
+
+  public static boolean hasProperSignificantDecimals(
```
Ifropc, line 20: Should we remove `mixDecimals` for now? It's only used in tests, not in the main code. We can always add an extra parameter latert

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +84,159 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+      logErrorAndThrow(
+          "'buy_amount' has incorrect number of significant decimals in the GET /rate response",
+          ServerErrorException.class);
+    }
+
+    FeeDetails fee = rate.getFee();
+    // if fee is preset, check the following
+    if (fee != null) {
+      // fee.total is present and is a positive number
+      if (!isPositiveNumber(fee.getTotal())) {
```
philipliu, line 153: It would be nice if this was a precondition for constructing a `FeeDetails` object.

platform/src/main/java/org/stellar/anchor/platform/utils/AssetValidationUtils.java
```diff
@@ -70,7 +58,19 @@ public static void validateFeeDetails(
...
+   *
+   * @param amount is the object containing the asset full name and the amount.
+   */
+  public static void validateAssetAmount(
+      String fieldName, AmountAssetRequest amount, AssetService assetService)
+      throws BadRequestException {
+    validateAssetAmount(fieldName, amount, false, assetService);
+  }
+
+  public static void validateAssetAmount(
```
philipliu, line 73: minor: Let's add Javadoc for these methods.

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
```diff
@@ -0,0 +1,273 @@
...
+
+    // Bad buy amount
+    rateResponse.rate.sellAmount = "100"
+    rateResponse.rate.buyAmount = badAmount
+    ex =
+      assertThrows<ServerErrorException> {
+        rateIntegration.validateRateResponse(request, rateResponse)
+      }
+    assertEquals(
+      "'buy_amount' is missing or not a positive number in the GET /rate response",
```
philipliu, line 141: Should we also test for the significant decimals validation?

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
```diff
@@ -0,0 +1,273 @@
...
+      ex.message,
+    )
+  }
+
+  @Test
+  fun `test mis-matched sell_amount and buy_amount`() {
+    // Bad sell amount
+    rateResponse.rate.sellAmount = "100.01"
```
philipliu, line 149: It's hard to test why the test fails. Can we also set `buyAmount` and `fee` here and maybe add a comment with the expected amount?

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
```diff
@@ -0,0 +1,273 @@
...
+      assertThrows<ServerErrorException> {
+        rateIntegration.validateRateResponse(request, rateResponse)
+      }
+    assertEquals(
+      "'fee.details[?].description.amount' is missing or not a positive number in the GET /rate response",
+      ex.message,
+    )
+
+    rateResponse.rate.fee.details[feeIndex].name = "Sell fee"
+    rateResponse.rate.fee.details[feeIndex].amount = "0.71"
```
philipliu, line 221: Can we add a comment with what the expected amount should be, and why?

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: `DEFAULT_ROUNDING_MODE` added to `NumberHelper` which is a temporary placeholder for future configuration.

platform/src/main/java/org/stellar/anchor/platform/utils/AssetValidationUtils.java
lijamie98: Improved.

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
lijamie98: Added.

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
lijamie98: Added `buyAmount` and commented.

platform/src/test/kotlin/org/stellar/anchor/platform/callback/RestRateIntegrationTest.kt
lijamie98: Added a comment.

service-runner/src/main/resources/config/assets.yaml
lijamie98: It is weird to use `7` in the reference server. WDYT we modify the reference to use `4` for the reason of "reference".

core/src/main/java/org/stellar/anchor/util/ErrorHelper.java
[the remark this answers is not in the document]
lijamie98: Ticket created for future improments.

[ANCHOR-779](https://stellarorg.atlassian.net/browse/ANCHOR-779)

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
philipliu: I mean including this validation in the `FeeDetails` constructor so it's impossible to initialize an instance with negative fees.

lijamie98: ah, I see. 

The `FeeDetails` is in the api-schema sub-project. We did not put much logic in that specific sub-project. 

If we decide it is good to introduce that pattern in the code, I may create a ticket to introduce this and do a project wide review.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +84,156 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+      }
+
+      if (fee.getAsset().equals(request.getSellAsset())) {
+        // when fee is in sell_asset,
+        // check that sell_amount is equal to price * buy_amount + (fee ?: 0)
+        BigDecimal expected =
+            new BigDecimal(rate.getPrice())
+                .multiply(new BigDecimal(rate.getBuyAmount()))
+                .add(new BigDecimal(fee.getTotal()))
+                .setScale(sellAsset.getSignificantDecimals(), DEFAULT_ROUNDING_MODE);
```
philipliu, line 170: I thought we decided not to include amount validation in 2.9 and add it in 3.0? Did I misremember?

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+      throws ServerErrorException {
+    AssetInfo sellAsset = assetService.getAssetByName(request.getSellAsset());
     GetRateResponse.Rate rate = getRateResponse.getRate();
     if (rate == null || rate.getPrice() == null) {
-      Log.error("missing 'price' in the GET /rate response");
-      throw new ServerErrorException("internal server error");
+      logErrorAndThrow("missing 'price' in the GET /rate response", ServerErrorException.class);
     }
 
     if (request.getType() != GetRateRequest.Type.INDICATIVE) {
```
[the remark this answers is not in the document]
lijamie98: merged.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: You remember it correctly. It is disabled for 2.9.0 in `equalsInScale` function.

core/src/main/java/org/stellar/anchor/util/ErrorHelper.java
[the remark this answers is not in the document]
Ifropc: If we still do logging + throw in this PR, can we at least inline this method and not introduce a reflection helper in the codebase?

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +83,131 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+    return getRateResponse;
+  }
+
+  void validateRateResponse(GetRateRequest request, GetRateResponse getRateResponse)
+      throws ServerErrorException {
+    AssetInfo sellAsset = assetService.getAssetByName(request.getSellAsset());
     GetRateResponse.Rate rate = getRateResponse.getRate();
     if (rate == null || rate.getPrice() == null) {
-      Log.error("missing 'price' in the GET /rate response");
-      throw new ServerErrorException("internal server error");
```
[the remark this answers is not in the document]
Ifropc: My point, do we _really_ need to 
```
Log.error
throw new ...
```
?
IIRC our generic handler already logs an error, no?

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +88,186 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+   * <p>TODO: Returns true in release 2.9.0 to disable the total amount comparison. Will readd in
+   * 3.0.0
+   *
+   * @param amount The amount to be compared
+   * @param expected The expected amount
+   * @param scale The scale to be used for comparison
+   * @return true if the amounts are equal in scale with any of the allowed rounding modes, false
+   *     otherwise
+   */
+  private boolean equalsInScale(BigDecimal amount, BigDecimal expected, int scale) {
```
Ifropc, line 264: Can we do abs approach for now? Otherwise we are not checking anything. It will return true for any 2 numbers, so this validation does nothing

Ifropc, line 264: Why does it accept scale as an argument? It should validate that amount and expected are already in the same scale

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +88,186 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+            new BigDecimal(rate.getPrice())
+                .multiply(new BigDecimal(rate.getBuyAmount()).add(new BigDecimal(fee.getTotal())));
+        // Since we don't know how the anchor rounds the amounts, we need to check the equality in
+        // all allowed rounding modes
+        if (!equalsInScale(
+            new BigDecimal(rate.getSellAmount()), expected, sellAsset.getSignificantDecimals())) {
+          logErrorAndThrow(
+              "'sell_amount' is not equal to price * (buy_amount + (fee ?: 0)) in the GET /rate response",
+              ServerErrorException.class);
+        }
```
Ifropc, line 196: This can be moved out of if/else branch

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +88,186 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+    // if fee is set, check the following
+    if (fee != null) {
+      // fee.total is present and is a positive number
+      if (!isPositiveNumber(fee.getTotal())) {
+        logErrorAndThrow(
+            "'fee.total' is missing or not a positive number in the GET /rate response",
+            ServerErrorException.class);
+      }
+      // fee.asset is a valid asset
+      AssetInfo feeAsset = assetService.getAssetByName(fee.getAsset());
```
Ifropc, line 160: Should also check for fee's proper significant decimals

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +88,186 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+            logErrorAndThrow(
+                "'fee.details[?].description.amount' has incorrect number of significant decimals in the GET /rate response",
+                ServerErrorException.class);
+          }
+          if (feeDescription.getName() == null) {
+            logErrorAndThrow(
+                "'fee.details.description[?].name' is missing in the GET /rate response",
+                ServerErrorException.class);
+          }
+          if (!isPositiveNumber(feeDescription.getAmount())) {
```
Ifropc, line 219: ALready checked on line 202

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +88,186 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+                "'fee.details[?].description.amount' is missing or not a positive number in the GET /rate response",
+                ServerErrorException.class);
+          }
+          totalFee = totalFee.add(new BigDecimal(feeDescription.getAmount()));
+        }
+
+        // check that sell_amount is equal to price * buy_amount + (fee ?: 0)
+        if (totalFee.compareTo(new BigDecimal(fee.getTotal())) != 0) {
+          logErrorAndThrow(
+              "'sell_amount' is not equal to price * buy_amount + (fee ?: 0) to  in the GET /rate response",
```
Ifropc, line 230: Error message copy-pasted

Ifropc (changes requested):

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: The function should take the scale as the prevision of the asset.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: They are actually different and cannot be moved out.
```
// check that sell_amount is equal to price * buy_amount + (fee ?: 0)
// check that sell_amount is equal to price * (buy_amount + (fee ?: 0))
```

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: Added.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: It is in `abs` approach.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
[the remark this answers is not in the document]
lijamie98: The refactoring of this code will be tabled after decision of using "log and throw" pattern.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +90,196 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+      if (fee.getAsset() == null || feeAsset == null) {
+        logErrorAndThrow(
+            "'rate.fee.asset' is missing or not a valid asset in the GET /rate response",
+            ServerErrorException.class);
+      }
+
+      if (!isPositiveNumber(Objects.requireNonNull(feeAsset.getSignificantDecimals()).toString())) {
+        logErrorAndThrow(
+            format(
+                "The rate fee.asset (%s) does not have proper significant decimals defined in the assets configuration.",
```
Ifropc, line 178: Error message doesn't match the condition

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +90,196 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+                .multiply(new BigDecimal(rate.getBuyAmount()))
+                .add(new BigDecimal(fee.getTotal()));
+
+        // Since we don't know how the anchor rounds the amounts, we need to check the equality in
+        // all allowed rounding modes
+        if (!withinRoundingError(
+            new BigDecimal(rate.getSellAmount()), expected, sellAsset.getSignificantDecimals())) {
+          logErrorAndThrow(
+              format(
+                  "'rate.sell_amount' (%s) is not within rounding error of the expected (%s) ('price * buy_amount + (fee?:0)') in the GET /rate response",
```
Ifropc, line 206: fee here is non-zero, so just `price * buy_amount + fee`

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -68,29 +90,196 @@ public GetRateResponse getRate(GetRateRequest request) throws AnchorException {
...
+        BigDecimal expected =
+            new BigDecimal(rate.getPrice())
+                .multiply(new BigDecimal(rate.getBuyAmount()).add(new BigDecimal(fee.getTotal())));
+        // Since we don't know how the anchor rounds the amounts, we need to check the equality in
+        // all allowed rounding modes
+        if (!withinRoundingError(
+            new BigDecimal(rate.getSellAmount()), expected, sellAsset.getSignificantDecimals())) {
+          logErrorAndThrow(
+              format(
+                  "'rate.sell_amount' (%s) is not within rounding error of the expected (%s) ('price * (buy_amount + (fee?:0))') in the GET /rate response",
```
Ifropc, line 222: Same here and in the comments ^

service-runner/src/main/resources/config/assets.yaml
Ifropc: Should we fix reference server impl?

Ifropc (approved):
LGTM except minor comments

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: Fixed.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: Fixed.

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
lijamie98: Fixed.

service-runner/src/main/resources/config/assets.yaml
lijamie98: I think `4` is good. We need to make AP and reference server match. It is also good to try different decimals.

PR #1456, opened 2024-08-12
clean rate request

Ifropc:
### Description

Clean rate request

### Context

Rate request have `id` field that is never used 

### Testing

- `./gradlew test`

### Documentation

N/A (already not in the spec)

### Known limitations

N/A


lijamie98 (changes requested):
We need to make sure the `id` is not used before clean it up.

Ifropc:
> We need to make sure the id is not used before clean it up.

It was only used in integration test and nowhere else. The code wouldn't compile if there were any other usages

lijamie98 (approved):
LGTM after discussion.

PR #1460, opened 2024-08-15
[ANCHOR-764] Add SEP-12 status callback tests

philipliu:
### Description

This adds client callback URL configuration and updates the SEP-6 end-to-end tests to check for SEP-12 status updates. During the implementation, I realized that the previously implemented `RequestCustomerInfoUpdate` RPC is no longer needed after the recent `NotifyCustomerInfoUpdate` change. I've marked the method as deprecated and will remove it in 3.0.

Since this change added new routes to the Wallet server, I took this opportunity to clean up the code a little.

### Context

This adds test coverage for a previous change.

### Testing

- `./gradlew test`

### Documentation

stellar-docs PR will be updated to reflect the `NotifyCustomerInfoUpdate` field name change and `RequestCustomerInfoUpdate` deprecation.

### Known limitations

N/A


essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep6End2EndTest.kt
```diff
@@ -120,6 +122,14 @@ open class Sep6End2EndTest : AbstractIntegrationTests(TestConfig()) {
...
       )
     assertWalletReceivedStatuses(deposit.id, expectedStatuses)
+
+    val expectedCustomerStatuses =
+      listOf(
+        Sep12Status.ACCEPTED, // initial customer status before SEP-6 transaction
+        Sep12Status.NEEDS_INFO, // SEP-6 transaction requires additional info
+        Sep12Status.ACCEPTED // additional info provided
+      )
+    assertWalletReceivedCustomerStatuses(customer.id, expectedCustomerStatuses)
```
Ifropc, line 132: Should it not be a state automata? (we always know the sequence of statuses)

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -182,7 +182,18 @@ event_processor:
 #   If the type is `custodial`,
 #   - signing_keys: (required) the custodial SEP-10 signing keys of the client.
 #   - callback_url: (optional) the URL of the client's callback API endpoint. Optional due to some wallets may opt to
```
Ifropc, line 184: I think we should also deprecate it now, and remove in 3.0

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
```diff
@@ -383,11 +383,21 @@ class Sep6EventProcessor(
           event.payload.transaction.id,
           missingFields,
         )
+        val existingCustomer =
```
Ifropc, line 386: Can we slightly change the logic for customer?
1. request_kyc
2. verify_kyc transition into PROCESSING if id_number is provided, otherwise more fileds needed
3. spawn a coroutine that waits 3 seconds and moves status into ACCEPTED
This will simulate a typical anchor flow. 
(We can do it in a separate PR)

Ifropc (approved):

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep6End2EndTest.kt
philipliu: Good point. It should use an `assertEquals` instead.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep6EventProcessor.kt
philipliu: That's a good idea. I'll do this in a separate PR. I would like the refactor this class first since it's becoming very difficult to understand and extend.

philipliu: Created ANCHOR-778.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -185,7 +185,7 @@ event_processor:
 #                  poll instead, or may use polling first before implementing callbacks at a later stage. If one of
 #                  SEP-specific URLs is also provided, then this URL will be used as a fallback. For example, if
 #                  `callback_url_sep6` is not provided, but `callback_url_sep24` is, SEP-6 transactions will use the
-#                  `callback_url` as the callback URL.
+#                  `callback_url` as the callback URL. This field will be deprecated in 3.0.
```
Ifropc, line 188: Will be removed*

PR #1461, opened 2024-08-16
[ANCHOR-732] Add SEP-6 SEP-24 as top-level field in Assets config

JiahuiWho:
### Description
This is part of the assets config refactor work

1. Unify the naming to `Sep#Info` for all sep attributions
2. Add `sep6` and `sep24` as top-level field, and move `deposit` and `withdraw` under each of them
3. Move the "sep#_enabled" under corresponding sep field
4. remove more `deposit` and `withdraw` from test fiat asset
5. rename `send` as `receive` for SEP-31 service

The new schema will be:
  shema:
  code:
  issuer:
  ...
  sep6:
    -- enabled:
    -- deposit:
    -- withdraw:
  sep24:
    -- enabled:
    -- deposit:
    -- withdraw:
  sep31:
    -- enabled:
    -- receive:
  sep38:
    -- enabled:

### Testing

- `./gradlew test`

### DIFF
--- a/service-runner/src/main/resources/config/assets.yaml
+++ b/service-runner/src/main/resources/config/assets.yaml
@@ -39,7 +39,7 @@ assets:
           - cash
     sep31:
       enabled: true
-      send:
+      receive:
         fee_fixed: 0
         fee_percent: 0
         max_amount: 1000000
@@ -95,7 +95,7 @@ assets:
         max_amount: 10
     sep31:
       enabled: true
-      send:
+      receive:
         fee_fixed: 0
         fee_percent: 0
         min_amount: 0
@@ -154,7 +154,7 @@ assets:
         enabled: false
     sep31:
       enabled: true
-      send:
+      receive:
         fee_fixed: 0
         fee_percent: 0
       quotes_supported: true
@@ -220,7 +220,7 @@ assets:
     significant_decimals: 7
     sep31:
       enabled: false
-      send:
+      receive:
         min_amount: 0
         max_amount: 1000000
     sep38:
@@ -241,7 +241,7 @@ assets:
     code: CAD
     significant_decimals: 7
     sep31:
-      send:
+      receive:
         min_amount: 0
         max_amount: 1000000
     sep38:
--- a/service-runner/src/main/resources/config/assets.yaml
+++ b/service-runner/src/main/resources/config/assets.yaml
@@ -5,21 +5,40 @@ assets:
     issuer: GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
     distribution_account: GBN4NNCDGJO4XW4KQU3CBIESUJWFVBUZPOKUZHT7W7WRB7CWOA7BXVQF
     significant_decimals: 2
-    deposit:
+    sep6:
       enabled: true
-      min_amount: 0
-      max_amount: 10
-      methods:
-        - SEPA
-        - SWIFT
-    withdraw:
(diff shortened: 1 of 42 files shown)


Ifropc (approved):

JakeUrban:
One ❓ -- do we need `sep31.send`, given that SEP-31 doesn't have multiple types of transactions like `deposit` and `withdraw`?

Another way to think about it is that SEP-31 has two perspectives, the sending anchor and receiving anchor. The anchor platform today only supports receive-side -- so maybe we should use `sep31.receive` instead?

JiahuiWho:
The decision to keep `send` is mainly about maintaining consistency with SEP-6 and SEP-24. I agree `receive` makes more sense than `send`.

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -208,15 +237,15 @@ assets:
       buy_delivery_methods:
         - name: WIRE
           description: Have USD sent directly to your bank account.
-    sep38_enabled: true
   - schema: iso4217
     code: CAD
     significant_decimals: 7
     sep31:
```
lijamie98, line 243:
```suggestion
    sep31:
      enabled: true
```

service-runner/src/main/resources/config/assets.yaml
```diff
@@ -5,22 +5,41 @@ assets:
     issuer: GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
     distribution_account: GBN4NNCDGJO4XW4KQU3CBIESUJWFVBUZPOKUZHT7W7WRB7CWOA7BXVQF
     significant_decimals: 2
-    deposit:
+    sep6:
```
lijamie98, line 8: Can we consider
```
  assets:
    - id: stellar:USDC::GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
```
to replace with
```
  assets:
    - schema: stellar
      code: USDC
      issuer: GDQOE23CFSUMSVQK4Y5JHPPYK73VYCNHZHA7ENKCV37P6SUEO6XQBKPP
```

JiahuiWho: I don't think that's necessary. If we have the former format we will need to break it down for other SEP naming.

Beside this we are going to separate stellar asset from fiat asset (there will be 2 list like `clients` config)

service-runner/src/main/resources/config/assets.yaml
JiahuiWho: Is this how we interpret if service is enabled now? 
I can use some clarification so that I can clean up the test asset config all at once.

This PR assume service is **NOT** enabled unless `enabled` is set to true

service-runner/src/main/resources/config/assets.yaml
JakeUrban: I don't think we should expect devs to use string formats defined in our SEPs. Its more human-readable to have key-val pairs for each piece of information.

JiahuiWho: Hi @lijamie98 @JakeUrban  I want to move this discussion to the main ticket https://stellarorg.atlassian.net/browse/ANCHOR-732 to save it for future reference

service-runner/src/main/resources/config/assets.yaml
JiahuiWho: I just tested it and my assumption is right.

PR #1462, opened 2024-08-16
[ANCHOR-732] Remove sep12 from asset config

JiahuiWho:
### Description

As title.

Discussed with Jake that  a customer could only have one id with the anchor, and the customer type is only used for identify KYC info purposes, so we don't actually need to traverse all the customer type in DELETE /customer request.

### Testing

- `./gradlew test`

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
@@ -10,9 +10,7 @@ import java.time.LocalDate;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 import java.util.Objects;
-import java.util.Set;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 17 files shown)


lijamie98:
Nice to clean up 👍

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -148,20 +132,14 @@ public void deleteCustomer(Sep10Jwt sep10Jwt, String account, String memo, Strin
...
-                  .memoType(memoType)
-                  .type(customerType)
-                  .build());
-      if (existingCustomer.getId() != null) {
-        existingCustomerMatch = true;
-        customerIntegration.deleteCustomer(existingCustomer.getId());
-      }
+    GetCustomerResponse existingCustomer =
+        customerIntegration.getCustomer(
+            GetCustomerRequest.builder().account(account).memo(memo).memoType(memoType).build());
```
philipliu, line 137: This assumes the business server gives unique IDs to account, and memo pairs rather than account, and memo, types. I don't think this assumption is captured in docs. I think we should mention this somewhere.

core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
```diff
@@ -524,17 +521,17 @@ class Sep12ServiceTest {
     assertInstanceOf(SepNotFoundException::class.java, ex)
     assertEquals("User not found.", ex.message)
     // Verify getting customer for every existing type
```
philipliu, line 523: nit: comment needs to be updated.

philipliu (approved):

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
JiahuiWho: Tracking here: https://stellarorg.atlassian.net/browse/ANCHOR-780

PR #1463, opened 2024-08-20
[ANCHOR-732] Add separate schema for stellar asset and fiat asset

JiahuiWho:
### Update for review

1. Pls take a look at files under `api-schema/src/main/java/org/stellar/anchor/api/asset/` this is where new object stays.
2. Also kindly check `Sep6Service` and `Sep24Service` as there will be some asset type cast.

### Description

This PR creates separate object schema for stellar asset and fiat asset

1. `schema`, `code` and `issuer` will be replaced by `id` in the format of <scheme:code:issuer>, which is the Asset Identification Format used in SEP-38
2. Difference between asset:
Only stellar asset will have `distribution_account`, `sep6`, `sep24`
Only fiat asset will have `sep38.buy_deliver_method` and `sep38.sell_delivery_method`

`Assets` config will eventually looks like:
 ```
  - id: stellar:USDC:GBRTVW74KXDLZQS2QXX6TEBT4CTCAZ7EUUB4N4YGLJZ7UNJIC2I6G2PL
    sep6:
      enabled: false
    sep24:
      enabled: false
    sep38:
      enabled: true
      exchangeable_assets:
        - iso4217:USD
  - id: iso4217:USD
    sep38:
      enabled: true
      exchangeable_assets:
        - stellar:USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5
      buy_delivery_methods:
        - name: ACH
          description: ACH withdrawals
      sell_delivery_methods:
        - name: ACH
          description: ACH deposits 

```
### Testing

- `./gradlew test`

### CODE REVIEW

**Comment by philipliu** [comment]
What do you think about removing the `enabled` fields in the sep configs? 

Today it's possible to configure the following:
```yaml
sep24:
  enabled: true
  deposit:
    enabled: false
  withdraw:
    enabled: false
```
What are we enabling here if both deposit and withdraw are disabled? I think it's better remove the top level `enabled` field.

lijamie98:
Is the purpose of this PR to replace [AssetInfo.java](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java)?

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
```diff
@@ -0,0 +1,27 @@
...
+   * - iso4217: Used for fiat currencies. The identifier follows the ISO 4217 three-character currency code.
+   * For example:
+   * - Stellar USDC would be identified as: stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN
+   * - Fiat USD would be identified as: iso4217:USD
+   *
+   * @return A string representing the asset identification name, formatted as <scheme>:<identifier>.
+   */
+  String getAssetIdentificationName();
+
+  InfoResponse.Asset toSEP38InfoResponseAsset();
```
lijamie98, line 24: What's the advantage of implementing this in the interface over adding the conversion function in the [sep38.InfoResponse](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/api-schema/src/main/java/org/stellar/anchor/api/sep/sep38/InfoResponse.java)?

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
```diff
@@ -0,0 +1,27 @@
...
+
+  Sep31InfoResponse.AssetResponse toSEP31InfoResponseAsset();
```
lijamie98, line 26: See above

JiahuiWho:
> What do you think about removing the `enabled` fields in the sep configs?
I thought about removing it for sep6 and sep24 cause it's kind duplicate.. but then if we want to keep consistency, we should keep it

> Today it's possible to configure the following:
> 
> ```yaml
> sep24:
>   enabled: true
>   deposit:
>     enabled: false
>   withdraw:
>     enabled: false
> ```
> 
> What are we enabling here if both deposit and withdraw are disabled? I think it's better remove the top level `enabled` field.
In this case none is enabled. And this is how we interpret the original asset config, we need to specify  both `sep6_enabled` and `deposit` to true to enable the service.

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
```diff
@@ -1,14 +1,45 @@
...
+  String code;
   String issuer;
 
   @SerializedName("distribution_account")
   String distributionAccount;
+
+  @SerializedName("significant_decimals")
+  Integer significantDecimals;
+
+  DepositWithdrawInfo sep6;
```
lijamie98, line 21: Is there a reason to use different class name from the variable name? 

ex: `Sep6Info sep6`

lijamie98 (changes requested):

JiahuiWho:
> Is the purpose of this PR to replace [AssetInfo.java](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java)?

No ... replace AssetInfo might be the result ... the purpose is to create schema for different asset. Right now the AssetInfo mixed all fields in one object, it's much more clear to have separate schema, even for validation and testing purpose

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
JiahuiWho: Because sep6 and sep24 share the same attributes

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
JiahuiWho: I think this function should be part of the asset itself (or maybe helper function) instead of going to service, it's much clear to maintain if we want to add new asset schema, or add new fields to existing schema, in the future

lijamie98:
> > Is the purpose of this PR to replace [AssetInfo.java](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/api-schema/src/main/java/org/stellar/anchor/api/sep/AssetInfo.java)?
> 
> No ... replace AssetInfo might be the result ... the purpose is to create schema for different asset. Right now the AssetInfo mixed all fields in one object, it's kind confusing what fields belongs to which schema

If we are not replacing, it will leave the code with duplicated classes. Is there a specific reason not to replace the existing class?

JiahuiWho:
> If we are not replacing, it will leave the code with duplicated classes. Is there a specific reason not to replace the existing class?

I'm saying the initial purpose is to schematize fiat asset and stellar asset, not just removing AssetInfo. Replacing AssetInfo is the result, and we will remove it in the end.

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
philipliu: We will still need to implement the conversion method either, whether that is in the `AssetInfo` class or `InfoResponse`. I'm not sure I understand the advantage of implementing the method here. Since we will be using this method only in the SEP-38 code, I would prefer if we move it there as well.

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
```diff
@@ -2,13 +2,91 @@
...
+  DepositWithdrawInfo sep24;
+  Sep31Info sep31;
```
philipliu, line 22: nit: Can we call `Sep31Info` something like `SendInfo`  for consistency? Same with `Sep38Info`.

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
```diff
@@ -2,13 +2,91 @@
...
+      assetResponse.setMaxAmount(getSep31().getReceive().getMaxAmount());
+      assetResponse.setFields(getSep31().getFields());
+      return assetResponse;
+    }
+    return null;
+  }
+
+  @Override
+  public InfoResponse.Asset toSEP38InfoResponseAsset() {
+    if (getSep38().getEnabled()) {
```
philipliu, line 53: Do we need a null check?

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
```diff
@@ -2,13 +2,91 @@
...
+      InfoResponse.Asset newAsset = new InfoResponse.Asset();
+      newAsset.setAsset(getId());
+
+      Sep38Info sep38Info = getSep38();
+      newAsset.setCountryCodes(sep38Info.getCountryCodes());
+      newAsset.setExchangeableAssetNames(sep38Info.getExchangeableAssets());
+
+      newAsset.setDecimals(7);
```
philipliu, line 61: Can we add a comment explaining why the decimals is always set to 7?

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
```diff
@@ -2,13 +2,91 @@
...
+  }
+
+  /**
+   * Determines if deposit or withdraw service is enabled in SEP-6 or SEP-24.
+   *
+   * @param info The DepositWithdrawInfo containing the service details.
+   * @param service The operation to check, either "deposit" or "withdraw" (case-insensitive).
+   * @return true if the specified operation is enabled; false otherwise.
+   */
+  public boolean getIsServiceEnabled(DepositWithdrawInfo info, String service) {
```
philipliu, line 74: nit: Can we use enums instead of String for service, or split this into `isDepositEnabled` and `isWithdrawEnabled`?

api-schema/src/main/java/org/stellar/anchor/api/sep/operation/Sep38Info.java
```diff
@@ -16,12 +16,6 @@ public class Sep38Info {
 
   Integer decimals;
 
-  @SerializedName("sell_delivery_methods")
```
philipliu, line 19: How does this class relate to `org.stellar.anchor.api.sep.sep38.InfoResponse`? It looks like we are going to be returning that instead.

core/src/main/java/org/stellar/anchor/asset/AssetService.java
```diff
@@ -10,7 +12,7 @@ public interface AssetService {
    *
    * @return a list of assets.
    */
-  List<AssetInfo> listAllAssets();
+  List<AssetInfo> getAllAssets();
```
philipliu, line 15: nit: What about calling it `getAssets`?

core/src/main/java/org/stellar/anchor/asset/AssetService.java
```diff
@@ -35,12 +37,19 @@ public interface AssetService {
    * @param asset the SEP-38 asset identifier
    * @return an asset with the given SEP-38 asset identifier.
    */
-  AssetInfo getAssetByName(String asset);
+  AssetInfo getAssetById(String asset);
```
philipliu, line 40: I think the parameter name and Javadoc needs to be updated.

core/src/main/java/org/stellar/anchor/asset/AssetServiceValidator.java
```diff
@@ -5,43 +5,65 @@
...
+
+  private static void validateQuoteSupportedIfRequired(AssetInfo assetInfo)
+      throws InvalidConfigException {
+    // Validate SEP-31 fields
+    if (assetInfo.getSep31() != null && assetInfo.getSep31().getEnabled()) {
+      boolean isQuotesSupported = assetInfo.getSep31().isQuotesSupported();
+      boolean isQuotesRequired = assetInfo.getSep31().isQuotesRequired();
+      if (isQuotesRequired && !isQuotesSupported) {
+        throw new InvalidConfigException(
+            "if quotes_required is true, quotes_supported must also be true for asset: "
```
philipliu, line 53: Hmm... do we need `quotes_supported` then?

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
lijamie98: I would feel more comforable to move the method to the service classes because conversion is part of the business logic.

api-schema/src/main/java/org/stellar/anchor/api/asset/AssetInfo.java
lijamie98: What's the advantage of implementing this in the interface over adding the conversion function in the `AssetInfo.java`

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
JiahuiWho: Hmm I thought about it name after `Sep6AndSep24Info` but that sounds so weird...I would like to keep the naming match the config as much as possible

core/src/main/java/org/stellar/anchor/asset/AssetServiceValidator.java
JiahuiWho: This is also part of the https://stellarorg.atlassian.net/browse/ANCHOR-794 we will remove once confirm its not needed. This PR does not make modification to any existing fields/function/validation etc..

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
philipliu: I think that's actually better than `SendInfo. ' Let's call it that for now?

api-schema/src/main/java/org/stellar/anchor/api/sep/operation/Sep38Info.java
JiahuiWho: `InfoResponse` is a list of asset mapped from `Sep38Info`

api-schema/src/main/java/org/stellar/anchor/api/asset/StellarAssetInfo.java
JiahuiWho: Sure it will be named `ReceiveInfo` tho to match the config

essential-tests/build.gradle.kts
```diff
@@ -37,6 +37,6 @@ tasks { bootJar { enabled = false } }
 
 tasks.test {
   exclude("**/org/stellar/anchor/platform/*Test.class")
-  exclude("**/org/stellar/anchor/platform/integrationtest/**")
+//  exclude("**/org/stellar/anchor/platform/integrationtest/**")
```
philipliu, line 40: Meant to uncomment?

philipliu (approved):
LGTM, just a minor comment.

lijamie98 (approved):
LGTM

PR #1465, opened 2024-08-22
[ANCHOR-764] Add callback URL validation for custodial clients

philipliu:
### Description

This adds validation for callback URLs for custodial clients.

### Context

This was missed in the previous commit.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -201,6 +201,10 @@ event_processor:
 #   If the type is `noncustodial`,
 #   - domains: (required) the domain of the client.
 #   - callback_url: (optional) the URL of the client's callback API endpoint
+#   - callback_url_sep6: (optional) the URL of the client's SEP-6 callback API endpoint.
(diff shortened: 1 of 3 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -201,6 +201,10 @@ event_processor:
 #   If the type is `noncustodial`,
 #   - domains: (required) the domain of the client.
 #   - callback_url: (optional) the URL of the client's callback API endpoint
+#   - callback_url_sep6: (optional) the URL of the client's SEP-6 callback API endpoint.
```
lijamie98, line 204: How does the `callback_url` and `callback_url_sep6` and others interact? Can we document here?

philipliu: It's documented under the `custodial` client config. I decided not to repeat the full descriptions here since we don't do this for the other fields.

JiahuiWho (approved):
LGTM

PR #1466, opened 2024-08-23
[ANCHOR-783] Remove amount_out requirement in RequestOffchainFunds

philipliu:
### Description

`amount_out` is not set if a transaction has an associated indicative quote but has not yet received funds.

### Context

This was fixed for `RequestOnchainFunds` already.

### Testing

- `./gradlew test`

### Documentation

Will update stellar-docs

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -199,6 +199,7 @@ public class Sep6Service {
           exchangeAmountsCalculator.calculateFromQuote(
               request.getQuoteId(), sellAsset, request.getAmount());
     } else {
+      // TODO(philip): remove this
       // If a quote is not provided, set the fee and out amounts to 0.
       // The business server should use the request_offchain_funds RPC to update the amounts.
       amounts =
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -372,6 +373,7 @@ public StartWithdrawResponse withdrawExchange(
           exchangeAmountsCalculator.calculateFromQuote(
               request.getQuoteId(), sellAsset, request.getAmount());
     } else {
+      // TODO(philip): remove this
```
lijamie98, line 376: Why not remove?

philipliu: I will need to update the end-to-end tests which will take some time, but I want to this PR merged into the 2.9 release. The SEP-6 changes can be released in 2.9.1.

lijamie98: It would be nice to create a ticket for the end-to-end tests update. Otherwise, LGTM.

lijamie98 (approved):

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: Yep. I'll create the tickets. Thanks!

PR #1477, opened 2024-08-28
[ANCHOR-796] Fix NotifyCustomerInfoUpdated RPC call

philipliu:
### Description

This fixes the call to `NotifyCustomerInfoUpdated`. When the customer does not exist prior to SEP-6 transaction creation, the reference server needs to create one so that the ID is passed to the RPC. Otherwise, the `NotifyCustomerInfoUpdated` to defaults setting the status to `PENDING_ANCHOR`.

### Context

Demo wallet fix.

### Testing

- `./gradlew test`
- Tested deposit and withdrawal with demo wallet.

### Documentation

N/A

### Known limitations

We need to add tests that mirror the demo wallet flow.


lijamie98 (approved):
LGTM

PR #1483, opened 2024-08-30
[ANCHOR-794] Remove `fee_fixed`, `fee_percent` and `fields` from sep31 asset definition

lijamie98:
### Description

- Remove `fee_fixed`, `fee_percent` and `fields` from sep31 asset definition

### Context

- Simplify assets definition

### Testing

- `./gradlew test`
### Documentation

N/A

### Known limitations

More to implement for ANCHOR-794


platform/src/main/resources/config/anchor-asset-default-values.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+# The list of the assets that the anchor supports.
+#
+# Note: The `id` of each asset is the unique identifier for the asset which is formated as `schema:code:issuer`.
+#    - `schema`: is the schema of the asset. It can be either `stellar` or `iso4217`.
+#    - `code`: is the code of the asset.
+#    - `issuer`: is the issuer of the asset. It is only used for Stellar assets that are not native.
+# Currently, the anchor supports the following asset schemas:
+#   - `stellar`: The Stellar asset.
+#   - `stellar:native`: The native asset (XLM).
```
JiahuiWho, line 13: `native` is essentially under `stellar` scheme, let's not make it a separate item here? We can make an excpetion explanation about it  under the `stellar`.

platform/src/main/resources/config/anchor-asset-default-values.yaml
```diff
@@ -0,0 +1,155 @@
...
+#   - `iso4217`: The fiat asset.
+#
+assets:
+  # The Stellar asset that is not native (XLM).
+  - id: "stellar:"
+    # The `distribution_account` field is the Stellar account that the anchor will use to distribute the asset.
+    distribution_account:
+    # The `significant_decimals` field is the number of decimals that the asset supports. It must not be defined if the asset is XLM (stellar:native).
+    # default: 7
+    significant_decimals: 7
```
JiahuiWho, line 23: I see we are removing `sep38.decimals` and only use `significant_decimals`?
Maybe some comment about why's the default value

platform/src/main/resources/config/anchor-asset-default-values.yaml
lijamie98: SGTM. 

Removed.

platform/src/main/resources/config/anchor-asset-default-values.yaml
lijamie98: Typo.  Removed.
```
  #  default:7
```

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -600,11 +577,8 @@ private static Sep31InfoResponse sep31InfoResponseFromAssetInfoList(List<AssetIn
         AssetResponse assetResponse = new AssetResponse();
```
JiahuiWho, line 577: We can remove the following. They've been checked in AssetValidator 
 
```
        boolean isQuotesSupported = assetInfo.getSep31().isQuotesSupported();
        boolean isQuotesRequired = assetInfo.getSep31().isQuotesRequired();
        if (isQuotesRequired && !isQuotesSupported) {
          throw new SepValidationException(
              "if quotes_required is true, quotes_supported must also be true");
        }
```

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
```diff
@@ -600,11 +577,8 @@ private static Sep31InfoResponse sep31InfoResponseFromAssetInfoList(List<AssetIn
...
         assetResponse.setQuotesSupported(isQuotesSupported);
         assetResponse.setQuotesRequired(isQuotesRequired);
-        assetResponse.setFeeFixed(assetInfo.getSep31().getReceive().getFeeFixed());
-        assetResponse.setFeePercent(assetInfo.getSep31().getReceive().getFeePercent());
         assetResponse.setMinAmount(assetInfo.getSep31().getReceive().getMinAmount());
         assetResponse.setMaxAmount(assetInfo.getSep31().getReceive().getMaxAmount());
-        assetResponse.setFields(assetInfo.getSep31().getFields());
         response.getReceive().put(assetInfo.getCode(), assetResponse);
```
JiahuiWho, line 582: Can we directly use `Sep31Info` here instead of mapping it to `assetResponse`?  They contain the same info essentially.

core/src/test/resources/test_assets.yaml
```diff
@@ -157,24 +151,11 @@ assets:
...
-      fields:
-        transaction:
-          receiver_routing_number:
-            description: routing number of the destination bank account
-          receiver_account_number:
-            description: bank account number of the destination
-          type:
-            description: type of deposit to make
-            choices:
-              - SEPA
```
philipliu, line 176: How does the wallet specify the `type` now? SEP-6 defines a `method` configuration, do we plan to add this for SEP-31 in another PR?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: Although it saves coding, it is usually safer to separate the concern of the /info response and the assets configuration.

Making them the same can unintentionally break the SEP-31 `info` response when we change the implementation.

core/src/test/resources/test_assets.yaml
lijamie98: The `fields` is optional and deprecated in [SEP-31](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#asset-object-schema). 

How do we use the fields for SEP-6?

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: Good catch. Removed in two places.

core/src/main/java/org/stellar/anchor/sep31/Sep31Service.java
lijamie98: Per our discussion, I will create another PR to add `sep31.receive.methods`

core/src/test/resources/test_assets.yaml
lijamie98: Per our discussion, I will create another PR to add sep31.receive.methods
[ANCHOR-809](https://stellarorg.atlassian.net/browse/ANCHOR-809)

philipliu (approved):

PR #1490, opened 2024-09-04
[ANCHOR-795] Use flat list for client config

JiahuiWho:
### Update
Support reading from YAML and JSON strings which can be set through env vars

### Description
Use flat list for client config

1. Replace the `custodial` and `noncustodial` fields with `items` to hold a single list of clients.
2. Create a `TempClient` class specifically for loading inline YAML configurations, as Java property binding requires a concrete class for instantiation.
3. Modify `PropertyClientsConfig` to process the configuration source into a unified client list, which will be accessed by DefaultClientService.
4. Rename `CustodialClientConfig` and `NoncustodialClientConfig` to `CustodialClient` and `NoncustodialClient`, respectively

### Context

To align with asset config

### Testing

- `./gradlew test`


platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -37,82 +33,85 @@ public boolean supports(@NotNull Class<?> clazz) {
...
     }
-  }
 
-  void validateCustodialClients(List<CustodialClientConfig> clients, Errors errors) {
-    for (CustodialClientConfig client : clients) {
-      debugF("Validating custodial client {}", client);
-      if (client.getSigningKeys() == null || client.getSigningKeys().isEmpty()) {
+    // validate custodial client and noncustodial client
+    for (TempClient item : items) {
+      if (item.getType().equals(ClientConfig.ClientType.CUSTODIAL)) {
```
philipliu, line 55: If `item.getType` is null, this will throw an NPE. 

We could change this to the following to get around this:
```java
ClientConfig.ClientType.CUSTODIAL.equals(item.getType())
```

But should we validate that all clients have a `type` field before validating them instead?

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -37,82 +33,85 @@ public boolean supports(@NotNull Class<?> clazz) {
...
+          "invalid-noncustodial-client-config",
+          String.format("NonCustodial client %s must have at least one domain", client.getName()));
     }
+    validateCallbackUrls(client, errors);
+  }
+
+  void validateCallbackUrls(ClientConfig client, Errors errors) {
+    debugF("Validating client {}", client);
+    ImmutableMap.of(
+            "callback_url_sep6",
```
philipliu, line 90: nit: The strings should start with `callback_urls`

platform/src/test/kotlin/org/stellar/anchor/platform/configurator/ConfigManagerTest.kt
```diff
@@ -30,17 +30,17 @@ class ConfigManagerTest {
...
       "https://callback.vibrant.com/api/v2/anchor/callback/sep24"
     )
 
-    assertEquals(config.get("clients.noncustodial[1].name").value, "lobstr")
-    assertEquals(config.get("clients.noncustodial[1].domains[0]").value, "lobstr.co")
+    assertEquals(config.get("clients.items[1].name").value, "lobstr")
+    assertEquals(config.get("clients.items[1].domains[0]").value, "lobstr.co")
     assertEquals(
-      config.get("clients.noncustodial[1].callback_urls.sep6").value,
+      config.get("clients.items[1].callback_urls.sep6").value,
```
philipliu, line 43: Do we still support setting client configs with environment variables?

core/src/main/java/org/stellar/anchor/client/CustodialClient.java
```diff
@@ -8,7 +8,7 @@
 @Builder
 @AllArgsConstructor
 @NoArgsConstructor
-public class CustodialClientConfig implements ClientConfig {
+public class CustodialClient implements ClientConfig {
```
lijamie98, line 11: Not to confuse with the API client, The original name `CustodialClientConfig` will be less ambiguous and consistent with the `ClientConfig` naming.

lijamie98 (changes requested):

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -25,90 +19,83 @@
 @NoArgsConstructor
 public class DefaultClientService implements ClientService {
   static final Gson gson = GsonUtils.getInstance();
-  List<CustodialClientConfig> custodialClients = emptyList();
-  List<NonCustodialClientConfig> nonCustodialClients = emptyList();
+  List<CustodialClient> custodialClients = new ArrayList<>();
```
lijamie98, line 22: Is there advantages to use `new ArrayList<>()` instead of `emptyList()`?

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -25,90 +19,83 @@
...
   }
 
-  public static Map<String, Object> parseFileToMap(String filePath) throws InvalidConfigException {
+  public static List<TempClient> parseFileToList(String filePath) throws InvalidConfigException {
     try {
       String fileContent = FileUtil.read(Path.of(filePath));
       String fileExtension = FilenameUtils.getExtension(filePath).toLowerCase();
-      Map<String, Object> map;
+      Map<String, List<Object>> map;
       if ("yaml".equals(fileExtension) || "yml".equals(fileExtension)) {
```
lijamie98, line 40: nit: To be consistent with `enum ClientConfigType`, can we use `enum` for the supported file extensions.

core/src/main/java/org/stellar/anchor/client/TempClient.java
```diff
@@ -0,0 +1,54 @@
...
+ * inline YAML configuration. This class is intended as a temporary workaround to facilitate the
+ * deserialization of client configurations. Note: If you add any new fields to the CustodialClient
+ * or NonCustodialClient classes, ensure that you also add those fields here to maintain
+ * consistency.
+ */
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class TempClient {
```
lijamie98, line 18: naming suggestion: 
```suggestion
public class RawClientConfig {
```

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
```diff
@@ -1,17 +1,21 @@
 package org.stellar.anchor.config;
 
+import com.google.gson.annotations.SerializedName;
 import java.util.List;
-import org.stellar.anchor.client.CustodialClientConfig;
-import org.stellar.anchor.client.NonCustodialClientConfig;
+import org.stellar.anchor.client.TempClient;
 
 public interface ClientsConfig {
```
lijamie98, line 7: Further reviewed the code, I think we don't need to use `interface`. If the implementation complexity can be reduced, it would be ok to use `class PropertyClientsConfig` directly and remove the `ClientsConfig`.

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -1,34 +1,30 @@
...
-import org.stellar.anchor.client.CustodialClientConfig;
-import org.stellar.anchor.client.NonCustodialClientConfig;
+import org.stellar.anchor.client.*;
 import org.stellar.anchor.config.ClientsConfig;
+import org.stellar.anchor.util.GsonUtils;
 
 @Data
 public class PropertyClientsConfig implements ClientsConfig, Validator {
-  String type;
+  ClientsConfig.ClientsConfigType type;
```
lijamie98, line 24: nit: Add on-demand static imports
```suggestion
  ClientsConfigType type;
```

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
```diff
@@ -37,82 +33,85 @@ public boolean supports(@NotNull Class<?> clazz) {
 
   @Override
   public void validate(@NotNull Object target, @NotNull Errors errors) {
```
lijamie98, line 35: Unit tests should be implemented at `PropertyClientsConfigTests`.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -702,17 +704,15 @@ clients:
   #     allowAnyDestination: false
   #     destinationAccounts:
   #       - GDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD
```
lijamie98, line 706: nit: use a string that matches the length of a public key.  `GDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD`

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
```diff
@@ -25,90 +19,83 @@
...
         map = new Yaml().load(fileContent);
       } else if ("json".equals(fileExtension)) {
-        map = gson.fromJson(fileContent, new TypeToken<Map<String, Object>>() {}.getType());
+        map = gson.fromJson(fileContent, new TypeToken<Map<String, List<Object>>>() {}.getType());
       } else {
         throw new InvalidConfigException(
             String.format("%s is not a supported file format", filePath));
       }
-      return map;
+      map.get("items").removeIf(Objects::isNull);
```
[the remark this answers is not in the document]
JiahuiWho: like there's an trailing comma

platform/src/test/kotlin/org/stellar/anchor/platform/configurator/ConfigManagerTest.kt
JiahuiWho: My understanding is that users can always set these through environment variables as long as they use the correct format. This is because all file/inline configurations are ultimately read into the environment, and in client config case `clients.noncustodial[1].callback_urls.sep6` is how it is preserved in the environment variables.

core/src/main/java/org/stellar/anchor/client/DefaultClientService.java
JiahuiWho: emptyList is immutable.

core/src/main/java/org/stellar/anchor/client/CustodialClient.java
JiahuiWho: I tried to avoid using 'config' in the naming because these two classes represent client objects. They are different from the other `xxxConfig` classes, which primarily describe configuration settings

platform/src/test/kotlin/org/stellar/anchor/platform/configurator/ConfigManagerTest.kt
philipliu: I had trouble setting the callback URLs when it is nested under `callback_urls`. That's why in 2.9.0, I opted to make this configuration flat fields `callback_url_sep_*`. Can you test if it can be overridden using the `.env` file?

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
JiahuiWho: let me double check on that

core/src/test/resources/test_clients.yaml
```diff
@@ -1,12 +1,14 @@
-custodial:
+items:
```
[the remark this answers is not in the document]
JiahuiWho: I though it would look a little bit redundant when there are two clients nested together:
```
clients:
  type:
  clients:
```
I had a discussion with @lijamie98 that we might want use `items` for all the list configuration eventually.

core/src/main/java/org/stellar/anchor/client/CustodialClient.java
lijamie98: Examples of API clients are `Sep12Client` or `Sep24Client`.

platform/src/test/kotlin/org/stellar/anchor/platform/configurator/ConfigManagerTest.kt
JiahuiWho: Ok I know where went wrong...

When load config from env var, [`loadConfigFromEnv()`](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java#L45), specifically in [`ConfigEnvironment`](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/platform/src/main/java/org/stellar/anchor/platform/configurator/ConfigHelper.java#L76), changed the var name from `clients.items[0].abc` to `clients_items[0].abc`, and this makes it failed validation against config schema.
transformation is like `clients.items[0].abc` ->

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
JiahuiWho: Remove it will cause the circular dependency platform.main and core.main

platform/src/test/kotlin/org/stellar/anchor/platform/configurator/ConfigManagerTest.kt
philipliu: We can implement a workaround specifically for the callback URLs if a general solution is too difficult.

Although I'm ok with delaying inline support.

lijamie98: We have the plan to migrate from Spring `@ConfigurationProperty` to straight YAML parsing. [ANCHOR-803](https://stellarorg.atlassian.net/browse/ANCHOR-803)

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
JiahuiWho: It was tested in file `ClientsConfigTest`

philipliu (approved):

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
lijamie98: The `ClientsConfig` and `DefaultClientService` are not used in `core`, it is better to move them to `platform` if possible.

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
lijamie98: `ClientsConfigTest` should be renamed as `PropertyClientsConfigTests` for JUnit naming convention. 
Reference: https://www.baeldung.com/java-unit-testing-best-practices

core/src/main/java/org/stellar/anchor/client/TempClient.java
lijamie98: Any feedback?

JiahuiWho: I think the naming is fine. This object is for temp use. It will be removed once we done migration from Spring

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
JiahuiWho: I would prefer to do it in a separate PR as tho these also applies to several other files.

platform/src/main/java/org/stellar/anchor/platform/config/PropertyClientsConfig.java
JiahuiWho: I've been trying to follow the naming convention and keep it unified in the codebase. I don't quite get what's going wrong with test under the name of ClientsConfigTests?

core/src/main/java/org/stellar/anchor/config/ClientsConfig.java
lijamie98: That's no problem.

lijamie98 (approved):

PR #1491, opened 2024-09-04
[ANCHOR-757] Return GET /customer response on PUT /customer callback

philipliu:
### Description

This commit updates the response type of the `PUT /customer` callback to match `GET /customer`.

### Context

We will send the updated customer in a notification to the wallet server. Returning the updated customer as part of the callback lets us save a GET call.

### Testing

- `./gradlew test`

### Documentation

Callback API docs will be updated

### Known limitations

N/A


api-schema/src/main/java/org/stellar/anchor/api/callback/CustomerIntegration.java
```diff
@@ -17,7 +17,7 @@ public interface CustomerIntegration {
    * @return The GET customer response.
    * @throws AnchorException if error happens
    */
-  GetCustomerResponse getCustomer(GetCustomerRequest request) throws AnchorException;
+  CustomerResponse getCustomer(GetCustomerRequest request) throws AnchorException;
```
lijamie98, line 20: Although `PutCustomerResponse` and `GetCustomerResponse` have the same payload, combining the two classes seems confusing when reading the code. For example, the `Sep12GetCustomerRequest` and `Sep12PutCustomerRequest` gives the code better readability.

Is there better solution not to duplicate the code but keep the readability?

philipliu: I recreated the `PutCustomerReponse` class as a subclass of `GetCustomerResponse`.

lijamie98 (approved):
LGTM

PR #1493, opened 2024-09-05
[ANCHOR-794] Use anchor-asset-default-values.yaml as the prototype of asset objects

lijamie98:
### Description

- Use `anchor-asset-default-values.yaml` as the prototype of asset objects

While this is backward compatible, the values defined in the `anchor-asset-default-values.yaml` are the default values for the created AssetInfo objects.

### Context

Version `3.0.0` feature implementation.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/asset/DefaultAssetService.java
```diff
@@ -95,7 +132,9 @@ private static DefaultAssetService createDASFromMap(Map<String, List<Object>> ma
         throw new InvalidConfigException(String.format("Invalid asset: " + id));
       }
     }
-    AssetValidator.validate(das);
+    if (validate) {
```
JiahuiWho, line 135: What’s the case where we wouldn’t want to validate?

lijamie98: When loading the prototype yaml file (`anchor-asset-default-values.yaml`), the validation is off.

core/src/main/resources/config/anchor-asset-default-values.yaml
```diff
@@ -41,7 +41,7 @@ items:
       # The Configuration for the withdrawal operation of the SEP-6.
       withdraw:
         # `true`, if the asset is available for the withdrawal operations.
-        enabled: true
+        enabled: false
         # The minimum amount that the user can withdraw.
         min_amount: 0
         # The maximum amount that the user can withdraw.
```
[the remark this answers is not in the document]
lijamie98: I think we should not. Removed.

JiahuiWho (approved):

PR #1497, opened 2024-09-09
[ANCHOR-810] Fix the rate response validation bug when there is no fee

lijamie98:
### Description

- Fix the rate response validation bug when there is no fee
- The `withinRoundingError` call was missing the `!` operator when there is no `fee` in the rate response.

### Context

Bub fix.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
@@ -110,10 +110,6 @@ public class RestRateIntegration implements RateIntegration {
     }
 
     if (request.getType() == GetRateRequest.Type.FIRM) {
-      if (Objects.requireNonNull(rate).getFee() == null) {
-        logErrorAndThrow(
-            "'rate.fee' is missing in the GET /rate response", ServerErrorException.class);
-      }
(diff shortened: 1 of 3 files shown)


platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -110,10 +110,6 @@ void validateRateResponse(GetRateRequest request, GetRateResponse getRateRespons
     }
 
     if (request.getType() == GetRateRequest.Type.FIRM) {
-      if (Objects.requireNonNull(rate).getFee() == null) {
-        logErrorAndThrow(
-            "'rate.fee' is missing in the GET /rate response", ServerErrorException.class);
-      }
```
JakeUrban, line 116: Do we want to make the fee object optional?

JakeUrban: Its required in the SEP-38 responses

lijamie98: Do we assume `fee` is necessary in a quote?

JakeUrban (approved):
LGTM!

PR #1498, opened 2024-09-09
[ANCHOR-801] Add Kafka  SASL-SSL authentication support

lijamie98:
### Description

- Support Kafka SASL-SSL authentication
- Add `docker-compose-kafka-sasl-ssl.yaml.yaml` to run Kafka in SASL_SSL mode for testing.

### Context

- Answer the requests to support SASL-SSL

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -681,15 +681,22 @@ events:
...
-      #     `PLAIN`: PLAIN SASL mechanism
-      sasl_mechanism:
+      # The SASL mechanism used for authentication and must match what is used on the broker.
+      # `sasl_mechanism` can be one of the following supported mechanisms:
+      #   `PLAIN`: PLAIN SASL mechanism
+      #   `SCRAM-SHA-256`: SCRAM-SHA-256 SASL mechanism
+      #   `SCRAM-SHA-512`: SCRAM-SHA-512 SASL mechanism
+      #   `GSSAPI`: GSSAPI SASL mechanism
+      #   `OAUTHBEARER`: OAUTHBEARER SASL mechanism
+      # ` AWS_MSK_IAM`: AWS MSK IAM SASL mechanism
```
philipliu, line 697: Do we support these?

service-runner/src/main/resources/dev-tools/docker-compose-kafka-sasl.yaml
```diff
@@ -0,0 +1,46 @@
+version: '3'
+services:
+  zookeeper:
```
philipliu, line 3: Do we need Zookeeper to test the new auth methods?

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
```diff
@@ -0,0 +1,79 @@
...
+      sslEngine.setUseClientMode(true);
+      return sslEngine;
+    } catch (NoSuchAlgorithmException | KeyManagementException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public SSLEngine createServerSslEngine(String peerHost, int peerPort) {
+    return null;
```
philipliu, line 51: Should this be implemented?

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
```diff
@@ -0,0 +1,79 @@
...
+import java.security.cert.X509Certificate;
+import java.util.Map;
+import java.util.Set;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import org.apache.kafka.common.security.auth.SslEngineFactory;
+
+public class TrustAllSslEngineFactory implements SslEngineFactory {
```
philipliu, line 16: Is this what Anchors are expected to run in production? Should we trust all certificates?

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
```diff
@@ -204,7 +206,22 @@ void configureAuth(Properties props) throws InvalidConfigException {
...
+    switch (kafkaConfig.getSecurityProtocol()) {
+      case SASL_PLAINTEXT:
+        props.put(SECURITY_PROTOCOL_CONFIG, "SASL_PLAINTEXT");
+        props.put(SASL_MECHANISM, kafkaConfig.getSaslMechanism().getValue());
         break;
+      case SASL_SSL:
+        props.put(SECURITY_PROTOCOL_CONFIG, "SASL_SSL");
+        props.put(SASL_MECHANISM, kafkaConfig.getSaslMechanism().getValue());
+        // Disable hostname verification
+        props.put("ssl.endpoint.identification.algorithm", "");
```
philipliu, line 222: Should we enable `HTTPS`?

philipliu (commented):
Should we add tests for the new auth mechanisms?

service-runner/src/main/resources/dev-tools/docker-compose-kafka-sasl.yaml
lijamie98: This is dev tools and not used in the pipeline. Kraft configuration is tricky and was not working well with SASL_SSL auth. May need to spend more time to replace zookeeper with `kraft`.

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
lijamie98: This only affects `Kafka`. I don't think we need `HTTPS`.

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
lijamie98: I thought about this. By not trusting all certificate, we need to configure `keystore` and `truststore`, mount the stores, configure the passwords, etc. It will make the configuration very difficult.

If there is a request from the anchors, we may add the configurations to read the CA certs from the file.

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
lijamie98: No, we are using this class for clients only. If we use this class for server accepting connections, we may add this method.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: The values are defined in the enum. We don't have to support the implementation. The Kafka client library implements these mechanisms.

philipliu: IAM-specific auth methods need additional properties configured https://github.com/aws/aws-msk-iam-auth?tab=readme-ov-file#configuring-a-kafka-client-to-use-aws-iam-with-aws_msk_iam-mechanism

There may be specific properties additionally required for the other SASL mechanisms.

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
philipliu: This is a security risk, right? I think we may want to trust all certificates in our tests, but defaulting to trust all certificates seems incorrect to me. I see that we have the Keystore and the cert will be checked in as part of this PR, if we are not using them, should we delete them?

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
philipliu: Is there a reason not to enable this? This seems like an extra layer of security. Or at least make it a configurable property?

lijamie98: Does `Kafka` use `HTTPS`?

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
lijamie98: We can delete those files. But they are good resources if we want to do the kafka auth tests in the `extended-tests`.

lijamie98: It will be quite hard to provide the stores and pass the passwords through the secret env vars. 

In most cases, the client and the kafka server are running in the same subnet and I think we can avoid the hassel of the keytore configurations.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Keep only `PLAIN`

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
philipliu: Yes, the default is `http` https://docs.confluent.io/platform/current/security/authentication/mutual-tls/overview.html#optional-settings

lijamie98: Ok. Removed it to leave the default to `https`

platform/src/main/java/org/stellar/anchor/platform/utils/TrustAllSslEngineFactory.java
philipliu: Don't we just need to pass the keystone and truststore locations in the properties?

lijamie98: We will enable the certificate verification after discussion.

platform/src/main/java/org/stellar/anchor/platform/config/PropertySecretConfig.java
```diff
@@ -21,6 +21,9 @@ public class PropertySecretConfig implements SecretConfig {
       "secret.events.queue.kafka.username";
   public static final String SECRET_EVENTS_QUEUE_KAFKA_PASSWORD =
       "secret.events.queue.kafka.password";
+  public static final String SECRET_SSL_KEYSTORE_PASSWORD = "secret.ssl.keystore.password";
```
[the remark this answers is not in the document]
lijamie98: I would prefer leave this here because this may be used for other purposes in the future.

service-runner/src/main/resources/dev-tools/secrets/jaas.conf
```diff
@@ -0,0 +1,15 @@
...
+  username="admin"
+  password="admin-secret"
+  user_admin="admin-secret"
+  serviceName="kafka"
+  user_alice="alice-secret";
+};
+Client {};
+
+KafkaClient {
+    org.apache.kafka.common.security.scram.ScramLoginModule required
```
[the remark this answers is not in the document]
lijamie98: This was for testing purposes. Removed.

service-runner/src/main/resources/profiles/kafka-sasl-ssl/config.env
```diff
@@ -0,0 +1,9 @@
+events.queue.kafka.security_protocol=SASL_SSL
+events.queue.kafka.ssl_keystore_location=C:\\projects\\java-stellar-anchor-sdk\\service-runner\\src\\main\\resources\\dev-tools\\secrets\\kafka.keystore.jks
```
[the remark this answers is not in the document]
lijamie98: Switched to `classpath://` which is more robust to access the resource file.

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
```diff
@@ -33,17 +37,33 @@
...
+  String sslTruststoreLocation;
 
-  KafkaSession(KafkaConfig kafkaConfig, String sessionName, EventQueue queue) {
+  KafkaSession(KafkaConfig kafkaConfig, String sessionName, EventQueue queue) throws IOException {
     this.kafkaConfig = kafkaConfig;
     this.sessionName = sessionName;
     this.topic = queue.name();
+
+    // If the keystore and truststore files exist, use them, otherwise, use the resources
+    if (new File(kafkaConfig.getSslKeystoreLocation()).exists()) {
```
philipliu: Is the default value of `SslKeyStoreLocation`, null? Do we need a null check here?

philipliu (approved):

platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
lijamie98: Added "SASL_SSL" check. If "SASL_SSL" is set, the keystore locations must be set. 

[ANCHOR-824](https://stellarorg.atlassian.net/browse/ANCHOR-824) is created for better validation.

PR #1501, opened 2024-09-11
[ANCHOR-799] Remove `platform_server.auth`

philipliu:
### Description

This PR simplifies the authentication configuration for the Platform API. Currently, this is configured in two places.

- `platform_server.auth`: This enables request authentication on the Platform server.
- `platform_api.auth`: This determines what is included in the HTTP header of the Platform API requests.

We are removing `platform_server.auth` in favor of using `platform_api.auth` on the server and request building sides.

### Context

These configurations are intertwined, making it confusing to configure auth. They will always need to be set to the same thing, and we should not duplicate it.

### Testing

- `./gradlew test`

### Documentation

Stellar docs PR will be created

### Known limitations

N/A


lijamie98 (approved):
LGTM

PR #1511, opened 2024-09-23
[ANCHOR-829] Allow `rate.fee.total` to be zero

JiahuiWho:
### Description

As title

### Context

Unblock chipper when no fee is applied to transaction

### Testing

- `./gradlew test`


philipliu (approved):

platform/src/main/java/org/stellar/anchor/platform/callback/RestRateIntegration.java
```diff
@@ -156,9 +155,9 @@ void validateRateResponse(GetRateRequest request, GetRateResponse getRateRespons
     // if fee is set, check the following
     if (fee != null) {
       // fee.total is present and is a positive number
-      if (!isPositiveNumber(fee.getTotal())) {
+      if (!isNonNegativeNumber(fee.getTotal())) {
```
Ifropc, line 158: Adding test for this behavior would be nice

Ifropc (approved):
LGTM with small comment

PR #1519, opened 2024-09-24
[ANCHOR-834] Add transaction_id to callback customer API when request provides it

lijamie98:
### Description

-  Add transaction_id to callback customer API when request provides it

### Context

- When the business server only implements the `GET /custemer` with `transaction_id`, it is necessary for the SEP-12 to pass the `transaction_id`.

### Testing

- `./gradlew test`
Added more tests to `Sep12ServiceTest.kt`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
+++ b/core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
@@ -190,13 +190,11 @@ class Sep12ServiceTest {
         )
         .build()
     every { platformApiClient.getTransaction(any()) } returns transaction
-    val jwtToken = createJwtToken(TEST_ACCOUNT)
 
-    val putRequestBase =
-      Sep12PutCustomerRequest.builder().transactionId(TEST_TRANSACTION_ID).build()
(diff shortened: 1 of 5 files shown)


core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
```diff
@@ -327,6 +328,75 @@ class Sep12ServiceTest {
     assertEquals(TEST_ACCOUNT, mockPutRequest.account)
   }
 
+  @Test
+  fun `Test put customer request with transaction_id`() {
```
philipliu, line 332: nit: We can combine this and the above the test into a `ParameterizedTest` on `PutCustomerRequest` since the only difference between them is the `transaction_id`.

philipliu (approved):
LGTM

core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
lijamie98: Good idea.

Ifropc (approved):

PR #1525, opened 2024-09-26
[ANCHOR-819] Update client config default value examples

JiahuiWho:
### Description

It was agreed offline that we will not use prototype loading for the clients configuration, as this configuration is typically empty.

Thus only updating the default client configuration example to offer clearer guidance on how to set it correctly.

### Testing

- `./gradlew test`

### DIFF
--- a/platform/src/main/resources/config/anchor-client-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-client-default-values.yaml
@@ -1,17 +1,22 @@
 #
 # This file contains client configuration details. The provided values are for illustrative purposes only.
 #
-
 items:
   # Custodial Client Configuration
     # (Required) A unique identifier for the client, typically a name that clearly represents the client entity.
   - name:
     # (Required) The type of the client. It can be either `custodial` or `noncustodial`.
     type: custodial
-    # (Required) A list of public keys used for SEP-10 authentication.
+    # (Required) A list of public keys used for SEP-10 authentication. Example:
+    # signing_keys:
+    #   - GDAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+    #   - GDBYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY
     signing_keys: []
     # (Optional) The URLs to which the service can send callbacks for different SEP types. Optional due to some wallets may opt
-    # to poll instead, or may use polling first before implementing callbacks at a later stage.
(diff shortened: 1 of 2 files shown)


platform/src/main/resources/config/anchor-client-default-values.yaml
```diff
@@ -1,39 +1,38 @@
...
+  # Noncustodial Client Configuration
+    # (Required) A unique identifier for the client, similar to custodial clients.
+  - name:
+    # (Required) The type of the client, similar to custodial clients.
     type: noncustodial
-    # The domains associated with the client, used for verifying the client's identity.
-    domains:
-      - wallet-server:8092
+    # (Required) A list of domains associated with the client, used to verify the client’s identity.
+    domains: []
```
lijamie98, line 32: Examples in comments will help.

platform/src/main/resources/config/anchor-client-default-values.yaml
```diff
@@ -1,39 +1,38 @@
...
+    signing_keys: []
+    # (Optional) The URLs to which the service can send callbacks for different SEP types. Optional due to some wallets may opt
     # to poll instead, or may use polling first before implementing callbacks at a later stage.
-    callbackUrls:
-      sep6: https://example.com/callbacks/sep6
-      sep24: https://example.com/callbacks/sep24
-      sep31: https://example.com/callbacks/sep31
-      sep12: https://example.com/callbacks/sep12
-    # A boolean flag that, when set to true, allows any destination for deposits. Defaults to false,
+    callback_urls:
```
lijamie98, line 15: Can you add the examples to the comments?

platform/src/main/resources/config/anchor-client-default-values.yaml
```diff
@@ -1,39 +1,38 @@
...
+  - name:
+    # (Required) The type of the client. It can be either `custodial` or `noncustodial`.
     type: custodial
-    # The public keys used for SEP-10 authentication.
-    signing_keys:
-      - GDADDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD
-      - GDBDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD
-    # The URLs to which the service can send callbacks for different SEP types. Optional due to some wallets may opt
+    # (Required) A list of public keys used for SEP-10 authentication.
+    signing_keys: []
```
lijamie98, line 12: Can we have the examples in the comments?

lijamie98 (approved):

PR #1528, opened 2024-09-27
[ANCHOR-836] Add configuration `event.queue.kafka.ssl_cert_verify` 

lijamie98:
### Description

- Add configuration `event.queue.kafka.ssl_cert_verify` 

### Context

This flag is for dev and test environment where a valid CA may not be present. 

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -702,6 +702,10 @@ events:
       # `sasl_mechanism` can be one of the following supported mechanisms:
       #   `PLAIN`: PLAIN SASL mechanism
       sasl_mechanism: PLAIN
+      # The configuration to enable SSL certificate verification. If true, the SSL certificate will be verified.
+      # For dev and test environments, it can be used to disable SSL certificate verification.
+      # For production environments, it is strongly recommended to set to true to avoid man-in-the-middle attach
+      ssl_verify_cert: true
       # The SSL keystore location. The file path to the keystore file (jks).
       ssl_keystore_location:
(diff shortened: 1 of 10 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -702,6 +702,10 @@ events:
       # `sasl_mechanism` can be one of the following supported mechanisms:
       #   `PLAIN`: PLAIN SASL mechanism
       sasl_mechanism: PLAIN
+      # The configuration to enable SSL certificate verification. If true, the SSL certificate will be verified.
+      # For dev and test environments, it can be used to disable SSL certificate verification.
+      # For production environments, it is strongly recommended to set to true to avoid man-in-the-middle attach
```
philipliu, line 707: typo: attach should be attack

service-runner/src/main/resources/version-info.properties
```diff
@@ -1 +1 @@
-version=2.10.2
\ No newline at end of file
+version=2.10.2R
```
philipliu, line 1: What is the `R`?

lijamie98: Thanks for catching the accidental key stroke.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Thanks. Fixed.

philipliu (approved):

JiahuiWho (approved):

PR #1529, opened 2024-09-27
Add reference server ingress

philipliu:
Adds an ingress resource for the reference server. The reference server needs to be publicly reachable for the SEP-24 flows.


lijamie98 (approved):
LGTM

PR #1530, opened 2024-09-28
Create demo environment for testing

philipliu:


lijamie98:
The config name `demo` is generic. Can we change to `m24-demo`?

PR #1554, opened 2024-10-25
[ANCHOR-858] Do not require keystore and truststore if ssl_cert_verify is false

lijamie98:
### Description

Do not search for the `keystore` and `truststore` files if `ssl_cert_verify` is to `false`. 

When ` ssl_cert_verify` is set to `false`, the `KafkaSession` still look for the file and cause IOException. 

### Context

Bug fix.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
@@ -269,6 +269,6 @@ public class KafkaSession implements EventService.Session {
   }
 
   String find(String sslKeystoreLocation) throws IOException {
-    return findFileThenResource(kafkaConfig.getSslKeystoreLocation()).getAbsolutePath();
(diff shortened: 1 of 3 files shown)


platform/src/main/java/org/stellar/anchor/platform/event/KafkaSession.java
```diff
@@ -258,4 +267,8 @@ void configureAuth(Properties props) throws InvalidConfigException {
         throw new IllegalStateException("Unexpected value: " + kafkaConfig.getSecurityProtocol());
     }
   }
+
+  String find(String sslKeystoreLocation) throws IOException {
+    return findFileThenResource(kafkaConfig.getSslKeystoreLocation()).getAbsolutePath();
```
JiahuiWho, line 272: nit. Should be the passed in param? And let's remove the `keyStore` from naming

JiahuiWho (approved):

PR #1559, opened 2024-10-30
[ANCHOR-790] Add deployment test profile

philipliu:
### Description

This adds a `deployment` test profile that points the SEP-24 interactive UI to the reference UI so that the SEP-24 flows can be completed via the demo walllet.

### Context

The deployments will be updated so that the SEP-24 interactive UI is deployed. This PR updates the docker-compose setup so it also works locally.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


lijamie98:
Please also add the test to `extended-tests`.

philipliu:
> Please also add the test to `extended-tests`.

Which test are we referring to? This is different than the other test profiles in that integration tests are not meant to run against it.

lijamie98:
The test can be added the workflow similar to https://github.com/lijamie98/java-stellar-anchor-sdk/blob/main/.github/workflows/sub_extended_tests.yml#L49

service-runner/src/main/resources/profiles/deployment/config.env
```diff
@@ -0,0 +1,62 @@
+# secrets
+secret.data.username=postgres
```
lijamie98, line 2: The configuration values are stacked on top of the [default/config.env](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/service-runner/src/main/resources/profiles/default/config.env). 

That means, only values not defined in the `default/config.env` are needed.

philipliu: Fixed!

lijamie98 (approved):
LGTM

PR #1569, opened 2024-11-04
[ANCHOR-865] Add missing columns on ref server start

philipliu:
### Description

This updates the reference server initialization to create new columns in the model but not in the database. It also updates the SELECT query builder to use the new method.

### Context

`customers.bank_name` does not exist, causing SEP-6 transactions to fail.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
+++ b/kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
@@ -20,12 +20,17 @@ interface CustomerRepository {
 
 class JdbcCustomerRepository(private val db: Database) : CustomerRepository {
   init {
-    transaction(db) { SchemaUtils.create(Customers) }
+    transaction(db) {
+      SchemaUtils.create(Customers)
+      val missingColumnsStatements = SchemaUtils.addMissingColumnsStatements(Customers)
+      missingColumnsStatements.forEach { exec(it) }
+    }
   }
 
   override fun get(id: String): Customer? =
     transaction(db) {
-        Customers.select { Customers.id.eq(id) }
+        Customers.selectAll()
+          .where { Customers.id.eq(id) }
           .mapNotNull {
             Customer(
               id = id,
@@ -99,7 +104,8 @@ class JdbcCustomerRepository(private val db: Database) : CustomerRepository {
         }
       }
     return transaction(db) {
-        Customers.select { query }
+        Customers.selectAll()
+          .where { query }
           .mapNotNull {
             Customer(
               id = it[Customers.id],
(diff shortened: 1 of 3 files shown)


lijamie98:
How did the reference server work if the column was missing in local dev?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
```diff
@@ -20,12 +20,17 @@ interface CustomerRepository {
...
+      SchemaUtils.create(Customers)
+      val missingColumnsStatements = SchemaUtils.addMissingColumnsStatements(Customers)
+      missingColumnsStatements.forEach { exec(it) }
+    }
   }
 
   override fun get(id: String): Customer? =
     transaction(db) {
-        Customers.select { Customers.id.eq(id) }
+        Customers.selectAll()
```
lijamie98, line 32: Why not `SELECT * WHERE id=?`?

This seems inefficient.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
```diff
@@ -99,7 +104,8 @@ class JdbcCustomerRepository(private val db: Database) : CustomerRepository {
         }
       }
     return transaction(db) {
-        Customers.select { query }
+        Customers.selectAll()
```
lijamie98, line 107: Same as above

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/TransactionKYCRepository.kt
```diff
@@ -18,12 +18,17 @@ interface TransactionKYCRepository {
 
 class JdbcTransactionKYCRepository(private val db: Database) : TransactionKYCRepository {
   init {
-    transaction(db) { SchemaUtils.create(TransactionKYCs) }
+    transaction(db) {
+      SchemaUtils.create(TransactionKYCs)
+      val missingColumnsStatements = SchemaUtils.addMissingColumnsStatements(TransactionKYCs)
```
lijamie98, line 23: Does this happen when we run the newer version of AP on the previous version of the DB?

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
philipliu: The old query builder is deprecated. This is the new way of writing the same query.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/TransactionKYCRepository.kt
philipliu: Yeah, that's right.

philipliu:
> How did the reference server work if the column was missing in local dev?

The database is recreated frequently in local dev whereas the development has a long-running Postgres instance.

kotlin-reference-server/src/main/kotlin/org/stellar/reference/dao/CustomerRepository.kt
lijamie98: The new design of the `exposed` looks weird to me.

lijamie98 (approved):
LGTM

PR #1575, opened 2024-11-11
Bump java-stellar-sdk to v1.0.0

overcat:
### Description

Bump java-stellar-sdk to v1.0.0

### Context

Due to the upcoming 1.0.0 containing many breaking changes, I created this PR to make the process a bit easier.
### Testing

- `./gradlew test`

### Documentation

https://github.com/lightsail-network/java-stellar-sdk/blob/master/CHANGELOG.md

### Known limitations

 N/A


overcat:
Wait for the wallet SDK update before continuing this PR. https://github.com/stellar/kotlin-wallet-sdk/pull/155

gradle/libs.versions.toml
```diff
@@ -27,7 +28,7 @@ jakarta-annotation-api = "3.0.0"
 jakarta-transaction-api = "2.0.1"
 jakarta-validation-api = "3.1.0"
 jakarta-xml-bind-api = "4.0.2"
-java-stellar-sdk = "0.44.1"
+java-stellar-sdk = "1.0.0-beta1"
```
lijamie98, line 31: This should be `1.0.0-rc0`

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarPaymentObserver.java
```diff
@@ -148,8 +148,8 @@ SSEStream<OperationResponse> startSSEStream() {
         new EventListener<>() {
           @Override
           public void onEvent(OperationResponse operationResponse) {
-            if (operationResponse.getTransaction().isPresent()) {
-              metricLatestBlockRead.set(operationResponse.getTransaction().get().getLedger());
+            if (operationResponse.getTransaction() != null) {
```
lijamie98, line 151: `getTransaction()` returns `Optional`.

```suggestion
            if (operationResponse.getTransaction().isPresent()) {
```

lijamie98:
Thanks for submitting this PR.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarPaymentObserver.java
overcat: Previously, some fields in the response were checked using null, while others were checked using Optional. Now we uniformly check for existence using null.

gradle/libs.versions.toml
overcat: I have already updated it to rc0. I expect to release the stable version of the stellar sdk next week. At that time, we can release the new wallet sdk and then update this repo.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarPaymentObserver.java
lijamie98: I think this introduces further inconsistency. The `getTransaction()` returns `Optional`. If we want to check the existence using `null`, the `getTransaction()` should return `null` / value instead of `Optional`. 

However, I don't think it is right to make the return value uniform. Whether a function returns `null` or `Optional` should depend on each function's definition. 

I think for functions returning `Optional` should be checked by `isPresent()`. Otherwise,  `!=null`.

overcat: Hi @lijamie98, I apologize for not mentioning earlier that in the latest version, `getTransaction()` no longer returns `Optional`. You can check the current code here: https://github.com/lightsail-network/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/responses/operations/OperationResponse.java#L55

Let me explain the rationale behind removing `Optional`:

- Since we're using `lombok`, there's no need to manually write get functions anymore. Not using `Optional` helps keep things simpler.
- The existence of response fields is heavily dependent on the Horizon version. For instance, a field might exist in Horizon 1.0 but be removed in Horizon 2.0. We want users to have the flexibility to determine field existence based on their specific Horizon version. This way, we avoid tight coupling between the SDK version and Horizon versions.

lijamie98:
@overcat Is the PR still in progress?

overcat:
Hi @lijamie98, we need to wait for the [wallet-sdk](https://central.sonatype.com/artifact/org.stellar/wallet-sdk) to release a new version. Maybe you can coordinate that?

lijamie98:
> Hi @lijamie98, we need to wait for the [wallet-sdk](https://central.sonatype.com/artifact/org.stellar/wallet-sdk) to release a new version. Maybe you can coordinate that?

I will. I am also working on the wallet-sdk which is pending our ops to add GH secrets to automate the release process. ETA next week.

overcat:
Hi @lijamie98, the code encountered a problem while running CI, but after I reran CI, there was no issue. If possible, could you help me take a look? I'm not very familiar with this area.
https://github.com/stellar/anchor-platform/actions/runs/12385701275/job/34572520338?pr=1575#step:11:695

lijamie98:
> Hi @lijamie98, the code encountered a problem while running CI, but after I reran CI, there was no issue. If possible, could you help me take a look? I'm not very familiar with this area.
> https://github.com/stellar/anchor-platform/actions/runs/12385701275/job/34572520338?pr=1575#step:11:695

Some of the tests in `essential-tests` can sometimes fail. The root cause seems like the message queue delivery being not stable. The problem usually goes away after re-run. We have a ticket in the backlog for fixing the essential tests instability.

PR #1581, opened 2024-11-20
[ANCHOR-809][SEP-31] Add funding_method to SEP-31

JiahuiWho:
### Description

- Add `funding_methods` to `/info` response. This field is mandatory for the anchor. 
- Add `funding_method` to `POST /transaction` param. This field is mandatory for wallet.

### Context

SEP-31 changes scoped in https://github.com/stellar/stellar-protocol/pull/1567

### Testing

- `./gradlew test`

### Next
The SEP-6 change will be handled in a separate PR.

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/asset/Sep31Info.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/asset/Sep31Info.java
@@ -1,6 +1,7 @@
 package org.stellar.anchor.api.asset;
 
 import com.google.gson.annotations.SerializedName;
+import java.util.List;
 import java.util.Map;
 import lombok.Data;
 
@@ -23,6 +24,8 @@ public class Sep31Info {
 
     @SerializedName("max_amount")
     Long maxAmount;
+
+    List<String> methods;
   }
 
   @Data
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
@@ -34,8 +34,12 @@ public class Sep31PostTransactionRequest {
   @SerializedName("receiver_id")
   String receiverId;
 
-  Sep31TxnFields fields;
+  @Deprecated Sep31TxnFields fields;
+
   String lang;
+
+  @SerializedName("funding_method")
+  String fundingMethod;
 
   @Data
   @AllArgsConstructor
--- a/core/src/main/java/org/stellar/anchor/util/SepHelper.java
+++ b/core/src/main/java/org/stellar/anchor/util/SepHelper.java
@@ -4,6 +4,7 @@ import static org.stellar.anchor.api.sep.SepTransactionStatus.*;
 import static org.stellar.anchor.util.MathHelper.decimal;
 
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import org.stellar.anchor.api.exception.AnchorException;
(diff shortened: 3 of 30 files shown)


api-schema/src/main/java/org/stellar/anchor/api/asset/Sep31Info.java
```diff
@@ -23,6 +24,8 @@ public static class ReceiveOperation {
 
     @SerializedName("max_amount")
     Long maxAmount;
+
+    List<String> methods;
```
philipliu, line 28: Do we call this `methods` here so that it's consistent with the SEP-6 asset configuration?

api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
```diff
@@ -34,9 +34,13 @@ public class Sep31PostTransactionRequest {
   @SerializedName("receiver_id")
   String receiverId;
 
-  Sep31TxnFields fields;
+  @Deprecated Sep31TxnFields fields;
```
philipliu, line 37: This is going into 3.0, right? Can we remove the field instead?

core/src/main/java/org/stellar/anchor/util/SepHelper.java
```diff
@@ -137,6 +138,28 @@ public static void validateAmountLimit(String messagePrefix, String amount, Long
...
+   * Validates whether a specified funding method is supported for a given asset.
+   *
+   * @param assetID the unique id of the asset being validated.
+   * @param method the funding method to validate.
+   * @param supportedMethods a list of funding methods supported for the asset.
+   * @throws BadRequestException if the provided funding method is not in the list of supported
+   *     methods.
+   */
+  public static void validateFundingMethod(
+      String assetID, String method, List<String> supportedMethods) throws AnchorException {
```
philipliu, line 151: Can we throw `BadRequestException` instead of `AnchorException`?

api-schema/src/main/java/org/stellar/anchor/api/asset/Sep31Info.java
JiahuiWho: Ah actually this is to match the naming is asset config, for data binding
```
    sep31:
      enabled: false
      receive:
        min_amount: 0
        max_amount: 1000000
        methods:
          - SEPA
          - SWIFT
```

api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
JiahuiWho: Yes this is going to 3.0 but no we cannot remove it. I tried to do so in the Protocol change but Jake insisit that we should keep it and continue support, and so far there isn't a plan to get rid of it :/

philipliu: I believe there are a number of changes going into 3.0 that are deprecated in the protocol but we decided to remove in 3.0. wdyt @lijamie98

philipliu:
By the way, `develop` is now the 3.0 branch, we should be merging into that.

api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
JiahuiWho: While I’d love to remove all those long-deprecated fields, I don’t think we can get everyone on the same page right now with the 3.0 release so close.

A more practical approach might be to create a plan after the 3.0 release, outlining the removal timeline, the specific fields to be deprecated, and a guide for identifying their replacements or alternatives.

philipliu: Sounds good. Let's merge this change as is and discuss whether we want to remove these fields in 3.x or 4.0.

philipliu (approved):

api-schema/src/main/java/org/stellar/anchor/api/sep/sep31/Sep31PostTransactionRequest.java
JiahuiWho: tracking in https://stellarorg.atlassian.net/browse/ANCHOR-888

PR #1582, opened 2024-11-20
[ANCHOR-809][SEP-6] Add funding_method to SEP-6

JiahuiWho:
### Description

- In `/info` response, replace `fields`(in deposit) and `types`(in withdraw) with `funding_methods` 
- In all 4 transaction creation request, replace `type` with `funding_method`, and this param will be mandatory

### Context

SEP-31 changes scoped in https://github.com/stellar/stellar-protocol/pull/1567

### Testing

- `./gradlew test`

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/InfoResponse.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/InfoResponse.java
@@ -1,6 +1,7 @@
 package org.stellar.anchor.api.sep.sep6;
 
 import com.google.gson.annotations.SerializedName;
+import java.util.List;
 import java.util.Map;
 import lombok.Builder;
 import lombok.Data;
@@ -77,13 +78,16 @@ public class InfoResponse {
     @SerializedName("max_amount")
     Long maxAmount;
 
+    @SerializedName("funding_methods")
+    List<String> fundingMethods;
+
     /**
      * The fields required to initiate a deposit.
      *
      * <p>The only field supported by the platform is <code>type</code>. Additional fields required
      * for KYC are supplied asynchronously through SEP-12 requests.
      */
-    Map<String, AssetInfo.Field> fields;
+    @Deprecated Map<String, AssetInfo.Field> fields;
   }
 
   /** Withdrawal configuration. */
@@ -109,6 +113,10 @@ public class InfoResponse {
     @SerializedName("max_amount")
     Long maxAmount;
 
+    /** The maximum amount that can be withdrawn. */
+    @SerializedName("funding_methods")
+    List<String> fundingMethods;
+
     /**
      * The types of withdrawal methods supported and their fields.
      *
@@ -116,7 +124,7 @@ public class InfoResponse {
      * account and KYC information is supplied asynchronously through PATCH requests and SEP-12
      * requests respectively.
      */
-    Map<String, WithdrawType> types;
+    @Deprecated Map<String, WithdrawType> types;
   }
 
   /** Withdrawal type configuration. */
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartDepositExchangeRequest.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartDepositExchangeRequest.java
@@ -49,8 +49,12 @@ public class StartDepositExchangeRequest {
   /** The memo to use for the deposit. */
   String memo;
 
-  /** Type of deposit. */
-  @NonNull String type;
+  /** Deposit method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** Type of deposit. Deprecated in favor of funding_method */
+  @Deprecated String type;
 
   /**
    * Defaults to en if not specified or if the specified language is not supported. Currently,
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawExchangeRequest.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawExchangeRequest.java
@@ -39,8 +39,12 @@ public class StartWithdrawExchangeRequest {
   /** The amount of the source asset the user would like to withdraw. */
   @NonNull String amount;
 
-  /** The type of withdrawal to make. */
-  @NonNull String type;
+  /** Withdraw method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** The type of withdrawal to make. Deprecated in favor of funding_method */
+  @Deprecated String type;
 
   /** The ISO 3166-1 alpha-3 code of the user's current address. */
   @SerializedName("country_code")
--- a/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawRequest.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawRequest.java
@@ -20,8 +20,12 @@ public class StartWithdrawRequest {
   @NonNull
   String assetCode;
 
-  /** Type of withdrawal. */
-  String type;
+  /** Withdraw method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** Type of withdrawal. Deprecated in favor of funding_method */
+  @Deprecated String type;
 
   /** The account to withdraw from. */
   String account;
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -6,7 +6,6 @@ import static org.stellar.anchor.util.AssetHelper.isWithdrawEnabled;
 import static org.stellar.anchor.util.MemoHelper.*;
 import static org.stellar.anchor.util.SepLanguageHelper.validateLanguage;
 
-import com.google.common.collect.ImmutableMap;
 import io.micrometer.core.instrument.Counter;
 import java.time.Instant;
 import java.util.*;
@@ -542,7 +541,7 @@ public class Sep6Service {
                 .authenticationRequired(true)
                 .minAmount(asset.getSep6().getDeposit().getMinAmount())
                 .maxAmount(asset.getSep6().getDeposit().getMaxAmount())
-                .fields(ImmutableMap.of("type", type))
+                .fundingMethods(methods)
                 .build();
 
         response.getDeposit().put(asset.getCode(), deposit);
@@ -551,10 +550,6 @@ public class Sep6Service {
 
       if (isWithdrawEnabled(asset.getSep6())) {
         List<String> methods = asset.getSep6().getWithdraw().getMethods();
-        Map<String, WithdrawType> types = new HashMap<>();
-        for (String method : methods) {
-          types.put(method, WithdrawType.builder().fields(new HashMap<>()).build());
-        }
 
         WithdrawAssetResponse withdraw =
             WithdrawAssetResponse.builder()
@@ -562,7 +557,7 @@ public class Sep6Service {
                 .authenticationRequired(true)
                 .minAmount(asset.getSep6().getWithdraw().getMinAmount())
                 .maxAmount(asset.getSep6().getWithdraw().getMaxAmount())
-                .types(types)
+                .fundingMethods(methods)
                 .build();
 
         response.getWithdraw().put(asset.getCode(), withdraw);
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -106,10 +106,11 @@ public class Sep6Service {
     }
 
     StellarAssetInfo asset = requestValidator.getDepositAsset(request.getAssetCode());
-    if (request.getType() != null) {
-      requestValidator.validateTypes(
-          request.getType(), asset.getCode(), asset.getSep6().getDeposit().getMethods());
-    }
+    String fundingMethod =
+        request.getFundingMethod() != null ? request.getFundingMethod() : request.getMemoType();
+    requestValidator.validateTypes(
+        fundingMethod, asset.getCode(), asset.getSep6().getDeposit().getMethods());
+
     if (request.getAmount() != null) {
       requestValidator.validateAmount(
           request.getAmount(),
@@ -185,8 +186,10 @@ public class Sep6Service {
     }
 
     StellarAssetInfo buyAsset = requestValidator.getDepositAsset(request.getDestinationAsset());
+    String fundingMethod =
+        request.getFundingMethod() != null ? request.getFundingMethod() : request.getType();
     requestValidator.validateTypes(
-        request.getType(), buyAsset.getCode(), buyAsset.getSep6().getDeposit().getMethods());
+        fundingMethod, buyAsset.getCode(), buyAsset.getSep6().getDeposit().getMethods());
     requestValidator.validateAmount(
         request.getAmount(),
         buyAsset.getCode(),
@@ -280,10 +283,11 @@ public class Sep6Service {
     }
 
     StellarAssetInfo asset = requestValidator.getWithdrawAsset(request.getAssetCode());
-    if (request.getType() != null) {
-      requestValidator.validateTypes(
-          request.getType(), asset.getCode(), asset.getSep6().getWithdraw().getMethods());
-    }
+    String fundingMethod =
+        request.getFundingMethod() != null ? request.getFundingMethod() : request.getType();
+    requestValidator.validateTypes(
+        fundingMethod, asset.getCode(), asset.getSep6().getWithdraw().getMethods());
+
     if (request.getAmount() != null) {
       requestValidator.validateAmount(
           request.getAmount(),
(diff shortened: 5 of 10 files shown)


core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
```diff
@@ -11,16 +11,8 @@ class Sep6ServiceTestData {
...
-                        "type": {
-                            "description": "type of deposit to make",
-                            "choices": [
-                                "SEPA",
-                                "SWIFT"
-                            ],
-                            "optional": false
-                        }
-                    }
+                    "funding_methods": ["SEPA", "SWIFT"]
```
philipliu, line 14: Does it make sense to support the `type` request parameter when the `/info` response returns `funding_methods`?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -357,8 +360,10 @@ public StartWithdrawResponse withdrawExchange(
     }
 
     StellarAssetInfo sellAsset = requestValidator.getWithdrawAsset(request.getSourceAsset());
+    String fundingMethod =
+        request.getFundingMethod() != null ? request.getFundingMethod() : request.getType();
     requestValidator.validateTypes(
-        request.getType(), sellAsset.getCode(), sellAsset.getSep6().getWithdraw().getMethods());
+        fundingMethod, sellAsset.getCode(), sellAsset.getSep6().getWithdraw().getMethods());
```
philipliu, line 366: Do we not need to persist this field in the database?

platform/src/main/java/org/stellar/anchor/platform/controller/sep/Sep6Controller.java
```diff
@@ -90,7 +92,8 @@ public StartDepositResponse depositExchange(
       @RequestParam(value = "account") String account,
       @RequestParam(value = "memo_type", required = false) String memoType,
       @RequestParam(value = "memo", required = false) String memo,
-      @RequestParam(value = "type") String type,
+      @RequestParam(value = "funding_method", required = false) String fundingMethod,
```
philipliu, line 95: In this PR's description and the protocol, `funding_method` is a required parameter. Is it supposed to be required?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
JiahuiWho: In db it's stilled saved as `type`. I'm not gonna rename it for now (at least in this pr) as type is a preserved word, and it's causing a lot of issue when I was trying.

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
JiahuiWho: The plan is to support `funding_method` while deprecating `type`. Same as discussion https://github.com/stellar/java-stellar-anchor-sdk/pull/1581#discussion_r1850640966

platform/src/main/java/org/stellar/anchor/platform/controller/sep/Sep6Controller.java
JiahuiWho: It 's not marked as required because we need to support both `type` and `funding_method` and they are referencing the same thing

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/InfoResponse.java
```diff
@@ -77,13 +78,16 @@ public static class DepositAssetResponse {
...
+    List<String> fundingMethods;
+
     /**
      * The fields required to initiate a deposit.
      *
      * <p>The only field supported by the platform is <code>type</code>. Additional fields required
      * for KYC are supplied asynchronously through SEP-12 requests.
      */
-    Map<String, AssetInfo.Field> fields;
+    @Deprecated Map<String, AssetInfo.Field> fields;
```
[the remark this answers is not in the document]
JiahuiWho: Can you point me to where it's missing ? I believe I have marked it as deprecated

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/InfoResponse.java
```diff
@@ -109,14 +113,18 @@ public static class WithdrawAssetResponse {
...
+
     /**
      * The types of withdrawal methods supported and their fields.
      *
      * <p>The platform does not allow fields to be configured for withdrawal methods. Financial
      * account and KYC information is supplied asynchronously through PATCH requests and SEP-12
      * requests respectively.
      */
-    Map<String, WithdrawType> types;
+    @Deprecated Map<String, WithdrawType> types;
```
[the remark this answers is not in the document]
JiahuiWho: Can you point me to where it's missing ? I believe I have marked it as deprecated

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartDepositExchangeRequest.java
```diff
@@ -49,8 +49,12 @@ public class StartDepositExchangeRequest {
...
   String memo;
 
-  /** Type of deposit. */
-  @NonNull String type;
+  /** Deposit method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** Type of deposit. Deprecated in favor of funding_method */
+  @Deprecated String type;
```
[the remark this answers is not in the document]
JiahuiWho: Can you point me to where it's missing ? I believe I have marked it as deprecated

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawExchangeRequest.java
```diff
@@ -39,8 +39,12 @@ public class StartWithdrawExchangeRequest {
...
   @NonNull String amount;
 
-  /** The type of withdrawal to make. */
-  @NonNull String type;
+  /** Withdraw method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** The type of withdrawal to make. Deprecated in favor of funding_method */
+  @Deprecated String type;
```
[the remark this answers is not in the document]
JiahuiWho: Can you point me to where it's missing ? I believe I have marked it as deprecated

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/StartWithdrawRequest.java
```diff
@@ -20,8 +20,12 @@ public class StartWithdrawRequest {
...
   String assetCode;
 
-  /** Type of withdrawal. */
-  String type;
+  /** Withdraw method used for the transaction. */
+  @SerializedName("funding_method")
+  String fundingMethod;
+
+  /** Type of withdrawal. Deprecated in favor of funding_method */
+  @Deprecated String type;
```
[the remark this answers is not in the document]
JiahuiWho: Can you point me to where it's missing ? I believe I have marked it as deprecated

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
philipliu: Removing the fields from the `/info` response is effectively removing support for `type`. If the client already supports `funding_methods` would they still send `type` in the transaction requests?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: If only `funding_method` is specified in the request, will the `type` field in the database get updated? It doesn't look like `Sep6TransactionBuilder `  call is updated.

JiahuiWho:
```
    String fundingMethod =
        request.getFundingMethod() != null ? request.getFundingMethod() : request.getType();
    requestValidator.validateTypes(
        fundingMethod, asset.getCode(), asset.getSep6().getDeposit().getMethods());
```
and we pass this calculated `fundingMethod` to `Sep6TransactionBuilder.type()`

philipliu: It looks `type` is still set to `request.getType()` on line 43. Can you point to me where it's passed into type, maybe I'm missing something?

JiahuiWho: My bad.

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
philipliu: > fields is only deprecated. Anchor can still use it if they prefer.
> We removed fields and type from all existing examples because we want to advocate the use of funding_methods for v3

I understand that we are deprecating `fields` in 3.0 but this PR is removing `fields` from the info response which means clients will need update their code to read `funding_methods` from the info response.

https://github.com/stellar/java-stellar-anchor-sdk/blob/3203bb469260bda710aaffb3704c9ebc28a80b7b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java#L544-L551

In this case, I don't see why client would continue sending `type` in their transaction requests. I think if the goal is to deprecate `fields` and `type`, it makes sense to continue returning `fields` in the info response.

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
philipliu: It would be great if we can include a unit test that tests that `funding_method` and `type` are persisted to the DB.

philipliu: Also, `type` is present in the platform transaction response. Should we rename that to `funding_method`?

api-schema/src/main/java/org/stellar/anchor/api/sep/sep6/InfoResponse.java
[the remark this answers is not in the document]
lijamie98:
```suggestion
    /**
     * Deprecated since SEP-6 version 4.1.0.
     * https://github.com/stellar/stellar-protocol/blob/21d30cd22caa6f25369828289d845c4366eaa641/ecosystem/sep-0006.md
     */
    @Deprecated Map<String, AssetInfo.Field> fields;
```

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
JiahuiWho: The platform transaction response is closely tied to platformTransactionData, and I prefer to avoid changing the name of the data column

api-schema/src/main/java/org/stellar/anchor/api/platform/GetTransactionResponse.java
```diff
@@ -12,4 +12,6 @@
  */
 @SuperBuilder
 @NoArgsConstructor
-public class GetTransactionResponse extends PlatformTransactionData {}
+public class GetTransactionResponse extends PlatformTransactionData {
+  String fundingMethod;
```
philipliu, line 16: Is there any reason not to add this field directly to the `PlatformTransactionData` class? Also, do we plan to remove the `type` field from that class as we no longer set it?

core/src/main/java/org/stellar/anchor/util/TransactionMapper.java
```diff
@@ -137,6 +137,7 @@ public static GetTransactionResponse toGetTransactionResponse(
         .sep(PlatformTransactionData.Sep.SEP_6)
         .kind(PlatformTransactionData.Kind.from(txn.getKind()))
         .status(SepTransactionStatus.from(txn.getStatus()))
+        .fundingMethod(txn.getType())
```
philipliu, line 140: Should this be `getFundingMethod`?

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTest.kt
```diff
@@ -172,6 +172,36 @@ class Sep6ServiceTest {
...
+        .amount("100")
+        .build()
+    sep6Service.deposit(token, request)
+
+    // Verify validations
+    verify(exactly = 1) { requestValidator.getDepositAsset(TEST_ASSET) }
+    verify(exactly = 1) { requestValidator.validateAccount(TEST_ACCOUNT) }
+
+    // Verify effects
+    verify(exactly = 1) { txnStore.save(any()) }
```
philipliu, line 198: Can we add an assertion on the transaction saved to the database? Updating all the tests in this class may be easier, replacing `type` with `funding_method` and adding a few additional tests for just `type`. WDYT?

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
```diff
@@ -11,8 +11,17 @@ class Sep6ServiceTestData {
...
-       
+                    "funding_methods": ["SEPA", "SWIFT"],
+                    "fields": {
+                        "type": {
+                            "description": "type of deposit to make",
+                            "choices": [
+                                "SEPA",
+                                "SWIFT"
+                            ],
+                            "optional": false
```
philipliu, line 22: Should `optional` be true now that we accept `funding_method` as well?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -107,10 +107,11 @@ public StartDepositResponse deposit(Sep10Jwt token, StartDepositRequest request)
     }
 
     StellarAssetInfo asset = requestValidator.getDepositAsset(request.getAssetCode());
-    if (request.getType() != null) {
-      requestValidator.validateTypes(
-          request.getType(), asset.getCode(), asset.getSep6().getDeposit().getMethods());
-    }
+    String fundingMethod =
+        request.getFundingMethod() != null ? request.getFundingMethod() : request.getType();
```
philipliu, line 111: Should we reject requests with `funding_method` and `type` specified?

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -107,10 +107,11 @@ public StartDepositResponse deposit(Sep10Jwt token, StartDepositRequest request)
...
+    requestValidator.validateTypes(
```
philipliu, line 112: Does this throw an exception if `fundingMethod` is empty?

JiahuiWho: yes

core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
JiahuiWho: I wouldn't worry about that. If a wallet is using `funding_method ` I believe they have read the SEP and it's clearly stated that `type` is deprecated in favor of `funding_method`.

api-schema/src/main/java/org/stellar/anchor/api/platform/GetTransactionResponse.java
JiahuiWho: Why do we want to add an extra field to `PlatformTransactionData ` ? Since they are referencing the same thing I though the best way is to rename it.

But since the field has a bad name `type` I'm not gonna touch it here. Will try in a separate PR

core/src/main/java/org/stellar/anchor/util/TransactionMapper.java
JiahuiWho: This is sepTxn from db. It'a still saved as `type`

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTest.kt
JiahuiWho: This seems to complicate things unnecessarily. If the goal is simply to test whether `funding_method` or `type` are processed correctly, I can extract the new logic into a helper function and test from there.

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTestData.kt
JiahuiWho: Can you help clarify what "optional" means in this context? Who is the intended audience, and what is it optional for?

Is it for anchor that this is optional to config? Or is it for wallet to optional past in some request?

philipliu: This field tells the wallet that `type` is an optional parameter for the deposit and withdrawal requests.

core/src/main/java/org/stellar/anchor/util/TransactionMapper.java
philipliu: I thought so but wasn't sure because on line 98, `fundingMethod` is set to `txn.getFundingMethod()`.

core/src/test/kotlin/org/stellar/anchor/sep6/Sep6ServiceTest.kt
philipliu: We should at least have a test that `funding_method` is persisted in the database. Whether that's done by refactoring into a helper method or replacing all the `.type` with `.fundingMethod`, I'll leave it to you. In my opinion, replacing all the `type` calls would be easier.

core/src/main/java/org/stellar/anchor/util/TransactionMapper.java
JiahuiWho: That's for SEP-31, which doesn't have `type` originally, and I added `fundingMethod `in https://github.com/stellar/java-stellar-anchor-sdk/pull/1581

philipliu (approved):

PR #1585, opened 2024-12-02
[ANCHOR-592] Add schema to database configuration

lijamie98:
### Description

- Add `schema` configuration
- Bump version to `2.12.0`

### Context

- configuration improvemetns

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -782,8 +782,8 @@ data:
   server:
 
   # The name of the database.
-  # For `sqlite` and `aurora`, this is the path to the database file.
-  # For `postgres`, this is the name of the database.
+  # For `sqlite`, this is the path to the database file.
+  # For `postgres` and `aurora`, this is the name of the database.
   database:
 
   # The schema of the database.
--- a/platform/src/main/resources/config/anchor-config-default-values.yaml
+++ b/platform/src/main/resources/config/anchor-config-default-values.yaml
@@ -781,8 +781,15 @@ data:
   # The hostname and port of the database server.
   server:
 
-  # Name of the database.
+  # The name of the database.
+  # For `sqlite` and `aurora`, this is the path to the database file.
+  # For `postgres`, this is the name of the database.
   database:
+
+  # The schema of the database.
+  # For `postgres` and `aurora`, default is `public`
+  # For `sqlite`, this is ignored.
+  schema:
 
   # Initial number of connections
   # For `sqlite`, set this to 1 to avoid database file lock exception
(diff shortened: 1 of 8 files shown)


service-runner/src/main/resources/profiles/default/config.env
```diff
@@ -27,6 +27,7 @@ custody_server.base_url=http://custody-server:8086
 data.type=postgres
 data.server=db:5432
 data.database=postgres
+data.schema=anchor-platform
```
lijamie98, line 30: This is for default testing configuration and will not impact the backward compatibility.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -781,9 +781,16 @@ data:
   # The hostname and port of the database server.
   server:
 
-  # Name of the database.
+  # The name of the database.
+  # For `sqlite` and `aurora`, this is the path to the database file.
```
philipliu, line 785: This should be the name of the database for `aurora`.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -781,9 +781,16 @@ data:
...
+  # For `postgres`, this is the name of the database.
   database:
 
+  # The schema of the database.
+  # For `postgres` and `aurora`, default is `public`
+  # For `sqlite`, this is ignored.
+  schema:
```
philipliu, line 792: What is the expected behavior of setting `schema` when tables have already been created under public?

lijamie98: The new set of tables will be created under the specified `schema`. If they set the `schema` to `public`, the existing tables will be used.

philipliu: What happens to in-progress SEP transactions? You can no longer query them since they are in the public schema, right?

lijamie98: When leaving as empty, backward compatibility will be preserved.

philipliu: Do we need to warn users that setting this while having tables in the public schema will break in-progress transactions?

lijamie98: We can. But this is like setting the different database may result in a different set of tables. I feel it is part of common sense. I am feeling very strong of adding or not adding the comments. WDYT?

philipliu: It's better to be explicit in my opinion.

PR #1608, opened 2025-01-07
[ANCHOR-935] Remove SEP-dependent Sep24DepositInfoGenerator bean creation

lijamie98:
### Description

- Remove SEP-dependent Sep24DepositInfoGenerator bean creation

### Context

- To fix a bug reported at: https://discord.com/channels/897514728459468821/961080541711568966/1324777127152713849

### Testing

- `./gradlew test`

### Documentation

N/A

- Attach stellar-docs pull request, documenting new feature
- If it's an urgent feature request, please create a ticket and attach ticket number to this PR

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
@@ -94,7 +94,6 @@ public class PlatformServerBeans {
   }
 
   @Bean
-  @ConditionalOnAnySepsEnabled(seps = {"sep24"})
   Sep24DepositInfoGenerator sep24DepositInfoGenerator(
       Sep24Config sep24Config, Optional<CustodyApiClient> custodyApiClient)
       throws InvalidConfigException {


JiahuiWho:
Can you add some explanation on why is this?

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
```diff
@@ -94,7 +94,6 @@ Sep31DepositInfoGenerator sep31DepositInfoGenerator(
   }
 
   @Bean
-  @ConditionalOnAnySepsEnabled(seps = {"sep24"})
```
philipliu, line 97: Instead of removing this annotation, is it possible to mark the deposit info generators as nullable in `RequestOnchainFundsHandler` bean?

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
```diff
@@ -117,7 +116,7 @@ Sep24DepositInfoGenerator sep24DepositInfoGenerator(
   }
 
   @Bean
-  @ConditionalOnAnySepsEnabled(seps = {"sep6"})
+  @OnAnySepsEnabled(seps = {"sep6"})
```
philipliu, line 119: Why is this only removed for SEP-24? Wouldn't this also affect SEP-31 and SEP-6?

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
lijamie98: Thought about the possibility of marking the direct injected components as nullable. However, this can create another can of worms because of the DI dependency complexity.

The `@ConditionalOnAnySepsEnabled` was designed to disable the SEP endpoint. When we extend it to optionally instantiate the beans, it creates another set of problems.

I kind of strongly hesitate to mark DI components optional.

platform/src/main/java/org/stellar/anchor/platform/component/platform/PlatformServerBeans.java
lijamie98: We should do it for SEP6 and 31.

lijamie98: SEP-6 and SEP-31 condition annotation removed.

PR #1624, opened 2025-01-24
[ANCHOR-916] Put Horizon behind interface

lijamie98:
### Description

- Put Horizon class behind the `LedgerApi` interface

### Context

- Stellar RPC integration

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
```diff
@@ -0,0 +1,111 @@
...
+  List<OperationResponse> getStellarTxnOperations(String stellarTxnId);
+
+  /**
+   * Submit a transaction to the network.
+   *
+   * @param transaction
+   * @return The transaction response.
+   * @throws NetworkException
+   */
+  TransactionResponse submitTransaction(Transaction transaction) throws NetworkException;
```
philipliu, line 50: `TransactionResponse` is Horizon's `submitTransaction` response. RPC's `sendTransaction` returns a `SendTransactionResponse` which only contains the transaction hash and its status. We will need to call `getTransaction` to fetch the transaction afterward. The schemas are not the same. Should we create a new type to encapsulate the transaction submission responses?

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
```diff
@@ -0,0 +1,111 @@
...
+import lombok.Getter;
+import lombok.Value;
+import org.stellar.sdk.KeyPair;
+import org.stellar.sdk.Transaction;
+import org.stellar.sdk.TransactionBuilderAccount;
+import org.stellar.sdk.exception.NetworkException;
+import org.stellar.sdk.responses.TransactionResponse;
+import org.stellar.sdk.responses.operations.OperationResponse;
+
+public interface LedgerApi {
```
philipliu, line 15: SEP-45 uses transaction simulation which is only available in RPC. How do we plan to expose RPC-specific methods in the interface?

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
```diff
@@ -0,0 +1,111 @@
...
+   */
+  Account getAccount(String account) throws NetworkException;
+
+  /**
+   * Get the operations for the given Stellar transaction.
+   *
+   * @param stellarTxnId The Stellar transaction ID.
+   * @return The operations for the transaction.
+   */
+  List<OperationResponse> getStellarTxnOperations(String stellarTxnId);
```
philipliu, line 41: `OperationResponse` is Horizon's `getOperation` response. If we use RPC, we must parse the operations from the transaction envelope. Should we create a new type to encapsulate the operation response?

core/src/main/java/org/stellar/anchor/ledger/StellarRpc.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public TransactionResponse submitTransaction(Transaction transaction) throws NetworkException {
+    // TODO: Implement this method
+    return null;
+  }
+
+  private AccountEntry fetchAccountEntry(String account) throws IOException {
+    // TODO: Implement this method
+    return null;
+//    GetLedgerEntriesResponse response;
```
philipliu, line 57: Test code?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -71,4 +115,9 @@ public List<OperationResponse> getStellarTxnOperations(String stellarTxnId) {
         .execute()
         .getRecords();
   }
+
+  @Override
+  public TransactionResponse submitTransaction(Transaction transaction) throws NetworkException {
```
philipliu, line 120: Do we only submit transactions in tests? It should not be part of the interface if that's the case.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
```diff
@@ -0,0 +1,111 @@
...
+      sequenceNumber = seqNum;
+    }
+
+    @Override
+    public Long getIncrementedSequenceNumber() {
+      return sequenceNumber + 1;
+    }
+
+    /** Increments sequence number in this object by one. */
+    public void incrementSequenceNumber() {
```
philipliu, line 78: How is this used?

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: In SEP-45, it should work directly with the implementation class instead of LedgerApi.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: I think it is fine to leave this function if it is only used in tests.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: This is required by the `TransactionBuilderAccount` interface.

core/src/main/java/org/stellar/anchor/ledger/StellarRpc.java
lijamie98: This is to put the Horizon behind the interface. Will add tests after the `StellarRpc` implementation in a different PR.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: And Horizon option (`horizon_url`) will cause conflict and fail to start during validation of `sep45.enabled` is `true`.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  public LedgerTransaction getTransaction(String transactionId) throws NetworkException {
+    TransactionResponse response = getServer().transactions().transaction(transactionId);
+    return LedgerTransaction.builder()
+        .hash(response.getHash())
+        .sourceAccount(response.getSourceAccount())
+        .envelopeXdr(response.getEnvelopeXdr())
+        .metaXdr(response.getResultMetaXdr())
```
philipliu, line 107: Do we need these fields?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -0,0 +1,145 @@
...
+        .sourceAccount(response.getSourceAccount())
+        .memo(response.getMemo())
+        .sequenceNumber(response.getSourceAccountSequence())
+        .createdAt(response.getCreatedAt())
+        .build();
+  }
+
+  @Override
+  public LedgerTransactionResponse submitTransaction(Transaction transaction)
```
philipliu, line 116: Do we need to implement a `submitTransaction` method if we only submit transactions in tests? I don't think we use the `Horizon` class today. Can we remove this method until we need to submit transactions? Otherwise, we have to deal with the maintenance burden of a method we don't use.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * Get payment operations for a transaction.
+   *
+   * @param stellarTxnId the transaction id
+   * @return the operations
+   * @throws NetworkException request failed, see {@link PaymentsRequestBuilder#execute()}
+   */
+  public List<OperationResponse> getStellarTxnOperations(String stellarTxnId) {
```
philipliu, line 137: Is the `OperationResponse` specific to Horizon's API?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -0,0 +1,145 @@
...
+                assert creditAsset != null;
+                return creditAsset.getCode().equals(assetCode)
+                    && creditAsset.getIssuer().equals(assetIssuer);
+              }
+              return false;
+            });
+  }
+
+  @Override
+  public Account getAccount(String account) throws NetworkException {
```
philipliu, line 61: Can we only return the account signers in this method? The `balances` field is already covered by the `hasTrustlines` method. Implementing this RPC would mean making two `GetLedgerEntries` calls when most of the time, we would only care about signers or trustlines/balances but not both.

platform/src/main/java/org/stellar/anchor/platform/observer/ObservedPayment.java
```diff
@@ -42,6 +45,91 @@ public class ObservedPayment {
   String transactionMemoType;
   String transactionEnvelope;
 
+  public static ObservedPayment from(LedgerTransaction txn, LedgerPaymentOperation paymentOp)
```
philipliu, line 48: When we move to unified events in the observer, do we plan to construct `LedgerPaymentOperation` and `LedgerPathPaymentOperation` from the `transfer` events?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
philipliu: Also, should this method be defined in the interface?

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Value;
+import org.stellar.sdk.KeyPair;
+import org.stellar.sdk.Transaction;
+import org.stellar.sdk.TransactionBuilderAccount;
+import org.stellar.sdk.exception.NetworkException;
+
+public interface LedgerApi {
```
philipliu, line 13: nit: Can we call this something else? Maybe `StellarNetwork`?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: yes, we do. [ObservedPayment.java](https://github.com/stellar/anchor-platform/blob/develop/platform/src/main/java/org/stellar/anchor/platform/observer/ObservedPayment.java#L77)

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: IIRC, we discussed about having submission function that may be required if we are expanding the observer to txn submitter.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: Yes, the `OperationResponse` is specific to Horizon's API only. The observer's logics for RPC and Horizon are very different. It won't fit in the interface.

platform/src/main/java/org/stellar/anchor/platform/observer/ObservedPayment.java
lijamie98: The transfer event can be very different in RPC and Horizon. In StellarRPC, it is most likely we are going to need this. Ther e is a [Jira ticket](https://stellarorg.atlassian.net/jira/software/c/projects/ANCHOR/boards/47/backlog?epics=visible&issueParent=20700&selectedIssue=ANCHOR-977).

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: I don't like `LedgerApi` and `StellarNetwork` which conflicts with many other things though. But if to let me choose, by looking at the functions, I would prefer `LedgerApi`.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
philipliu: Do we use `resultMetaXdr`?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
philipliu: I don't think we have this planned right now, so I think it makes sense to hold off on the implementation until then.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
philipliu: Is the plan to use the interface in the observer then?

platform/src/main/java/org/stellar/anchor/platform/observer/ObservedPayment.java
philipliu: I guess my point is that because they are so different, I think we shouldn't refactor the observer to use the `LedgerPaymentOperation` abstraction until we know what the unified events solution looks like in AP.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
philipliu: What about `StellarClient`?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: This class is based on RPC ledger endpoint. It returns both `signers` and `trustlines`. 

The `GetLedgerEntries` takes a `List<LedgerKey>`. This is only one call though.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: Do you have a better one?....

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: But the `hasTrustline` should be a helper function instead of a interface.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
philipliu: `StellarNetworkAdapter` or `StellarNetworkClient`?

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: we currently don't. Considering reducing the maintainance load, I will remove

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: The implementation was there and the tests needs it. We need submit for both RPC and horizon for testing.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: Yes, if possible. But they can be different.

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
lijamie98: I added balance because it was used somewhere in the code.

platform/src/main/java/org/stellar/anchor/platform/observer/ObservedPayment.java
lijamie98: The `ObservedPayment` is also used in the RPC handler module. The other Jira ticket will take care of that.

core/src/main/java/org/stellar/anchor/ledger/LedgerApi.java
lijamie98: `StellarClient` is too close to others. `StellarNetworkAdapter` or `StellarNetworkClient` are quite close.

The functions are 
- `getTransaction`
- `getAccount (signers, trustlines, balance(optional))`
- `submitTransaction`

I think `LedgerApi` is not perfect but close.

How about `LedgerClient`?

lijamie98:
Let's move the discussion to https://github.com/stellar/anchor-platform/pull/1658/files

PR #1628, opened 2025-02-12
[ANCHOR-708][PD] Add sep24-reference-ui helm charts

JiahuiWho:
### Description

1. Add sep24-reference-ui helm charts
2. Add reference server ingress
3. Fix interactive url jwt secrets mismatch 
4. Update minikube scripts

### Testing

- `./gradlew test`
Test with demo wallet and transaction complete


helm-charts/sep24-reference-ui/templates/configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sep24-reference-ui-config
+  namespace: {{ .Values.namespace }}
+data:
+  BUSINESS_SERVER_ENDPOINT: "http://localhost:8091"
```
philipliu, line 7: This should use the Kubernetes hostname of the reference server. It looks like `[srv-name].[namespace].svc.cluster.local`.

helm-charts/sep24-reference-ui/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Values.fullName }}-svc-{{ .Values.services.ref.name }}
```
philipliu, line 4: Maybe we can call the configuration block `ref-ui` instead of `ref`? It doesn't affect the behavior but the reference server already calls itself `ref` in its values file.

helm-charts/sep24-reference-ui/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+  namespace: {{ .Values.namespace }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.fullName }}-sep24-reference-ui
```
philipliu, line 7: Can we parameterize `sep24-reference-ui`? Not just in this spot but everywhere. We do so for the other services.

helm-charts/sep24-reference-ui/templates/ingress.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ .Values.ingress.name }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.fullName }}-ingress
```
philipliu, line 6: The same comments as the reference server ingress about the annotations and the ingress name.

helm-charts/sep24-reference-ui/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: Service
+metadata:
+  name: {{ .Values.fullName }}-svc-{{ .Values.services.ref.name }}
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{- if .Values.services.ref.lsabels }}
+      {{- range $key, $value := .Values.services.ref.labels }}
+        {{ $key }}: {{ $value | quote }}
+      {{- end }}
+    {{- end }}
```
philipliu, line 11: It looks like we are missing some labels here compared to the other `service.yaml`s.

helm-charts/README.md
```diff
@@ -34,13 +34,23 @@ helm upgrade --install reference-server ./reference-server/ -f ./reference-serve
 # Install the Anchor Platform
 helm upgrade --install anchor-platform ./sep-service/ -f ./sep-service/values.yaml
 
+# Build and Install SEP-24 Reference UI
+docker build -t sep24-reference-ui:local ../../sep24-reference-ui
```
philipliu, line 38: Can we use the image we publish to Dockerhub? Developers likely would not need to update the UI so using the latest image is fine. Using a local build complicates the local deployment process.

helm-charts/sep24-reference-ui/templates/deployment.yaml
JiahuiWho: Will just use `ui`. Hyphen (-) is not allowed in Helm template variable names unless properly quoted.

helm-charts/reference-server/templates/ingress.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ .Values.ingress.name }}
+  labels:
+    app.kubernetes.io/name: {{ .Values.fullName }}-ingress
```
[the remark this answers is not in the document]
JiahuiWho: I'm  a little bit confused .. Each of the 3 service have their own values file with fullname being `anchor-platform` / `reference-server` / `sep24-reference-ui` respectively. How will the resources end up with duplicate names?

And actually I have ingress.name specified in the values file, below is out put for get ingress

```
NAME                         CLASS   HOSTS                      ADDRESS   PORTS   AGE
ingress-anchor-platform      nginx   *                                    80      3h31m
ingress-reference-server     nginx   reference-server.local               80      3h31m
ingress-sep24-reference-ui   nginx   sep24-reference-ui.local             80      3h28m
```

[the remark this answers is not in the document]
philipliu: Ah, I didn't realize they had different `fullName`s. That's fine then.

helm-charts/sep24-reference-ui/templates/configmap.yaml
JiahuiWho: The k8s hostname (only accessible inside the cluster) is not reachable because the ui is a client-side app running in the browser and trying accessing it via localhost:3000 (which is my local machine, not inside the cluster). 

I will go ahead try switch ingress to NodePort

philipliu: Oh right, I think this should be the hostname then.

philipliu (approved):

PR #1629, opened 2025-02-17
[ANCHOR-767] Implement SEP-6 deposit flows for contract accounts

philipliu:
### Description

This implements SEP-6 deposit + deposit exchange flows for contract accounts. The main change is to allow contract accounts as source and destination accounts for SEP-6 deposit requests. The bulk of this PR is adding test code.

### Context

Contract account support

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

SEP-45 does not support memos so we cannot create a new customer every time the test is run. Therefore, we cannot assert on the intermediate PENDING_CUSTOMER_INFO_UPDATE status since a previous run may have already created a customer with enough KYC.


lijamie98 (approved):
LGTM

PR #1633, opened 2025-02-20
[ANCHOR-924] Update sep24 more_info_url config

lijamie98:
### Description

- Update `sep24.more_info_url` config for local helm deployment

### Context

- Support sep24 flow.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/helm-charts/sep-service/values.yaml
+++ b/helm-charts/sep-service/values.yaml
@@ -128,7 +128,7 @@ config:
     interactive_url:
       base_url: http://localhost:3000
     more_info_url:
-      base_url: http://localhost:3000
+      base_url: http://localhost:3000/txn
 
 ingress:
   name: ingress-anchor-platform


philipliu (approved):

helm-charts/sep-service/values.yaml
```diff
@@ -128,7 +128,7 @@ config:
     interactive_url:
       base_url: http://localhost:3000
     more_info_url:
-      base_url: http://localhost:3000
+      base_url: http://localhost:3000/txn
```
philipliu, line 131: Do we want to set this for our `deployment` profile too? I don't remember what the state of the profile was.

lijamie98: This is for local development value. The kube deployment will need to be added separately after @JiahuiWho deploys a stable sep24 flow.

PR #1634, opened 2025-02-21
[PD] Fix ingress host key mismatch between local and kube

JiahuiWho:
### Description

Match the ingress host key between local and kube

### Context

Ingress host is defined with key `ingress.host` in local value file, and this is different from `ingress.rules.host` in Kube, resulting ingress host value might not be properly overwritten

### Testing

- Tested with minikube evrything works locallly.


philipliu (approved):

lijamie98 (approved):
LGTM

PR #1635, opened 2025-02-21
[PD] Fix indention

JiahuiWho:
As title 😭 

Tested with `helm template --debug --dry-run`


philipliu (approved):
It would be great if our CI could catch this. Do you think you can add a step to check the helm charts? I think we can start with running `helm template` to catch the simplest issues.

JiahuiWho:
Track in https://stellarorg.atlassian.net/browse/ANCHOR-972

PR #1647, opened 2025-02-25
[ANCHOR-789] Update SEP-12 to explicitly support contract accounts

philipliu:
### Description

SEP-12 already supports contract accounts since it passes the `account` field from the `WebAuthJwt` directly into the customer callback without checking its type. This PR just adds some unit tests to explicitly test that passing a contract account to all endpoints does not throw exceptions. The SEP-6 end-to-end tests further test that the endpoint works as expected. 

### Context

Contract account support

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
+++ b/core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
@@ -16,11 +16,7 @@ import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.NullSource
 import org.junit.jupiter.params.provider.ValueSource
 import org.skyscreamer.jsonassert.JSONAssert
-import org.stellar.anchor.api.callback.CustomerIntegration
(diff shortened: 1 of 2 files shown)


core/src/test/kotlin/org/stellar/anchor/sep12/Sep12ServiceTest.kt
```diff
@@ -42,6 +38,8 @@ import org.stellar.anchor.util.StringHelper.json
 class Sep12ServiceTest {
   companion object {
     private const val TEST_ACCOUNT = "GBFZNZTFSI6TWLVAID7VOLCIFX2PMUOS2X7U6H4TNK4PAPSHPWMMUIZG"
+    private const val TEST_CONTRACT_ACCOUNT =
+      "CCAASCQKVVBSLREPEUGPOTQZ4BC2NDBY2MW7B2LGIGFUPIY4Z3XUZRVTX"
```
lijamie98, line 42: Do we need to create this in the testnet reset script?

lijamie98 (approved):

PR #1648, opened 2025-02-25
[PD] use call.request.queryParameters

JiahuiWho:
### Description

use `call.request.queryParameters` to properly fetch url query params following the `?`

### Context

When testing with Sep24TestRoute on kubernetes, reference server throws below error
```
16:39:37.446 INFO  - o.s.r.s.Sep24TestRoute - Called /sep24/interactive with parameters Parameters [transaction_id=[1ff8fcca-a5ca-4d2a-b883-723f8abe3b01]]
16:39:37.449 ERROR - o.s.r.ReferenceServer - Unexpected exception
org.stellar.reference.ClientException: Missing token parameter in the request
```
The potential cause could be proxy, encoding,  middlewares or as simple as this PR. 

### Testing

- `./gradlew test`
- Printed out in local that params were correctly parsed.

### DIFF
--- a/kotlin-reference-server/src/main/kotlin/org/stellar/reference/sep24/Sep24TestRoute.kt
+++ b/kotlin-reference-server/src/main/kotlin/org/stellar/reference/sep24/Sep24TestRoute.kt
@@ -24,11 +24,13 @@ fun Route.testSep24(
 ) {
   route("/sep24/interactive") {
     get {
-      log.info { "Called /sep24/interactive with parameters ${call.parameters}" }
+      log.info {
+        "Called /sep24/interactive with parameters ${call.request.queryParameters.entries()}"
+      }
 
       val token =
         JwtDecoder.decode(
-          call.parameters["token"]
+          call.request.queryParameters["token"]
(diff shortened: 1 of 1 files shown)


lijamie98 (approved):

lijamie98:
wow, surprised of this bug and nice catch.

philipliu:
The log makes it seem like `transaction_id` is accessible by `call.parameters`. Any idea why `token` is behaving differently? I didn't have to do this when I setup the UI for the Meridian demo.

JiahuiWho:
This only isn't working with the test route on Kubernetes; it works perfectly fine locally (both for the test route and UI). I haven't tested the full UI flow on Kubernetes, but since it works for you when Meridian, I wouldn't expect any surprises.

Why it's not working - Actually I'm surprised it has been working for so long. There's no documentation indicating that `call.parameters` can be used to fetch query parameters (it does here weirdly); all it mentions is how to fetch route parameters like `/user/{id}`

Lastly I'm not sure this is the ultimate solution for this specific issue, I can only give it a try. But at least this is a solid fix

PR #1661, opened 2025-03-04
[Fix] Update anchor usdc desc in helm chart

JiahuiWho:
### Description

improve the description


philipliu (approved):
Can we also update the org description?

PR #1666, opened 2025-03-06
Remove SEP-24 secret from the reference server

philipliu:
### Description

This combines the secret keys used by the reference server for payments.

### Context

There's no need to have two separate accounts.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


JiahuiWho:
LGTM. Do you need to make another release to sync with the stg and testacnhor?

JiahuiWho (approved):

helm-charts/secret-store/templates/secretstore.yaml
```diff
@@ -25,9 +25,8 @@ spec:
             {
               "POSTGRES_USER": "postgres",
               "POSTGRES_PASSWORD": "123456789",
-              "SEP6_SECRET": "SAJW2O2NH5QMMVWYAN352OEXS2RUY675A2HPK5HEG2FRR2NXPYA4OLYN",
+              "SECRET": "SAJW2O2NH5QMMVWYAN352OEXS2RUY675A2HPK5HEG2FRR2NXPYA4OLYN",
```
lijamie98, line 28: `SECRET` is too general. Can we rename the reference server's `AppConfig::secret` to something more meaningful since it is shared between SEPs. 

ex: `paymentSigningSecret`, etc.

philipliu:
> LGTM. Do you plan to make another release to sync with the stg and testacnhor?

This isn't urgent and can go out in the next planned release.

PR #1679, opened 2025-03-13
Send sentry initialization message in debug level

lijamie98:
### Description

- Send sentry initialization message in debug level. (Log level not available)

### Context

- This is to indicate the sentry is activated instead of waiting for an exception to happen.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/configurator/SentryConfigAdapter.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/configurator/SentryConfigAdapter.java
@@ -1,9 +1,11 @@
 package org.stellar.anchor.platform.configurator;
 
+import static io.sentry.Sentry.captureMessage;
 import static org.stellar.anchor.util.Log.*;
 import static org.stellar.anchor.util.StringHelper.*;
 
 import io.sentry.Sentry;
+import io.sentry.SentryLevel;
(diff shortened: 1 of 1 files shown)


platform/src/main/java/org/stellar/anchor/platform/configurator/SentryConfigAdapter.java
```diff
@@ -30,6 +32,12 @@ void initSentry(ConfigMap config, String dsn, String sentryEnv) {
           options.setTracesSampleRate(1.0);
           options.setEnableUncaughtExceptionHandler(true);
         });
+
+    captureMessage(
```
philipliu, line 36: Does the sentry initialization throw an exception if the dsn or auth tokens are incorrectly set? I'm still unsure if I'm convinced that sending a message is necessary.

philipliu (approved):

platform/src/main/java/org/stellar/anchor/platform/configurator/SentryConfigAdapter.java
lijamie98: unfortunately, no. It only throws exception when posting message or events.

PR #1680, opened 2025-03-13
[ANCHOR-917] Stellar RPC (Soroban RPC) configuration

lijamie98:
### Description

- Stellar RPC (Soroban RPC) configuration

### Context

- Allows AP to communicate with an RPC server.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/config/PropertyAppConfig.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/config/PropertyAppConfig.java
@@ -74,9 +74,7 @@ public class PropertyAppConfig implements AppConfig, Validator {
       }
     }
 
-    if (isEmpty(config.getRpcUrl())) {
-      // TODO: this can only be empty when SEP-45 is disabled
-    } else {
+    if (!isEmpty(config.getRpcUrl())) {
       if (!NetUtil.isUrlValid(config.getRpcUrl())) {
         errors.rejectValue(
             "rpcUrl",
--- a/platform/src/main/java/org/stellar/anchor/platform/config/PropertyAppConfig.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/config/PropertyAppConfig.java
@@ -23,6 +23,9 @@ public class PropertyAppConfig implements AppConfig, Validator {
   @Value("${stellar_network.horizon_url}")
   private String horizonUrl;
 
+  @Value("${stellar_network.rpc_url}")
+  private String rpcUrl;
+
   @Value("${languages}")
   private List<String> languages;
 
@@ -68,6 +71,18 @@ public class PropertyAppConfig implements AppConfig, Validator {
             String.format(
                 "The stellar_network.horizon_url:%s is not in valid format.",
                 config.getHorizonUrl()));
+      }
+    }
+
+    if (isEmpty(config.getRpcUrl())) {
+      // TODO: this can only be empty when SEP-45 is disabled
+    } else {
+      if (!NetUtil.isUrlValid(config.getRpcUrl())) {
+        errors.rejectValue(
+            "rpcUrl",
+            "rpc-url-invalid",
+            String.format(
+                "The stellar_network.rpc_url:%s is not in valid format.", config.getRpcUrl()));
       }
     }
   }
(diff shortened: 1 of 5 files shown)


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,9 +37,19 @@ stellar_network:
   # Use TESTNET or PUBLIC
   network: TESTNET
 
+  # The rpc server endpoint.
+  # If the rpc_url is set, Anchor Platform will use the provided rpc_url to connect to the Stellar network except the
+  # observer server which requires the `horizon_url` to be set until the observer server is updated to use the rpc_url
+  # in the coming P23 release.
```
philipliu, line 43: I think it's enough to say that setting this means that the Anchor Platform will connect to Horizon instead of RPC. I don't think we need to say anything about observer or protocol 23 because when this is released, protocol 23 will be released and the observer would be updated.

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,9 +37,19 @@ stellar_network:
...
+  #
+  # If SEP-45 is enabled, the rpc_url must be set.
```
philipliu, line 45: I mentioned that `rpc_url` needs to be set [here](https://github.com/stellar/anchor-platform/blob/feature/c-accounts/platform/src/main/resources/config/anchor-config-default-values.yaml#L425-L430). Do you think calling it out here is better? Do we want to call it out in both places?

platform/src/main/java/org/stellar/anchor/platform/config/PropertyAppConfig.java
```diff
@@ -70,6 +73,18 @@ void validateConfig(AppConfig config, Errors errors) {
                 config.getHorizonUrl()));
       }
     }
+
+    if (isEmpty(config.getRpcUrl())) {
+      // TODO: this can only be empty when SEP-45 is disabled
```
philipliu, line 78: The validation is done in the SEP-45 config. We should just check that it's not malformed here.

lijamie98: I took the code from the c-accounts branch. I did think it is better to be done in SEP-45. So, I will remove this code.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: removed.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: We should comment it out in sep-45 and so as the validation.

JiahuiWho:
```
if (isEmpty(config.getHorizonUrl())) {
      errors.rejectValue(
          "horizonUrl", "horizon-url-empty", "The stellar_network.horizon_url is not defined.");
    } else { ...}
```
^ We probably want to change the validation here as well

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,9 +37,14 @@ stellar_network:
   # Use TESTNET or PUBLIC
   network: TESTNET
 
+  # The rpc server endpoint.
+  rpc_url: https://soroban-testnet.stellar.org
+
   # The horizon server endpoint.
   horizon_url: https://horizon-testnet.stellar.org
 
+
```
JiahuiWho, line 46: nit: 1 empty line (my ocd :grimacing:)

lijamie98:
> ```
> if (isEmpty(config.getHorizonUrl())) {
>       errors.rejectValue(
>           "horizonUrl", "horizon-url-empty", "The stellar_network.horizon_url is not defined.");
>     } else { ...}
> ```
> 
> ^ Do we need to change the validation here? `stellar_network.horizon_url` will no longer be required once p23 released right?

Before p23 is released, the horizon URL is still needed. We do not want to put anything about p23 at this moment. In fact, I just took out some P23 information based on @philipliu comments.

platform/src/main/resources/config/anchor-config-default-values.yaml
lijamie98: Removed.

philipliu (approved):

PR #1683, opened 2025-03-14
Move secrets into a .env file

philipliu:
### Description

This PR moves all secrets used in tests into environment variables. After this change, developers will need to set up a `.env` to run the Anchor Platform and its tests. The Github Actions workflows have also been modified to pull secrets from Github secrets. Due to a security limitation, workflows will only run successfully for PRs created from branches originating from this repository. In other words, PRs from forks will no longer build. I think this tradeoff is acceptable since there is no real benefit to using forks, and we don't have any external contributors. If this changes, we can reevaluate how to manage secrets.

### Context

We shouldn't be hardcoding secrets in our codebase.

### Testing

- `./gradlew test`

### Documentation

README needs to be updated with instructions for filling out .env

### Known limitations

PRs from forks will no longer build


philipliu:
> Just curious why were SECRET_SEP10_SIGNING_SEED, APP__PAYMENT_SIGNING_SEED, and SECRET__KEY moved along with test secrets? Shouldn't they be kept in config.env with the other production secrets?

We don't want to store the secrets in `config.env` because that file is checked in. The test secrets don't have to be stored along the Anchor Platform secrets but I'm not sure if there is a huge benefit to separating them into individual `.env` files.

lijamie98:
> > Just curious why were SECRET_SEP10_SIGNING_SEED, APP__PAYMENT_SIGNING_SEED, and SECRET__KEY moved along with test secrets? Shouldn't they be kept in config.env with the other production secrets?
> 
> We don't want to store the secrets in `config.env` because that file is checked in. The test secrets don't have to be stored along the Anchor Platform secrets but I'm not sure if there is a huge benefit to separating them into individual `.env` files.

I think this brings up a good point. We used to store secrets in source code, and probably config.env. The good goal about this PR is to make sure no `SECRETS` are visible in the public repo. 

@philipliu Can we document "no secret visible" somewhere in the "How to contribute"?

service-runner/src/main/resources/common/docker-compose.yaml
```diff
@@ -5,6 +5,8 @@ services:
     command: -t
     volumes:
       - ../config:/config
+    env_file:
+      - ../../../../../.env
```
lijamie98, line 9:
```suggestion
      - ../../../../../.env # Relative to docker-compose.yml
```

lijamie98 (approved):
Looks great. A lot of details attended.

lijamie98:
I found `APP_DISTRIBUTION_WALLET` in the variables but not used in the PR. Should we remove?

lijamie98:
Can we also make sure the names are consistent with https://stellarorg.atlassian.net/wiki/spaces/AE/pages/4472766566/Testnet+Reset+Checklist

philipliu:
> I found APP_DISTRIBUTION_WALLET in the settings->variables but not used in the PR. Should we remove?

Yeah, this should be removed.

> Can we also make sure the names are consistent with https://stellarorg.atlassian.net/wiki/spaces/AE/pages/4472766566/Testnet+Reset+Checklist

Yeah, I plan to update the names to use the `TEST_` prefix.

philipliu:
> I found APP_DISTRIBUTION_WALLET in the settings->variables but not used in the PR. Should we remove?

Actually, this is used here https://github.com/stellar/anchor-platform/blob/e6ee729a6bb69608d7274a7680a234e3d54ece99/kotlin-reference-server/src/main/kotlin/org/stellar/reference/event/processor/Sep31EventProcessor.kt#L51

PR #1685, opened 2025-03-18
Fix contract address using a salt on deploy

philipliu:
### Description

This adds the salt parameter to the contract deploy instructions so that the contract addresses stay the same between Testnet resets. The salt used is just the hash hex of "anchor-platform".

### Context

We don't want to update the code when there is a testnet reset.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/soroban/README.md
+++ b/soroban/README.md
@@ -31,15 +31,17 @@ stellar contract deploy \
 --wasm target/wasm32-unknown-unknown/release/account.wasm \
 --source-account ${MY_ACCOUNT} \
 --network testnet \
+--salt 616e63686f722d706c6174666f726d \
 -- \
 --admin ${MY_ACCOUNT}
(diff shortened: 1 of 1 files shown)


lijamie98 (approved):

soroban/README.md
```diff
@@ -31,15 +31,17 @@ stellar contract deploy \
 --wasm target/wasm32-unknown-unknown/release/account.wasm \
 --source-account ${MY_ACCOUNT} \
 --network testnet \
+--salt 616e63686f722d706c6174666f726d \
```
lijamie98, line 34: The salt should be considered as a secret. Can we use one of the variables in Github?

philipliu: No it doesn't need to be a secret, it just needs to be a hash. I just chose it arbitrarily.

PR #1694, opened 2025-03-27
[ANCHOR-1018] Rename fee_details column and bump version

JiahuiWho:
#### What's Changed
Rename db column `fee_details `to `fee_details_list`.
Bump to 3.0.4 for patch release

#### Context
The database stored `feeDetailsList` (a list) under serialized name "fee_details", while sep24 transactionResponse expected a`FeeDetails` (a object) at the same key "fee_details". This caused Gson to fail with `Expected BEGIN_OBJECT but was BEGIN_ARRAY` at [here](https://github.com/stellar/anchor-platform/blob/develop/core/src/main/java/org/stellar/anchor/sep24/Sep24Helper.java#L122). 

#### WHY it didn't work
By default, Gson uses the JSON structure to map fields, not getter/setter methods. Even though `getFeeDetails` been implemented to help construct the object, Gson will still look for a nested `fee_details` object in the JSON.

However in our system the error (expect object but found array) was somehow silent.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -216,7 +216,7 @@ subprojects {
 
 allprojects {
   group = "org.stellar.anchor-sdk"
-  version = "3.0.3"
+  version = "3.0.4"
 
   tasks.jar {
     manifest {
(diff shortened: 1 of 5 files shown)


philipliu (approved):

JiahuiWho:
Confirmed by reporter it works

PR #1706, opened 2025-04-16
Remove on_pull_request_comments

JiahuiWho:
It does not make sense to trigger action on each PR comment, and the action were always skipped
example: https://github.com/stellar/anchor-platform/actions/runs/14498922622


lijamie98 (approved):
LGTM after doc updates

PR #1708, opened 2025-04-17
[ANCHOR-1030] Send THE test payment if not found

lijamie98:
### Description

The integration tests rely on the TESTNET reset script sending the test payment. Instead of relying on the testnet reset script sending the test payment, the integration tests should re-send the payment if not existing. 

### Context

- Improve test stability

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


JiahuiWho:
Changes looks good to me. Just to clarify my understanding — Together with [Remove send test payment script](https://github.com/stellar/useful-scripts/pull/11), the motivation behind was that StellarRPC doesn't index payments, so we shouldn't rely on querying them after submission. But in this update, although the transaction submission was moved into the integration test, it looks like we're still using Horizon to fetch and verify the payment.

Am I understanding that right? If so, shouldn’t we also update the verification part to rely on StellarRPC instead of Horizon?

lijamie98:
> Changes looks good to me. Just to clarify my understanding — Together with [Remove send test payment script](https://github.com/stellar/useful-scripts/pull/11), the motivation behind was that StellarRPC doesn't index payments, so we shouldn't rely on querying them after submission. But in this update, although the transaction submission was moved into the integration test, it looks like we're still using Horizon to fetch and verify the payment.
> 
> Am I understanding that right? If so, shouldn’t we also update the verification part to rely on StellarRPC instead of Horizon?

Yes, your understanding is correct. This PR is targeting the `develop` branch. The StellarRPC verification will be implemented in the `feature/stellar-rpc` branch.

JiahuiWho (approved):
Feel free to merge once extended tests is fixed

PR #1722, opened 2025-05-01
[ANCHOR-977] SorobanPaymentObserver with mock event stream 

lijamie98:
### Description

- Create `SorobanPaymentObserver` to observe payment transactions with Stellar RPC API.
- Use `getTransactions` endpoint to mock `getEvents`
- The `getTransactions` endpoint will be refactored to `getEvents` after P23 is released.
- Add `soroban-rpc-default` to replace Horizon with RPC.
- Cursor serialization is not implemented until `getEvents` implementation.
- `SorobanPaymentObserver` unit tests will be added after `getEvents` implementation.

### Context

- Stellar RPC integration.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
```diff
@@ -113,15 +116,24 @@ public static ParsedTransaction parseTransaction(TransactionEnvelope txnEnv, Str
                 txnEnv.getV1().getTx().getSourceAccount().getEd25519().getUint256());
         memo = txnEnv.getV1().getTx().getMemo();
         break;
+      case ENVELOPE_TYPE_TX_FEE_BUMP:
```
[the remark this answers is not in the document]
philipliu: We need to process this because the majority of payments from smart wallets will be wrapped in them.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
```diff
@@ -226,4 +240,74 @@ public static LedgerOperation toLedgerOperation(OperationResponse op) {
...
+  }
+
+  /**
+   * Converting from Soroban transaction to LedgerTransaction.
+   *
+   * @param txn
+   * @return
+   * @throws LedgerException
+   */
+  public static LedgerTransaction fromSorobanTransaction(GetTransactionsResponse.Transaction txn)
```
philipliu, line 286: What's the difference between this method and the one above?

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
```diff
@@ -226,4 +240,74 @@ public static LedgerOperation toLedgerOperation(OperationResponse op) {
...
+        .createdAt(Instant.ofEpochSecond(txnResponse.getCreatedAt()))
+        .operations(operations)
+        .build();
+  }
+
+  /**
+   * Converting from Soroban transaction to LedgerTransaction.
+   *
+   * @param txn
+   * @return
```
philipliu, line 283: nit: Javadoc is missing

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/StellarObserverTests.kt
```diff
@@ -30,37 +31,40 @@ class StellarObserverTests : AbstractIntegrationTests(TestConfig()) {
...
+    assertEquals(2, checks.size)
+    assertNotNull(checks["config"])
 
-    val stellarPaymentObserverCheck = checks["stellar_payment_observer"] as Map<*, *>
-    Assertions.assertEquals(2, stellarPaymentObserverCheck.size)
-    Assertions.assertTrue(
-      (stellarPaymentObserverCheck["status"] as String) in setOf("GREEN", "YELLOW")
-    )
+    if (isNotEmpty(this.config.env["stellar_network.rpc_url"])) {
+      val stellarPaymentObserverCheck = checks["soroban_payment_observer"] as Map<*, *>
```
philipliu, line 51: nit: Can we call it `rpc_payment_observer`?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
```diff
@@ -0,0 +1,387 @@
...
+  final List<PaymentListener> paymentListeners;
+  final StellarPaymentStreamerCursorStore paymentStreamerCursorStore;
+  final PaymentObservingAccountsManager paymentObservingAccountsManager;
+
+  final ExponentialBackoffTimer publishingBackoffTimer;
+  final ExponentialBackoffTimer streamBackoffTimer;
+  final ExponentialBackoffTimer databaseBackoffTimer = new ExponentialBackoffTimer(1, 20);
+
+  int silenceTimeoutCount = 0;
+  HorizonPaymentObserver.ObserverStatus status = RUNNING;
```
philipliu, line 55: Should `ObserverStatus` be defined outside of `HorizonPaymentObserver`?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
```diff
@@ -0,0 +1,387 @@
...
+  void checkStatus() {
+    switch (status) {
+      case NEEDS_SHUTDOWN:
+        infoF("shut down the observer");
+        shutdown();
+        break;
+      case STREAM_ERROR:
+        // We got stream connection error. We will use the backoff timer to reconnect.
+        // If the backoff timer reaches max, we will shut down the observer
+        if (streamBackoffTimer.isTimerMaxed()) {
```
philipliu, line 196: RPC's event endpoint does not provide an SSEStream, so this error would not be relevant to the RPC observer.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
```diff
@@ -0,0 +1,387 @@
...
+    @SerializedName("thread_shutdown")
+    boolean threadShutdown;
+
+    @SerializedName("thread_terminated")
+    boolean threadTerminated;
+
+    boolean stopped;
+
+    @SerializedName("last_event_id")
+    String lastEventId;
```
philipliu, line 334: Where is this set?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
```diff
@@ -0,0 +1,317 @@
...
+    String latestCursor = fetchStreamingCursor();
+    infoF("SSEStream cursor={}", latestCursor);
+
+    PaymentsRequestBuilder paymentsRequest =
+        horizon
+            .getServer()
+            .payments()
+            .cursor(latestCursor)
+            .order(RequestBuilder.Order.ASC)
+            .limit(MAX_RESULTS);
```
philipliu, line 89: Previously, we set the request to include the transactions, but now we make a call to `getTransaction`. Is there a way to avoid this secondary request?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
```diff
@@ -0,0 +1,317 @@
...
+      return null;
+    }
+    String token = pageOpResponse.getRecords().get(0).getPagingToken();
+    infoF("The latest cursor fetched from Stellar network is: {}", token);
+    return token;
+  }
+
+  void processOperation(OperationResponse operationResponse) {
+    try {
+      PaymentTransferEvent transferEvent = toPaymentTransferEvent(operationResponse);
```
philipliu, line 193: Why do we need to represent the operation in terms of an event? I thought we got rid of `ObservedPayment` since it was an unnecessary abstraction.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
```diff
@@ -0,0 +1,317 @@
...
+      // only fetch transaction if the operation is a payment operation
+      return null;
+    }
+
+    LedgerTransaction txn = horizon.getTransaction(operation.getTransactionHash());
+    if (txn == null) {
+      debugF("Transaction not found: {}", operation.getTransactionHash());
+      return null;
+    }
+    if (operation instanceof PaymentOperationResponse paymentOp) {
```
philipliu, line 232: These two branches have the same logic.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
```diff
@@ -0,0 +1,317 @@
...
+  PaymentTransferEvent toPaymentTransferEvent(OperationResponse operation) throws LedgerException {
+    if (!(operation instanceof PaymentOperationResponse
+        || operation instanceof PathPaymentBaseOperationResponse)) {
+      // only fetch transaction if the operation is a payment operation
+      return null;
+    }
+
+    LedgerTransaction txn = horizon.getTransaction(operation.getTransactionHash());
+    if (txn == null) {
+      debugF("Transaction not found: {}", operation.getTransactionHash());
```
philipliu, line 229: This seems like an error?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
```diff
@@ -0,0 +1,184 @@
...
+
+  void shutdownMockStream() {
+    task.cancel(true);
+  }
+
+  private void fetchEvent() {
+    String cursor = fetchCursor();
+    GetTransactionsRequest request;
+    if (cursor == null) {
+      request = GetTransactionsRequest.builder().startLedger(getLatestLedger()).build();
```
philipliu, line 97: Shouldn't we be fetching the next ledger, otherwise we may skip payments?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
```diff
@@ -0,0 +1,184 @@
...
+    } else {
+      request =
+          GetTransactionsRequest.builder()
```
philipliu, line 100: Is there a reason we are not calling the `getEvents` endpoint?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
```diff
@@ -0,0 +1,184 @@
...
+      }
+      saveCursor(response.getCursor());
+    }
+  }
+
+  private void processOperation(LedgerTransaction ledgerTxn, LedgerTransaction.LedgerOperation op)
+      throws IOException, AnchorException {
+    PaymentTransferEvent event =
+        switch (op.getType()) {
+          case PAYMENT -> {
```
philipliu, line 139: Why do we differentiate payment vs path payments? The transfer events look the same.

lijamie98: The class for `PAYMENT `and `PATH_PAYMENT_STRICT_SEND, PATH_PAYMENT_STRICT_RECEIVE ` are different. Although the code look the same, I would not merge them.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
lijamie98: The `getEvents` does not have classic asset transfer events. This will be available after P23.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
lijamie98: We should be fetching from the last saved cursor which will be implemented[ here](https://github.com/stellar/anchor-platform/blob/wip/anchor-977-observer-rpc-4/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java).

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: If this happens, it will be a horizon bug for which, I don't want to throw error here instead a debug message is logged.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: They look the same logic but they are different because the OperationResponse classes are different. I tried merging but did not find obviously better solutions.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: The event is to generalize `OperationResponse` and `GetTransactionResponse`. After P23, it is used to generalize Soroban `Transaction`. The event is designed to be very close to CAP-67 events.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: Previously, we are streaming operations with transactions. ([here](https://github.com/stellar/anchor-platform/blob/develop/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarPaymentObserver.java#L143))

I changed to avoid streaming transactions but only fetch them when necessary.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
lijamie98: https://github.com/stellar/anchor-platform/blob/wip/anchor-977-observer-rpc-4/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java#L298

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
lijamie98: The RPC observer cursor/timeout/error handling may be refactored after P23. 

After switching to `getEvents`, I will have better visibility.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
lijamie98: Good catch. The `ObserverStatus` is defined `AbstractPaymentObserver` although `HorizonPaymentObserver.ObserverStatsus` is also valid. 

Removed `Horizon`.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/StellarObserverTests.kt
lijamie98: `rpc` is quite ambiguous in the AP name space. We have the Platform `RPC` API. 

I have renamed to `stellar_rpc_payment_observer`.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: added.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: The previous method is only used for `getTransactions` endpoint. After switching to `getEvents`, it can be removed.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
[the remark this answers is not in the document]
lijamie98: Will handle this in a separate ticket: https://stellarorg.atlassian.net/browse/ANCHOR-1041

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
philipliu: I guess what I'm asking is if the type of transfer matters from the observer's perspective. If not, I think we should just not differentiate them.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
philipliu: Is this field only relevant to SSEStream?

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: ok with me

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
lijamie98: I am not sure until `getEvetns` is used.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: No we cannot. One transaction may have multiple events.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
lijamie98: Classic transfers are not emitting events in `getEvents`. It won't work with classic assets today.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
lijamie98: It matters because the classes are different.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: This is good catch of a bug. Fix: https://github.com/stellar/anchor-platform/pull/1722/files#diff-12f04e318f10ff475cd812290222623b3083afd23b16543a4706307b6035a92dR227

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/HorizonPaymentObserver.java
lijamie98: When we switch to `getEvents`, I would keep the information for the `event`. I also don't think we should separate the logic for Horizon and StellarRPC. I use the event generalize the processing logic.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SorobanPaymentObserver.java
philipliu: That's fine because the SAC transfer events look like the classic events. I don't think we need to wait until P23 to implement the `getEvents` observer.

lijamie98: The scope of this PR is not for SAC transfers. The scope is to mock the events (PaymentTransferEvent) to make sure CAP-67 is not missing anything for AP. 

This PR makes sure SEP-6, 24, 31 work with the `PaymentTransferEvent` which mimics the Stellar RPC events. The `getEvents` observer is defined in a separate ticket. https://stellarorg.atlassian.net/browse/ANCHOR-981

JiahuiWho (approved):

PR #1728, opened 2025-05-09
[ANCHOR-1049] Fix NPE when the `max_amount` is not set in assets.yaml

lijamie98:
### Description

If the following attributes are not set in the `assets.yaml`, NPE is thrown:

- `sep6.deposit.max_amount`
- `sep6.withdraw.max_amount`
- `sep24.deposit.max_amount`
- `sep24.withdraw.max_amount`
- `sep31.receive.max_amount`

### Context

- Bug fix

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/resources/config/anchor-asset-default-values.yaml
+++ b/core/src/main/resources/config/anchor-asset-default-values.yaml
@@ -32,7 +32,7 @@ items:
         # The minimum amount that the user can deposit.
         min_amount: 0
         # The maximum amount that the user can deposit.
-        max_amount:
+        max_amount: 999999999
         # The list of the methods that the user can use to deposit the asset.
         # Example:
         #   - SEPA
@@ -45,7 +45,7 @@ items:
         # The minimum amount that the user can withdraw.
         min_amount: 0
         # The maximum amount that the user can withdraw.
-        max_amount: 1000
+        max_amount: 999999999
         # The list of the methods that the anchor supports for the withdrawal of the asset.
         # Example:
         #   - bank_account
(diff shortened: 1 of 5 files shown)


JiahuiWho:
Would you mind sharing more insights on where is this NPE being thrown from? If `max_amount` needs to be set, does it mean `min_amount` should be set as well?

lijamie98:
> Would you mind sharing more insights on where is this NPE being thrown from? If `max_amount` needs to be set, does it mean `min_amount` should be set as well?

The origin of NPE is https://github.com/stellar/anchor-platform/blob/develop/core/src/main/java/org/stellar/anchor/util/AssetValidator.java#L172

The `min_amount` was set in `anchor-asset-default-values.yaml`. However, the `max_amount` was not set.

We have two options:
- Set `max_amount` to `Long.MAX_VALUE` in the yaml. This is a little confusing.
- Set `max_amount` to `Long.MAX_VALUE` in the class.

This PR implements the second option.

core/src/main/resources/config/anchor-asset-default-values.yaml
```diff
@@ -45,7 +45,7 @@ items:
         # The minimum amount that the user can withdraw.
         min_amount: 0
```
philipliu, line 46: Can we also set the min amount in code for consistency, and document the default values in this file?

lijamie98: After discussion with Jiahui, we decided to set the values at the `anchor-assets-default-values.yaml` for consistency.

JiahuiWho (approved):
LGTM

philipliu (approved):

PR #1734, opened 2025-05-12
[ANCHOR-1044] The lookupAndUpdate function should handle the mux account.

lijamie98:
### Description

- The lookupAndUpdate function should handle the mux account.
- When a mux account is seen from the observer, we will use the G-account of the M-account for account lookup.

### Context

- CAP-67 memo support

### Testing

- `./gradlew test`


platform/src/main/java/org/stellar/anchor/platform/observer/stellar/PaymentObservingAccountsManager.java
```diff
@@ -110,10 +111,21 @@ public List<ObservingAccount> getAccounts() {
...
+   * @param account The account to be checked. The account can be a muxed account or a G-account.
    * @return true if the account is being observed. false, otherwise.
    */
   public boolean lookupAndUpdate(String account) {
+    if (account == null) {
+      return false;
+    }
+
+    // MuxedAccount handles both muxed and non-muxed accounts
+    if (account.startsWith("M")) {
```
philipliu, line 123: Is this change all that is required to handle transfers from muxed accounts in the SEPs?

philipliu: I don't think we need to wait until Protocol 23 for this, right? We should be able to submit transactions from muxed accounts today.

philipliu (approved):

PR #1735, opened 2025-05-15
[ANCHOR-1054] Handle contract payment events in the StellarRpcPaymentObserver and tests.

lijamie98:
### Description

- Handle contract payment events in the StellarRpcPaymentObserver and tests.
- Minor refactoring of LedgerClientHelp to reduce the size of the class and move them to calling classes if not commonly used.

### Context

- Add c-account support in the observer

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
+++ b/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
@@ -6,6 +6,7 @@ import static org.stellar.anchor.util.Log.*;
 import static org.stellar.sdk.xdr.HostFunctionType.HOST_FUNCTION_TYPE_INVOKE_CONTRACT;
 import static org.stellar.sdk.xdr.OperationType.*;
 
+import java.math.BigInteger;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
@@ -52,7 +53,7 @@ public class LedgerClientHelper {
                 LedgerPaymentOperation.builder()
                     .id(operationId)
                     .asset(payment.getAsset())
-                    .amount(payment.getAmount().getInt64())
+                    .amount(BigInteger.valueOf(payment.getAmount().getInt64()))
                     .from(sourceAccount)
                     .sourceAccount(sourceAccount)
                     .to(
@@ -98,7 +99,7 @@ public class LedgerClientHelper {
                 LedgerPathPaymentOperation.builder()
                     .id(operationId)
                     .asset(asset)
-                    .amount(amount)
+                    .amount(BigInteger.valueOf(amount))
                     .from(sourceAccount)
                     .to(toAddress)
                     .sourceAccount(sourceAccount)
@@ -126,7 +127,7 @@ public class LedgerClientHelper {
                     .contractId(contractId)
                     .hostFunction("transfer")
                     .id(operationId)
-                    .amount(amount.getI128().getLo().getUint64().getNumber().longValue())
+                    .amount(amount.getI128().getLo().getUint64().getNumber())
                     .from(getAddressOrContractId(from.getAddress()))
                     .to(getAddressOrContractId(to.getAddress()))
                     .sourceAccount(sourceAccount)
--- a/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
+++ b/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
@@ -5,8 +5,8 @@ import static org.stellar.anchor.ledger.LedgerTransaction.*;
 import static org.stellar.anchor.util.AssetHelper.toXdrAmount;
 import static org.stellar.anchor.util.Log.*;
 import static org.stellar.sdk.responses.sorobanrpc.SendTransactionResponse.*;
-import static org.stellar.sdk.xdr.OperationType.PATH_PAYMENT_STRICT_RECEIVE;
-import static org.stellar.sdk.xdr.OperationType.PAYMENT;
+import static org.stellar.sdk.xdr.HostFunctionType.HOST_FUNCTION_TYPE_INVOKE_CONTRACT;
+import static org.stellar.sdk.xdr.OperationType.*;
 import static org.stellar.sdk.xdr.SignerKeyType.*;
 import static org.stellar.sdk.xdr.SignerKeyType.SIGNER_KEY_TYPE_ED25519_SIGNED_PAYLOAD;
 
@@ -115,6 +115,31 @@ public class LedgerClientHelper {
                     .sourceAccount(sourceAccount)
                     .build())
             .build();
+      }
+      case INVOKE_HOST_FUNCTION -> {
+        HostFunction hostFunction = op.getBody().getInvokeHostFunctionOp().getHostFunction();
+        if (hostFunction.getDiscriminant() != HOST_FUNCTION_TYPE_INVOKE_CONTRACT) yield null;
+        if (!hostFunction
+            .getInvokeContract()
+            .getFunctionName()
+            .getSCSymbol()
+            .toString()
+            .equals("transfer")) yield null;
+        SCAddress contractAddress = hostFunction.getInvokeContract().getContractAddress();
+        String contractId = StrKey.encodeContract(contractAddress.getContractId().getHash());
+        SCVal from = hostFunction.getInvokeContract().getArgs()[0];
+        SCVal to = hostFunction.getInvokeContract().getArgs()[1];
+        SCVal amount = hostFunction.getInvokeContract().getArgs()[2];
+        LedgerInvokeHostFunctionOperation.builder()
+            .stellarAssetContractId(contractId)
+            .hostFunction("transfer")
+            .id(operationId)
+            .amount(amount.getI64().getInt64())
+            .from(StrKey.encodeEd25519PublicKey(from.getAddress().getAccountId()))
+            .to(StrKey.encodeEd25519PublicKey(to.getAddress().getAccountId()))
+            .sourceAccount(sourceAccount)
+            .build();
+        yield LedgerOperation.builder().build();
       }
       default -> null;
     };
--- a/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
+++ b/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
@@ -3,10 +3,8 @@ package org.stellar.anchor.ledger;
 import static java.lang.Thread.sleep;
 import static org.stellar.anchor.ledger.LedgerTransaction.*;
 import static org.stellar.anchor.util.Log.*;
-import static org.stellar.sdk.responses.sorobanrpc.SendTransactionResponse.*;
 import static org.stellar.sdk.xdr.HostFunctionType.HOST_FUNCTION_TYPE_INVOKE_CONTRACT;
 import static org.stellar.sdk.xdr.OperationType.*;
-import static org.stellar.sdk.xdr.SignerKeyType.*;
 
 import java.time.Instant;
 import java.util.ArrayList;
@@ -125,7 +123,7 @@ public class LedgerClientHelper {
             .type(INVOKE_HOST_FUNCTION)
             .invokeHostFunctionOperation(
                 LedgerInvokeHostFunctionOperation.builder()
-                    .stellarAssetContractId(contractId)
+                    .contractId(contractId)
                     .hostFunction("transfer")
                     .id(operationId)
                     .amount(amount.getI128().getLo().getUint64().getNumber().longValue())
--- a/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
+++ b/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
@@ -121,21 +121,38 @@ public class LedgerClientHelper {
         SCVal from = hostFunction.getInvokeContract().getArgs()[0];
         SCVal to = hostFunction.getInvokeContract().getArgs()[1];
         SCVal amount = hostFunction.getInvokeContract().getArgs()[2];
-        LedgerInvokeHostFunctionOperation.builder()
-            .stellarAssetContractId(contractId)
-            .hostFunction("transfer")
-            .id(operationId)
-            .amount(amount.getI64().getInt64())
-            .from(StrKey.encodeEd25519PublicKey(from.getAddress().getAccountId()))
-            .to(StrKey.encodeEd25519PublicKey(to.getAddress().getAccountId()))
-            .sourceAccount(sourceAccount)
+        yield LedgerOperation.builder()
+            .type(INVOKE_HOST_FUNCTION)
+            .invokeHostFunctionOperation(
+                LedgerInvokeHostFunctionOperation.builder()
+                    .stellarAssetContractId(contractId)
+                    .hostFunction("transfer")
+                    .id(operationId)
+                    .amount(amount.getI128().getLo().getUint64().getNumber().longValue())
+                    .from(getAddressOrContractId(from.getAddress()))
+                    .to(getAddressOrContractId(to.getAddress()))
+                    .sourceAccount(sourceAccount)
+                    .build())
             .build();
-        yield LedgerOperation.builder().build();
       }
       default -> null;
     };
   }
 
+  static String getAddressOrContractId(SCAddress address) {
+    return switch (address.getDiscriminant()) {
+      case SC_ADDRESS_TYPE_ACCOUNT -> StrKey.encodeEd25519PublicKey(address.getAccountId());
+      case SC_ADDRESS_TYPE_CONTRACT -> StrKey.encodeContract(address.getContractId().getHash());
+    };
+  }
+
+  /**
+   * Parse the transaction envelope and extract the operations, source account, and memo.
+   *
+   * @param txnEnv the transaction envelope
+   * @param txnHash the transaction hash
+   * @return a ParseResult containing the operations, source account, and memo
+   */
   public static ParseResult parseOperationAndSourceAccountAndMemo(
       TransactionEnvelope txnEnv, String txnHash) {
     Operation[] operations;
--- a/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
+++ b/core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
@@ -2,27 +2,18 @@ package org.stellar.anchor.ledger;
 
 import static java.lang.Thread.sleep;
 import static org.stellar.anchor.ledger.LedgerTransaction.*;
-import static org.stellar.anchor.util.AssetHelper.toXdrAmount;
 import static org.stellar.anchor.util.Log.*;
 import static org.stellar.sdk.responses.sorobanrpc.SendTransactionResponse.*;
 import static org.stellar.sdk.xdr.HostFunctionType.HOST_FUNCTION_TYPE_INVOKE_CONTRACT;
 import static org.stellar.sdk.xdr.OperationType.*;
 import static org.stellar.sdk.xdr.SignerKeyType.*;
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
```diff
@@ -116,10 +107,52 @@ static LedgerOperation convert(
...
+        SCVal to = hostFunction.getInvokeContract().getArgs()[1];
+        SCVal amount = hostFunction.getInvokeContract().getArgs()[2];
+        yield LedgerOperation.builder()
+            .type(INVOKE_HOST_FUNCTION)
+            .invokeHostFunctionOperation(
+                LedgerInvokeHostFunctionOperation.builder()
+                    .stellarAssetContractId(contractId)
+                    .hostFunction("transfer")
+                    .id(operationId)
+                    .amount(amount.getI128().getLo().getUint64().getNumber().longValue())
```
philipliu, line 131: The unit is in stroops, which I think is `10^-7`, so it needs to be multiplied to match the Horizon amount type. Also, can you check the units of the other operations to make sure they're correct, as well?

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
```diff
@@ -116,10 +107,52 @@ static LedgerOperation convert(
...
+        SCAddress contractAddress = hostFunction.getInvokeContract().getContractAddress();
+        String contractId = StrKey.encodeContract(contractAddress.getContractId().getHash());
+        SCVal from = hostFunction.getInvokeContract().getArgs()[0];
+        SCVal to = hostFunction.getInvokeContract().getArgs()[1];
+        SCVal amount = hostFunction.getInvokeContract().getArgs()[2];
+        yield LedgerOperation.builder()
+            .type(INVOKE_HOST_FUNCTION)
+            .invokeHostFunctionOperation(
+                LedgerInvokeHostFunctionOperation.builder()
+                    .stellarAssetContractId(contractId)
```
philipliu, line 128: nit: We should rename this to just `contractId` since we will eventually want to support all SEP-41 tokens.

core/src/main/java/org/stellar/anchor/ledger/LedgerTransaction.java
```diff
@@ -70,6 +71,32 @@ public static class LedgerPathPaymentOperation implements LedgerPayment {
     String sourceAccount;
   }
 
+  @Builder
+  @Data
+  public static class LedgerInvokeHostFunctionOperation implements LedgerPayment {
```
philipliu, line 76: nit: There are contract invocations besides transfers, but the fields of this class are only relevant to transfers. Can we rename this to something like `LedgerInvokeHostFunctionTransferOperation`?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/AbstractIntegrationTests.kt
```diff
@@ -123,6 +127,43 @@ abstract class AbstractIntegrationTests(val config: TestConfig) {
       .build()
   }
 
+  private fun toLedgerOperation(op: OperationResponse): LedgerOperation? {
```
philipliu, line 130: Do we need to handle `InvokeHostFunctionOperation`s?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -43,10 +52,18 @@ class PaymentObserverTests {
     @JvmStatic
     @BeforeAll
     fun start() {
-      fromKeyPair = createAndFundAccount()
-      toKeyPair = createAndFundAccount()
-      fromKeyPair2 = createAndFundAccount()
-      toKeyPair2 = createAndFundAccount()
+      fromKeyPair =
```
philipliu, line 55: Can we remove the secrets out of the code?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -43,10 +52,18 @@ class PaymentObserverTests {
...
+        KeyPair.fromSecretSeed("SD4XJTFKVTCUC7XOHJLUC7YZHOCZRKOTQ3OAR7VAUKOUCKPGPNC25ZXB")
+      toKeyPair = KeyPair.fromSecretSeed("SAB7GZ6KN2YLJ2NM52JYM4OWXY55AVG54LPYUEVUZGIPUY2PUVFN6OJB")
+      fromKeyPair2 =
+        KeyPair.fromSecretSeed("SAJPASKHPZAMLY27IQBTBYB45G7D62QNXUV3WE5NP42YCVHSC7EMWGDQ")
+      toKeyPair2 =
+        KeyPair.fromSecretSeed("SBRU7TKRPRM2EQC3KLAWNLNKIJ6HTEQVIJ4AQQJCUQJX725NN3M6BWKF")
+
+      //      fromKeyPair = createAndFundAccount()
```
philipliu, line 63: We can remove these comments.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -138,6 +181,101 @@ class PaymentObserverTests {
...
+        .build()
+
+    // sign and send the transaction
+    val preparedTransaction = rpc.prepareTransaction(authorizedTransaction)
+    preparedTransaction.sign(signer)
+    return rpc.sendTransaction(preparedTransaction)
+  }
+
+  //
+  //  private fun assertEventsSacPayment(
```
philipliu, line 261: Can we remove this?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -91,6 +108,32 @@ class PaymentObserverTests {
     eventCaptureListenerStellarRpc.reset()
   }
 
+  @Test
+  fun `submit a payment to a contract and assert events are received as expected`(): Unit =
```
philipliu, line 112: I don't think we need to monitor payments to a contract, but we need to monitor payments from a contract. We can keep this test but we need to add one for the reverse.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
```diff
@@ -0,0 +1,81 @@
...
+      return null;
+    }
+
+    String assetName = scAssetName.getStr().getSCString().toString();
+    Asset asset = org.stellar.sdk.Asset.create(assetName).toXdr();
+    sacToAssetMap.put(sac, asset);
+    return asset;
+  }
+
+  private SCVal fetchSacMetadata(String sac) throws IOException {
```
philipliu, line 50: Do we need to fetch this from RPC?

Another approach would be to calculate the contract IDs of the assets we care about on start-up and store them as part of the `AssetConfig` class so we don't have to perform a lookup on every transfer event we observe.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```diff
@@ -30,10 +32,12 @@
 import org.stellar.sdk.responses.sorobanrpc.GetLatestLedgerResponse;
 import org.stellar.sdk.responses.sorobanrpc.GetTransactionsResponse;
 import org.stellar.sdk.xdr.OperationType;
+import org.stellar.sdk.xdr.TransactionEnvelope;
 
 public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
-  final StellarRpc stellarRpc;
-  final SorobanServer sorobanServer;
+  @Getter final StellarRpc stellarRpc;
+  @Getter final SorobanServer sorobanServer;
```
philipliu, line 39: Isn't `StellarRpc` a wrapper around `SorobanServer`? Why do we need both classes?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```diff
@@ -189,13 +204,28 @@ private void processOperation(LedgerTransaction ledgerTxn, LedgerOperation op)
...
                 .amount(pathPaymentOp.getAmount())
                 .txHash(ledgerTxn.getHash())
                 .operationId(pathPaymentOp.getId())
                 .ledgerTransaction(ledgerTxn)
                 .build();
           }
+          case INVOKE_HOST_FUNCTION -> {
+            LedgerTransaction.LedgerInvokeHostFunctionOperation invokeOp =
+                op.getInvokeHostFunctionOperation();
+            invokeOp.setAsset(
```
philipliu, line 217: `getStellarAssetContractId` can return null if the contract is not a SAC. Is that ok?

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: The `amount` in LedgerOperatio` is in stroops which is `10^-7`.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: renamed.

core/src/main/java/org/stellar/anchor/ledger/LedgerTransaction.java
lijamie98: That's actually considered during development. In the design of these common classes, we only implement subsets of the operations. Renaming `LedgerInvokeHostFunctionTransferOperation` causes certain inconsistency with  the `OperationType` of the `LedgerOperation` class. IMO, it will be more confusing. I would keep `LedgerInvokeHostFunctionOperation` unless we have a use case that causes issues.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/AbstractIntegrationTests.kt
lijamie98: The function is used by Horizon tests only. 

Can we use Horizon to call SAC to send assets?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
lijamie98: Removed.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
lijamie98: These are the code that should replace the test accounts. Uncommented.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
lijamie98: removed.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
lijamie98: The mapper does not fetch from RPC for every transfer event that we observe. It only fetch the first time it observes.

Fetch them may move the fetch earlier to the initialization but it does not reduce the number of calls. We may further optimize if we observe the laze fetch an issue.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: Yes it is.

In line 50: 
```
this.sorobanServer = stellarRpc.getSorobanServer();
```

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: `LedgerInvokeHostFunctionOperation` always has a `contractId` because we only convert convert when the type is `HOST_FUNCTION_TYPE_INVOKE_CONTRACT`.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
lijamie98: removed

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/AbstractIntegrationTests.kt
philipliu: Yeah, we can, but it isn't relevant here.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
philipliu: The test is still here, and that's fine. I think we need a test that submits a payment _from_ a contract and asserts that the observer observes it.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
philipliu: With this approach, there will be an entry for every asset we observe at least once. But we only care about a handful of them, so why not calculate the contract IDs of those assets and check against that?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
philipliu: Right, I'm wondering why not just remove the dependency on `SorobanServer` when we can just implement the functionality into `StellarRpc`?

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: The Stellar RPC defines the amount in stroops and the `LedgerTransaction` class follows that and the type is `Long`. It is converted to Horizon's `amount` in `BigDecimal` when necessary.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
lijamie98: Oops, my comments were entered incorrectly. It was meant for other comments. 

The ticket is created for the test https://stellarorg.atlassian.net/browse/ANCHOR-1056

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
lijamie98: This approach doe s not keep an entry of every asset that we observe but only the assets of the accounts that are in the `PaymentObservingAccountsManager`.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: Calling `getTransaction` is temporary and will be replaced by `getEvents` which should not be part of the `LedgerClient` interface.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: The mapper checks of the contract is `CONTRACT_EXECUTABLE_STELLAR_ASSET`. This is always an SAC though.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
philipliu: Okay, I see that `submit a payment to a contract and assert events are received as expected` correctly asserts the amount. There is a bug here, though. We need to consider the `Hi` part of `I128`, not just the `Lo`.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
philipliu: The amount of assets we store in the map is going to be small because only a handful of assets will be transferred to the Anchor's distribution account, but there is a way to avoid the lookup altogether and not have to deal with SEP-41 compliant tokens that are not SAC.

I think it's fine to implement it like this, but just calling out that there are more edge cases we need to handle by implementing it this way.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```diff
@@ -189,13 +204,27 @@ private void processOperation(LedgerTransaction ledgerTxn, LedgerOperation op)
...
                 .amount(pathPaymentOp.getAmount())
                 .txHash(ledgerTxn.getHash())
                 .operationId(pathPaymentOp.getId())
                 .ledgerTransaction(ledgerTxn)
                 .build();
           }
+          case INVOKE_HOST_FUNCTION -> {
+            LedgerTransaction.LedgerInvokeHostFunctionOperation invokeOp =
+                op.getInvokeHostFunctionOperation();
+            invokeOp.setAsset(sacToAssetMapper.getAssetFromSac(invokeOp.getContractId()));
```
philipliu, line 217: I think we need to check `sacToAssetMapper.getAssetFromSac(invokeOp.getContractId())` is not null, right?

philipliu (approved):
Suggested some potential improvements and a couple of minor fixes, but LGTM at a high level.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/SacToAssetMapper.java
lijamie98: Yes, I also consider SEP-41 scenarios. But I decided to approach that problem when we support Custom Tokens in Q3 so I may reduce the scope of Stellar RPC.

core/src/main/java/org/stellar/anchor/ledger/LedgerClientHelper.java
lijamie98: Hmm... we may need to do some refactor to change amount to `BigInteger`.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: If `getAssetFromSac` returns `null`, this means the `SAC` does not exist.  In that case, the `invokeOp.asset` will be `null` that causes an `IllegalStateException` thrown as expected.

philipliu (approved):
I think the changes look good at a high level, but we should fix https://github.com/stellar/anchor-platform/pull/1735/files#r2096633863 in this PR.

PR #1742, opened 2025-05-21
[ANCHOR-1029] Support SRT in anchor platform

JiahuiWho:
### What's changed

1. Supports SRT usage in demo wallet SEP-6 by adding the `rate`
2. Improving error msg.  `....resultCodes` includes sth more meaningful like `op_src_no_trust `, whereas the `....resultCodes?.transactionResultCode` just always return `tx_failed`

### Test 
Tested with demo wallet with local setup, also tried SEP-6 and SEP-24 e2e tests with hardcoded SRT asset.
Once this PR being checked in, we can test against ap-dev instance


philipliu (approved):

lijamie98 (approved):
LGTM

PR #1750, opened 2025-05-30
[CHORE] In build.gradle.kts, add doLast when necessary. 

lijamie98:
### Description

- In build.gradle.kts, add doLast when necessary. 
- Add eclipse plugin for vscode

### Context

- Some actions in the task configurations should not be run when evaluating the Gradle script.


reecexlm (approved):
LGTM

PR #1756, opened 2025-06-11
[ANCHOR-1068] Merge with `develop` post `c-account` merge

lijamie98:
### Description

- Merge with `develop` 

### Context

- Branch maintainance.


JiahuiWho (approved):
LGTM

core/src/main/java/org/stellar/anchor/ledger/LedgerTransaction.java
```diff
@@ -60,11 +62,17 @@ public static class LedgerPaymentOperation implements LedgerPayment {
...
+    @Override
+    public OperationType getType() {
+      return OperationType.PAYMENT;
+    }
   }
 
   @Builder
   @Data
   public static class LedgerPathPaymentOperation implements LedgerPayment {
+    OperationType type;
```
philipliu, line 75: Do we need to override `getType`?

philipliu (approved):

core/src/main/java/org/stellar/anchor/ledger/LedgerTransaction.java
lijamie98: `getType` is overriden by `@Data`.

In this case, we cannot skip the `OperationType` by overriding the `getType()` because it can be PATH_PAMENT_STRICT_SEND or PATH_PAYMENT_STRICT_RECEIVE. The value is assigned in the builder.

PR #1764, opened 2025-06-30
[ANCHOR-937] Add Horizon configuration to end-2-end tests

lijamie98:
### Description

The default test configuration is set to Soroban Rpc server through `stellar_network.rpc_url`. We should add e-2-e tests for Horizon server.

- When SEP-45 is required in Horizon test configuration, the test will be skipped.
- Add Horizon configuration to end-2-end tests

### Context

- Test refactoring


.github/workflows/sub_essential_tests.yml
```diff
@@ -66,70 +66,92 @@ jobs:
...
         with:
-          url: "http://localhost:8092/.well-known/stellar.toml"
-          expected-status: "200"
-          timeout: "300"
+          url: "http://host.docker.internal:8080/.well-known/stellar.toml"
           interval: "1"
 
-      #
-      # Run Essential Tests
-      #
```
philipliu, line 113: nit: I think we can remove comments like this from the file since most of the time, it's the same as the `name`.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep31End2EndTests.kt
```diff
@@ -127,34 +127,61 @@ open class Sep31End2EndTests : IntegrationTestBase(TestConfig()) {
...
-        ?.map { it.key!! }
-        .orEmpty()
-    wallet.sep12.putCustomer(
-      gson.fromJson(
-        gson.toJson(additionalRequiredFields.associateWith { receiverKycInfo[it]!! }),
-        Sep12PutCustomerRequest::class.java,
-      )
+    waitStatuses(
+      postTxResponse.id,
+      listOf(SepTransactionStatus.PENDING_CUSTOMER_INFO_UPDATE, SepTransactionStatus.COMPLETED)
```
philipliu, line 132: If the status is completed before KYC is supplied, we either have a race condition somewhere or there is a bug. I don't think we should allow `COMPLETED` here.

service-runner/src/main/resources/profiles/default-horizon/config.env
```diff
@@ -1,2 +1,3 @@
+stellar_network.rpc_url=
```
philipliu, line 1: We still have a `default` profile that uses a combination of Horizon and RPC, right? Does it make sense to keep that profile if we have Horizon-only and RPC-only modes?

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep31End2EndTests.kt
lijamie98: When the Sep31End2EndTests are run the second time without cleaning up the database, the status will not transition to `PENDING_CUSTOMER_INFO_UPDATE` because the customer is already registered in the previous tests. 

If we want to avoid this, we need to create different customer information every time we run the tests. In my opinion, it is ok. But If we don't like the way it is now, we may create a different ticket.

service-runner/src/main/resources/profiles/default-horizon/config.env
lijamie98: The `default` is the template for all the other tests. It is default to rpc-only mode so we only need horizon-only and there is no rpc-only mode.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/e2etest/Sep31End2EndTests.kt
philipliu: I agree that it's an issue we should address, but we should avoid making this change in this PR, as it could unintentionally hide future bugs in SEP-31.

lijamie98: I make this test repeats x2 to test both routes.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -479,9 +481,24 @@ internal fun createContractWithWasmIdAndGetContractId(
...
+      .hostFunction
+      .createContractV2
+      .contractIDPreimage
+  return StrKey.encodeContract(preImage.fromAddress.address.accountId.accountID.toXdrByteArray())
+  //
+  //
+  //  return StrKey.encodeContract(
+  //
+  // getTransactionResponse.parseResultMetaXdr().v3.sorobanMeta.returnValue.address.contractId.hash
+  //  )
```
philipliu: Looks like the tx meta v4 issue has been fixed so this test no longer needs to be disabled

philipliu (approved):
LGTM

PR #1768, opened 2025-07-14
[ANCHOR-1077] SEP-6 config does not respect SEP-38 config

lijamie98:
### Description

If sep-38 is not not enabled:

- Remove the deposit-exchange and withdraw-exchange from the SEP-6 /info response
- Return 501 not implemented 

### Context

- Bug fixes

### Testing

- `./gradlew test`

### DIFF
--- a/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
+++ b/core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
@@ -186,7 +186,7 @@ public class Sep6Service {
     }
     if (!sep38Config.isEnabled()) {
       throw new SepNotImplementedException(
-          "SEP-38 is not enabled. Please enable SEP-38 to use deposit-exchange.");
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/stellar/anchor/sep6/Sep6Service.java
```diff
@@ -179,6 +184,10 @@ public StartDepositResponse depositExchange(Sep10Jwt token, StartDepositExchange
     if (request == null) {
       throw new SepValidationException("missing request");
     }
+    if (!sep38Config.isEnabled()) {
+      throw new SepNotImplementedException(
+          "SEP-38 is not enabled. Please enable SEP-38 to use deposit-exchange.");
```
philipliu, line 189: Can we change this to "deposit-exchange is not enabled`? This is the message that is shown to client so they have don't have control over whether SEP-38 is enabled by the anchor.

philipliu (approved):

PR #1777, opened 2025-07-23
[ANCHOR-1105] Fix Fix sep24 query returns multiple results

JiahuiWho:
### Description

This pr fixes issue: 
1. observer not processing transactions due to `findOneByToAccountAndFromAccountAndStatus` could return multiple results if multiple transactions from the same account were stuck in the same status (e.g., dropped out at `pending_user_transfer_start`).

2. `expectedAmount` could be null


JiahuiWho:
I understand the concern about Prometheus metrics, but at this point, it's outside the scope of what I'm trying to do.
I looked into this and helped identify the root cause to help unblock my work, but the Prometheus metrics aren’t something I’m currently focused on.

 If anyone were affected by recent changes, it might make sense for the relevant owner to take a look and follow up.

lijamie98:
I think the title of this PR should be `Sep24 query return multiple results`

PR #1778, opened 2025-07-30
[ANCHOR-1109] Add filters based on the distribution account

lijamie98:
### Description

- Add filters based on the distribution account

### Context

- Improve the filter efficiency

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -244,7 +244,7 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
     }
   }
 
-  private GetEventsRequest buildEventRequest(String cursor) throws IOException {
+  GetEventsRequest buildEventRequest(String cursor) throws IOException {
     List<String> uniqueDistributionAccounts =
         assetService.getStellarAssets().stream()
             .filter(asset -> asset.getSchema() == AssetInfo.Schema.STELLAR)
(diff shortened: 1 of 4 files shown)


platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```diff
@@ -236,21 +245,55 @@ private ShouldProcessResult shouldProcess(EventInfo event) {
...
-            .build();
+    List<String> uniqueDistributionAccounts =
+        assetService.getStellarAssets().stream()
+            .filter(asset -> asset.getSchema() == AssetInfo.Schema.STELLAR)
+            .map(StellarAssetInfo::getDistributionAccount)
+            .distinct()
+            .toList();
+
+    List<EventFilter> filters =
+        uniqueDistributionAccounts.stream()
```
philipliu, line 256: Do we also want to filter it by assets?

platform/src/test/kotlin/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserverTest.kt
```diff
@@ -42,6 +44,7 @@ class StellarRpcPaymentObserverTest {
     paymentObservingAccountsManager = mockk(relaxed = true)
     paymentStreamerCursorStore = mockk(relaxed = true)
     sacToAssetMapper = MockSacToAssetMapper()
+    assetService = mockk(relaxed = true)
```
philipliu, line 47: Can we update the tests? I think asserting that the request includes the filter is good enough.

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
lijamie98: Although adding asset filter may feel more efficient, it also makes it more brittle while the efficiency gain is debatably marginal. 

I would keep it this way unless it has obvious advantage.

platform/src/test/kotlin/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserverTest.kt
lijamie98: tests added

PR #1779, opened 2025-08-01
[ANCHOR-1111] Add Stellar RPC authorization configuration

lijamie98:
### Description

- Add Stellar RPC authorization configuration

### Context

- Improve compatibility

### Testing

- `./gradlew test`


platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,11 +37,37 @@ stellar_network:
...
+      http_header: Aurhorization
+    # If the type is `url`, the following configuration can be set to configure how the API key is added to the URL.
+    url:
+      # The `type` can be one of the following:
+      #   `path_append`: The API key is appended to the URL path.
+      #       For example, QuickNode appends the API key to the URL path like this: `https://bitter-few-friday.stellar-testnet.quiknode.pro/{YOUR_API_KEY}/`.
+      #   `query_param`: The API key is added as a query parameter to the URL.
+      type: path_append
+      # If the type is `query_param`, the `query parameter name` must be set.
+      query_param_name: api_key
```
philipliu, line 69: Can't the full URL be set in the `rpc_url`?

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,11 +37,37 @@ stellar_network:
...
+    # The `type` can be one of the following:
+    #   `none`: No authentication is used.
+    #   `api_key`: The authentication is done using an API key added to the HTTP request header.
+    #   `url`: The authentication is done using an API key added to the URL path.
+    # If the type is `api_key` or `url_path`, the secret must be set by the environment variable `secret.stellar_network.rpc.auth.secret`.
+    type: none
+    # If the type is `api_key`, the following configuration can be set to configure how the API key is added to the HTTP request header.
+    x_api_key:
+      # The HTTP header name of the API key
+      http_header: Aurhorization
```
philipliu, line 60: Typo: Authorization

platform/src/main/resources/config/anchor-config-default-values.yaml
```diff
@@ -37,11 +37,37 @@ stellar_network:
...
   # The rpc server endpoint.
-  # TODO: Take care of authentication
   # example:
   #   rpc_url: https://soroban-testnet.stellar.org
   rpc_url:
+  rpc_auth:
+    # The rpc_auth is used to authenticate with the Soroban RPC server.
+    # The `type` can be one of the following:
+    #   `none`: No authentication is used.
+    #   `api_key`: The authentication is done using an API key added to the HTTP request header.
```
philipliu, line 53: We call it `api_key` here, but the config block is called `x_api_key` below.

platform/src/main/resources/config/anchor-config-default-values.yaml
JiahuiWho, line 60: Do you anticipate adding more fields under `x_api_key` in the future? f not, it might be clearer to flatten the structure by moving `http_header` directly under `rpc_auth`?

lijamie98: It should be `X-Api-Key`

PR #1782, opened 2025-08-07
[ANCHOR] Update sep helm values according to rpc config change

JiahuiWho:
Need to remove from local values as well

### DIFF
--- a/helm-charts/sep-service/values.yaml
+++ b/helm-charts/sep-service/values.yaml
@@ -75,8 +75,11 @@ config:
 
   stellar_network:
     network: TESTNET
-    network_passphrase: "Test SDF Network ; September 2015"
+    type: rpc
     horizon_url: "https://horizon-testnet.stellar.org"
+    rpc_url: https://soroban-testnet.stellar.org
+    rpc_auth:
+      type: none
 
   data:
     type: postgres


philipliu (approved):

helm-charts/sep-service/values.yaml
```diff
@@ -75,8 +75,11 @@ config:
 
   stellar_network:
     network: TESTNET
-    network_passphrase: "Test SDF Network ; September 2015"
+    type: rpc
```
lijamie98, line 78: The `type` had not been implemented yet. This will be ignored.

lijamie98 (changes requested):

helm-charts/sep-service/values.yaml
lijamie98: It will be default to `rpc`. 
https://github.com/stellar/anchor-platform/blob/develop/platform/src/main/java/org/stellar/anchor/platform/component/share/UtilityBeans.java#L109

We probably want to wait until ANCHOR-1103 is merged.

JiahuiWho: I'm only fixing the invalid config error throwed here: [[ANCHOR] Fix pods crashing due to recent RPC config change](https://github.com/stellar/kube/pull/3440)

helm-charts/sep-service/values.yaml
```diff
@@ -75,8 +75,11 @@ config:
...
     horizon_url: "https://horizon-testnet.stellar.org"
+    rpc_url: https://soroban-testnet.stellar.org
```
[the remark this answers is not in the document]
JiahuiWho: Isn't the comment saying it's used for observer.
>   # If the rpcURL is set, the horizonURL is not used other than the observer.

PR #1856, opened 2025-11-21
Fix: Wrap all RPC exceptions in try/catch in the observer

philipliu:
### Description

This fixes the observer stalling when RPC returns a transient error.

### Context

Exceptions coming from the SDK are not mapped to IOExceptions so they are not handled, killing the thread. The RPC observer implementation does not set `lastActivityTime` or change the status so the health checks stay green.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A


lijamie98:
Can we add a quick unit test for this error handling?

lijamie98 (approved):
minor suggestions. please feel free to merge after resolve these comments

PR #1918, opened 2026-04-08
[ANCHOR-1187] Fix `MEMO_ID` validation to support full Stellar uint64 range

JiahuiWho:
### Description

  - Fix MEMO_ID validation to support the full Stellar uint64 range (0 to 18,446,744,073,709,551,615)
  - Replace Long.parseLong and Long.longValue() with BigInteger for MEMO_ID parsing and conversion
  - Consolidate memo ID creation into a shared MemoHelper.makeMemoId() method   

### Context

Partner reported that SEP-24 requests with refund_memo 11872666534918305457 were rejected with "Invalid Memo"  due to this refund_memo value above Java's `Long.MAX_VALUE `(9,223,372,036,854,775,807) 

Stellar protocol defines `MEMO_ID` as uint64, but the platform was using Java's signed long for parsing, which only supports half the range.

The same issue existed in SEP-10 memo validation, `xdrMemoToString` (used by the payment observer for memo matching), and muxed account memo handling in `DefaultPaymentListener`

### Testing

  - ./gradlew test
  - Added new unit tests to cover all cases

### Documentation

N/A

### Known limitations

N/A


marwen-abid (approved):
LGTM

PR #1922, opened 2026-04-09
[ANCHOR-1190] Fix Docker release to publish multi-arch images    

JiahuiWho:
### Description

Each release job now pushes to a temporary arch-specific tag (-amd64, -arm64), and a new `create_multi_arch_manifest` job combines them into a single multi-arch manifest.

Same for `on_push_to_develop`

### Context

Since 4.1.1, the release workflow has been publishing only one architecture per release due to a race condition, the amd64 and arm64 jobs both push to the same Docker tag, so whichever finishes last overwrites the other.

### Testing

- `./gradlew test`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/.github/workflows/on_push_to_develop.yml
+++ b/.github/workflows/on_push_to_develop.yml
@@ -51,7 +51,7 @@ jobs:
           build-args: |
             TARGETARCH=amd64
           tags: |
-            stellar/anchor-platform:edge,stellar/anchor-platform:edge-${{ steps.get_date.outputs.DATE }}-${{ steps.get_sha.outputs.SHA }}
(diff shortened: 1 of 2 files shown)


.github/workflows/on_push_to_develop.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           build-args: |
```
marwen-abid, line 54: The CodeQL alert on both files is correct. We should probably limit permissions by adding the following after `on:` and before `jobs:` 

```
  permissions:
    contents: read
```

marwen-abid (approved):

marwen-abid (approved):

PR #1924, opened 2026-04-20
[ANCHOR -1195] Fix error serialization when Horizon look up fails

JiahuiWho:
### Description

Include HTTP status code and response body in the error message string instead of passing the exception object, matching the existing pattern in `RpcService`. Remove the dangerous `(String, Throwable)` constructors to prevent this class of bug.

### Context

When Horizon fails (e.g., rate limiting, network issues), the platform tries to return an error but crashes on serializing the exception, resulting in an unhelpful HTTP 500 (`HttpMessageNotWritableException`).

Root cause: `InternalErrorException(message, Throwable)` looks like it sets the Java exception cause, but actually passes the `Throwable` into `RpcException`'s `additionalData` field (an `Object`). Spring then tries to JSON-serialize that `Throwable` and fails because `Throwable#detailMessage` is a private field.

### Testing

- `./gradlew test`
- Updated assertions in `NotifyOnchainFundsReceivedHandlerTest`, `NotifyOnchainFundsSentHandlerTest`

### Documentation

N/A

### Known limitations

N/A

### DIFF
--- a/api-schema/src/main/java/org/stellar/anchor/api/exception/rpc/InvalidParamsException.java
+++ b/api-schema/src/main/java/org/stellar/anchor/api/exception/rpc/InvalidParamsException.java
@@ -10,8 +10,4 @@ public class InvalidParamsException extends RpcException {
   public InvalidParamsException(String message) {
     super(INVALID_PARAMS, message);
   }
-
-  public InvalidParamsException(String message, Throwable e) {
-    super(INVALID_PARAMS, message, e);
-  }
 }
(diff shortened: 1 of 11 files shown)


api-schema/src/main/java/org/stellar/anchor/api/exception/rpc/InvalidParamsException.java
```diff
@@ -10,8 +10,4 @@ public class InvalidParamsException extends RpcException {
   public InvalidParamsException(String message) {
     super(INVALID_PARAMS, message);
   }
-
-  public InvalidParamsException(String message, Throwable e) {
-    super(INVALID_PARAMS, message, e);
```
marwen-abid, line 15: nit: maybe it's worth also removing `RpcException(RpcErrorCode errorCode, String message, Object additionalData)` and removing the `additionalData` field because they're not used anymore and can lead to future mistakes

marwen-abid (approved):
LGTM!

PR #1935, opened 2026-05-12
[ANCHOR-1934] enforce single use on sep38 quotes

### LINKED ISSUE #1934: Enforce single-use semantics on SEP-38 firm quotes
### Context                                                                                           
                                                                                                     
SEP-38 has two quote endpoints:
  - `GET /price` returns an indicative rate, non-binding, for estimation.
  - `POST /quote` returns a firm rate. The anchor commits liquidity at a fixed price until `expires_at`.
The response includes a `quote_id` that the user then references when creating a SEP-31 / SEP-24 / SEP-6 transaction.

The intended lifecycle is: create quote → reference it once in a transaction within the expiry window → quote is consumed. Today the platform implements steps 1 and 2 but never marks step 3, so the quote remains redeemable indefinitely until expiry.

The bug affects all three exchange protocols because they share Sep38QuoteStore and route validation through `ExchangeAmountsCalculator.validateQuoteAgainstRequestInfo` (SEP-6, SEP-24) `Sep31Service.preValidateQuote` (SEP-31). 

### Exploit scenario

1. Alice (SEP-10 authenticated) calls POST /sep38/quote: sell 10,000 USDC, buy XLM, 30-minute expiry. Anchor returns quote Q at 5 XLM/USDC (50,000 XLM total).
2. XLM drops 20% on the open market within the window.
3. Alice creates T1 with quote_id = Q, sends 10,000 USDC, settles at 50,000 XLM.
4. Alice creates T2, T3, … each with the same quote_id, sending fresh USDC each time. Each settles at the now-stale 50,000 XLM payout. Anchor pays N × 50,000 XLM at the locked rate while market-fair would be N × 40,000 XLM. Operator eats the delta.

### Required changes

Enforce single-use at three layers — all three must land together:                                 
                                                                                                       
1. Validator check. Reject in `ExchangeAmountsCalculator.validateQuoteAgainstRequestInfo` and  `Sep31Service.preValidateQuote` when quote.getTransactionId() != null.
2. Atomic bind. In SEP-31 / SEP-24 / SEP-6 transaction-creation paths, mark the quote consumed via a  conditional UPDATE on `exchange_quote` (set `transaction_id only` where it's currently NULL), inside the same DB transaction that persists the SEP transaction. Zero ro`ws affected → reject and roll back. This closes the TOCTOU race that a plain read-then-write would leave open. Implement at the shared `Sep38QuoteStore` layer so cross-protocol attempts are covered.                                        
3. DB backstop. Migration adding `UNIQUE` on `exchange_quote.transaction_id`` for defense in depth.

### Tests to add                                                                                         
                                                                                                     
Follow the existing convention in `core/src/test/kotlin/...` (MockK on Sep38QuoteStore; see `ExchangeAmountsCalculatorTest.kt` and `Sep31ServiceTest.kt`).

- Validator rejects when `quote.getTransactionId() != null` (both validators).
- Validator still passes for unbound quote (regression).
- Sep38QuoteStore atomic-bind method returns true once, false on second call.
- Integration: SEP-31 → SEP-31 reuse rejected; SEP-31 → SEP-24 reuse rejected; SEP-31 → SEP-6 reuse rejected.
- Cancellation regression: bind T1, transition to error/cancelled/expired, attempt to bind T2 → still rejected.
- Concurrency: two parallel `POST /sep31/transactions` with same `quote_id` → exactly one 201, one 400. 
- Migration test: `UNIQUE` constraint rejects duplicate non-null `transaction_id` inserts on supported DBs. 

### Acceptance criteria

- Validator check, atomic bind, and UNIQUE migration all landed.
- Same `quote_id` cannot be bound twice within SEP-31, SEP-24, or SEP-6.
- Same `quote_id` cannot be bound across different SEPs.
- Cancelled / refunded / expired transactions do not free the quote.
- Concurrent transaction-creation requests with the same `quote_id` result in exactly one success and one rejection.
- Existing SEP-31 / SEP-24 / SEP-6 happy paths continue to pass.

amandagonsalves:
### Description
This PR solves #1934 by adding `bindToTransaction` to `Sep38QuoteStore` interface and `bindQuoteToTransaction` to `ExchangeAmountsCalculator`, `Sep6Service` and `Sep24Service`. It throws if a quote is already bound to a transaction.

#### Changes
- [x] Validator check. Reject in ExchangeAmountsCalculator.validateQuoteAgainstRequestInfo and Sep31Service.preValidateQuote when quote.getTransactionId() != null.
- [x] Atomic bind. In SEP-31 / SEP-24 / SEP-6 transaction-creation paths, mark the quote consumed via a conditional UPDATE on exchange_quote (set transaction_id only where it's currently NULL), inside the same DB transaction that persists the SEP transaction. Zero rows affected → reject and roll back. This closes the TOCTOU race that a plain read-then-write would leave open. Implement at the shared Sep38QuoteStore` layer so cross-protocol attempts are covered.
- [x] DB backstop. Migration adding UNIQUE on `exchange_quote.transaction_id`` for defense in depth.

#### Acceptance Criteria
- [x] Validator check, atomic bind, and UNIQUE migration all landed.
- [x] Same quote_id cannot be bound twice within SEP-31, SEP-24, or SEP-6.
- [x] Same quote_id cannot be bound across different SEPs.
- [x] Cancelled / refunded / expired transactions do not free the quote.
- [x] Concurrent transaction-creation requests with the same quote_id result in exactly one success and one rejection.
- [x] Existing SEP-31 / SEP-24 / SEP-6 happy paths continue to pass.

### Context
#1934 

### Testing
- [x] Validator rejects when quote.getTransactionId() != null (both validators).
- [x] Validator still passes for unbound quote (regression).
- [x] Sep38QuoteStore atomic-bind method returns true once, false on second call.
- [x] Integration: SEP-31 → SEP-31 reuse rejected; SEP-31 → SEP-24 reuse rejected; SEP-31 → SEP-6 reuse rejected.
- [x] Cancellation regression: bind T1, transition to error/cancelled/expired, attempt to bind T2 → still rejected.
- [x] Concurrency: two parallel POST /sep31/transactions with same quote_id → exactly one 201, one 400.
- [x] Migration test: UNIQUE constraint rejects duplicate non-null transaction_id inserts on supported DBs.


platform/src/main/java/org/stellar/anchor/platform/data/JdbcSep38QuoteRepo.java
```diff
@@ -1,9 +1,21 @@
 package org.stellar.anchor.platform.data;
 
 import java.util.Optional;
+import org.springframework.data.jpa.repository.Modifying;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 import org.springframework.lang.NonNull;
+import org.springframework.transaction.annotation.Transactional;
```
JiahuiWho: Let's unify this Annotation. `Sep24Service` and `Sep6Service` import `jakarta.transaction.Transactional` while here imports `org.springframework.transaction.annotation.Transactional` 

Both work, just pick one.

platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
```diff
@@ -250,6 +250,8 @@ Sep31Service sep31Service(
       RateIntegration rateIntegration,
       Clock clock,
       EventService eventService) {
+    ExchangeAmountsCalculator exchangeAmountsCalculator =
+        new ExchangeAmountsCalculator(sep38QuoteStore, clock);
```
JiahuiWho: `SepBeans` now has three new `ExchangeAmountsCalculator(sep38QuoteStore, clock)`. This is a pre-existing pattern, but a single `@Bean ExchangeAmountsCalculator(...)` would be cleaner and dedupe state if the calculator ever holds any.

JiahuiWho (commented):
LGTM!

platform/src/main/java/org/stellar/anchor/platform/component/sep/SepBeans.java
JiahuiWho: Here you forgot this one 😆

JiahuiWho (dismissed):

JiahuiWho (approved):
Thanks for working on this!

PR #1942, opened 2026-05-19
[ANCHOR-1939]: add ambiguous routing safety

### LINKED ISSUE #1939: Per-client event isolation + ambiguous routing safety
### Context

Two independent gaps in the default v4.3.0 configuration chain into a deterministic exploit that strands victim withdrawal payments.                                       
                                          
Leak side. `EventProcessorManager` spawns one `ClientStatusCallbackProcessor` per entry in `clients.yaml` with `callback_enabled = true`, each in its own Kafka consumer group on the shared `TRANSACTION` topic (`KafkaSession.java`:202). Different consumer groups = broadcast. `ClientStatusCallbackHandler.handleEvent`(`ClientStatusCallbackHandler.java`:74-91) then dispatches to its own configured callback URL without checking whether `event.getTransaction().getClientName()` / `getClientDomain()` matches this handler's client. Net effect: every configured wallet's callback receiver gets a webhook for every transaction across every other wallet, including the auto-generated `withdraw_memo`.
                                                                                                                                                                             
Routing side. `DefaultPaymentListener.processAndDispatchLedgerPayment` uses `findOneByWithdrawAnchorAccountAndMemoAndStatus(...)` to map incoming on-chain payments to SEP-24 rows. Declared single-result, no backing uniqueness constraint. On > 1 match Hibernate throws `NonUniqueResultException`, the surrounding `catch (Exception ex)` silently wallows it, and the SSE cursor advances. Same shape for SEP-6 (line 191) and SEP-31's `findByStellarMemoAndStatus`.

### Exploit scenario           
                                                                                                                                                                             
1. Anchor has two callback-enabled wallets in `clients.yaml`: `victim_wallet` and `attacker_wallet`.
2. Alice (under victim_wallet) creates a default SEP-24 withdrawal. `Sep24DepositInfoSelfGenerator` writes memo M and the row advances to `pending_user_transfer_start`.    
3. `TRANSACTION_STATUS_CHANGED` event is published with memo=M. `attacker_wallet`'s handler reads  it and POSTs to attacker's callback URL.                                       
4. Bob (under attacker_wallet) creates a SEP-24 withdrawal with `memo_type=id, memo=M`. The self-generator forwards the user-supplied memo verbatim. Bob's row also lands at `pending_user_transfer_start`.                
5. Alice sends a real on-chain payment to the anchor's distribution account with memo M. Observer matches two rows → `NonUniqueResultException` → caught and swallowed → SSE cursor advances.                            
6. Alice's asset has settled to the anchor's distribution account but no SEP-24 row advances. Funds stranded; manual operator reconciliation required.                     
                                              
### Required changes                                                                                                                                                           
                             
Two changes: (1) closes the reported exploit chain, (2) is defense-in-depth for the silent-drop failure mode independent of attacker model.                               
                 
                                                                                                                                                                             
1. Per-client event filter at the consumer. In `ClientStatusCallbackHandler.handleEvent`, short-circuit when the event does not belong to this handler's client.
2. Loud-fail on ambiguous routing in `DefaultPaymentListener`. Replace  `findOneByWithdrawAnchorAccountAndMemoAndStatus` with
  `findAllByWithdrawAnchorAccountAndMemoAndStatus`for SEP-24, SEP-6 , and the SEP-31 analogue (`findByStellarMemoAndStatus`). On `size() > 1`: log at ERROR with the payment id, source tx hash, the matched routing tuple, and the matched row IDs; increment a `payment_observer.ambiguous_routing` metric (you need to create this metrix); do **not** route the payment; do allow the surrounding flow to advance the SSE cursor (i.e. return from `processAndDispatchLedgerPayment` rather than throwing). The point is to make the failure observable for operator reconciliation, not to block the payment stream.
   

                                                                                                                                                                             
### Tests to add                                                                                                                                                               
- ClientStatusCallbackHandlerTest — handler bound to client A, event with `clientName=B`: no HTTP request issued. Event with `clientName=A`: request fires.                                                                                 
- ClientStatusCallbackHandlerTest — event with null `clientName` and null `clientDomain` must not match any handler (events with no client attribution should not leak to any callback URL).                                                                                                                                                             
- ClientStatusCallbackHandlerTest — customer event path: same filter logic applies on `event.getCustomer()`.
- DefaultPaymentListenerTest — seed two pending SEP-24 rows with the same (`withdraw_anchor_account`, `memo`, `status='pending_user_transfer_start'`) tuple, feed `onReceived`, assert 
  - (a) no `notifyOnchainFundsReceived` RPC is fired, 
  - (b) the `payment_observer.ambiguous_routing` metric is incremented, 
  - (c) the ERROR log line is emitted with both row IDs, 
  - (d) `processAndDispatchLedgerPayment` returns normally so the SSE cursor advances. 
  - (e) Mirror cases for SEP-6 and SEP-31.
- DefaultPaymentListenerTest — single-match regression: one pending row matches the routing tuple, payment routes as before, metric not incremented.                       
- Integration (`Sep24Tests` or new `MultiClientCallbackTests` under `essential-tests/`) — configure two callback-enabled clients in `clients.yaml` with separate mock callback receivers; drive a SEP-24 withdrawal under one client; assert only that client's receiver gets the webhook.                          
- Regression: existing single-client SEP-24/SEP-6/SEP-31 callback paths continue to pass.                                                                                  
                                              
### Acceptance criteria                                                                                                                                                        
                             
- Both changes landed.                                                                                                                                                     
- Wallet A's callback URL never receives an event whose transaction (or customer) belongs to wallet B, for any combination of client types.                                
- When the routing tuple (`withdraw_anchor_account`, `memo`, `status='pending_user_transfer_start'`) matches more than one pending row, the on-chain payment is **not** routed, an operator-facing ERROR log + metric are emitted with the matched row IDs, and the SSE cursor advances normally so subsequent payments continue to process.                  
- Existing SEP-24 / SEP-6 / SEP-31 happy paths and single-client callback flows continue to pass.

amandagonsalves:
## Description
  Before this change DefaultPaymentListener.processAndDispatchLedgerPayment uses findOneByWithdrawAnchorAccountAndMemoAndStatus(...) to map incoming on-chain payments to SEP-24 rows. Declared single-result, no backing uniqueness constraint. On > 1 match Hibernate throws NonUniqueResultException, the surrounding catch (Exception ex) silently wallows it, and the SSE cursor advances. Same shape for SEP-6 (line 191) and SEP-31's findByStellarMemoAndStatus.

  #### Changes
  - [x] Replace `findOneByWithdrawAnchorAccountAndMemoAndStatus` with `findAllByWithdrawAnchorAccountAndMemoAndStatus` for SEP-24 and SEP-6, and the SEP-31 analogue, in
  `DefaultPaymentListener.processAndDispatchLedgerPayment`.                                                                                                      
  - [x] On `size() > 1`: log at ERROR with payment ID, source tx hash, routing tuple, and matched row IDs; increment `payment_observer.ambiguous_routing` metric; return without routing
  (allows SSE cursor to advance).
  - [x] Add `findAllBy…` methods to `JdbcSep24TransactionRepo`, `JdbcSep6TransactionRepo`, `JdbcSep31TransactionRepo` and their store wrappers.
  - [x] Add `PAYMENT_OBSERVER_AMBIGUOUS_ROUTING` to `AnchorMetrics`.

  #### Acceptance Criteria

  - [x] When the routing tuple matches more than one pending row, the on-chain payment is not routed, an operator-facing ERROR log and metric are emitted with the matched row IDs, and
  the SSE cursor advances normally so subsequent payments continue to process.
  - [x] Existing SEP-24 / SEP-6 / SEP-31 happy paths and single-client callback flows continue to pass.

  ### Context
  #1939

  ### Testing

  - [x] `DefaultPaymentListenerTest` — seed two pending SEP-24 rows with the same routing tuple, feed `onReceived`, assert:
      - No `notifyOnchainFundsReceived` RPC is fired.
      - `payment_observer.ambiguous_routing` metric is incremented.
      - ERROR log is emitted with both matched row IDs in the varargs.
      - `processAndDispatchLedgerPayment` returns normally so the SSE cursor advances.
  - [x] `DefaultPaymentListenerTest` — mirror ambiguous cases for SEP-6 and SEP-31.
  - [x] `DefaultPaymentListenerTest` — single-match regression: one pending row matches the routing tuple, payment routes as before, `payment_observer.ambiguous_routing` metric is not
  incremented (asserted at `0.0` for SEP-24, SEP-6, and SEP-31).
  - [x] Regression: existing single-client SEP-24/SEP-6/SEP-31 callback paths continue to pass.

  ### Documentation   
Timeout for transactions changed from 10000 to 60000L because of network ingestion and tests failing.

**Release note / upgrade warning:** V28 adds partial unique indexes on `(withdraw_anchor_account, memo_type, memo)` for SEP-24 and SEP-6 rows in `pending_user_transfer_start`, and on `(to_account, stellar_memo_type, stellar_memo)` for SEP-31 rows in `pending_sender`. If duplicate rows already exist in those states, the migration will fail on startup.

Before upgrading, operators should run:
  ```sql
  -- SEP-24
  SELECT withdraw_anchor_account, memo_type, memo, COUNT(*)
  FROM sep24_transaction
  WHERE status = 'pending_user_transfer_start'
  GROUP BY withdraw_anchor_account, memo_type, memo
  HAVING COUNT(*) > 1;

  -- SEP-6
  SELECT withdraw_anchor_account, memo_type, memo, COUNT(*)
  FROM sep6_transaction
  WHERE status = 'pending_user_transfer_start'
  GROUP BY withdraw_anchor_account, memo_type, memo
  HAVING COUNT(*) > 1;

  -- SEP-31
  SELECT to_account, stellar_memo_type, stellar_memo, COUNT(*)
  FROM sep31_transaction
  WHERE status = 'pending_sender'
  GROUP BY to_account, stellar_memo_type, stellar_memo
  HAVING COUNT(*) > 1;
```

Any rows returned must be resolved (cancelled or completed) before deploying this version.

  ### Known limitations
  N/A

### DIFF
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
@@ -342,7 +342,7 @@ class PaymentObserverTests {
   private suspend fun waitForEventsCoroutine(
     fromAccountId: String,
     listener: EventCapturingListener,
-    timeout: Long = 10000L,
+    timeout: Long = 60000L,
   ) {
     val startTime = System.currentTimeMillis()
     while (System.currentTimeMillis() - startTime <= timeout) {
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java
@@ -133,7 +133,8 @@ public class DefaultPaymentListener implements PaymentListener {
             memo,
             SepTransactionStatus.PENDING_SENDER,
             ids);
-        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString()).increment();
+        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString(), "sep", "31")
+            .increment();
         return;
       }
       if (!sep31Txns.isEmpty()) {
@@ -190,7 +191,8 @@ public class DefaultPaymentListener implements PaymentListener {
             memoAsString(memo),
             SepTransactionStatus.PENDING_USR_TRANSFER_START,
             ids);
-        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString()).increment();
+        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString(), "sep", "24")
+            .increment();
         return;
       }
       if (!sep24Txns.isEmpty()) {
@@ -231,7 +233,8 @@ public class DefaultPaymentListener implements PaymentListener {
             memoAsString(memo),
             SepTransactionStatus.PENDING_USR_TRANSFER_START,
             ids);
-        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString()).increment();
+        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString(), "sep", "6")
+            .increment();
         return;
       }
       if (!sep6Txns.isEmpty()) {
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java
@@ -112,21 +112,32 @@ public class DefaultPaymentListener implements PaymentListener {
 
     try {
       String memo = xdrMemoToString(ledgerTransaction.getMemo());
-      JdbcSep31Transaction sep31Txn =
-          sep31TransactionStore.findByToAccountAndMemoAndStatus(
+      List<JdbcSep31Transaction> sep31Txns =
+          sep31TransactionStore.findAllByToAccountAndMemoAndStatus(
               ledgerPayment.getTo(), memo, SepTransactionStatus.PENDING_SENDER.toString());
-      if (sep31Txn == null) {
-        if (ledgerPayment.getTo().startsWith("M")) {
-          // Try again if the destination account is a muxed account.
-          MuxedAccount muxedAccount = new MuxedAccount(ledgerPayment.getTo());
-          sep31Txn =
-              sep31TransactionStore.findByToAccountAndMemoAndStatus(
-                  muxedAccount.getAccountId(),
-                  String.valueOf(muxedAccount.getMuxedId()),
-                  SepTransactionStatus.PENDING_SENDER.toString());
-        }
+      if (sep31Txns.isEmpty() && ledgerPayment.getTo().startsWith("M")) {
+        MuxedAccount muxedAccount = new MuxedAccount(ledgerPayment.getTo());
+        sep31Txns =
+            sep31TransactionStore.findAllByToAccountAndMemoAndStatus(
+                muxedAccount.getAccountId(),
+                String.valueOf(muxedAccount.getMuxedId()),
+                SepTransactionStatus.PENDING_SENDER.toString());
       }
-      if (sep31Txn != null) {
+      if (sep31Txns.size() > 1) {
+        List<String> ids = sep31Txns.stream().map(JdbcSep31Transaction::getId).toList();
+        errorF(
+            "Ambiguous SEP-31 payment routing: paymentId={}, txHash={}, toAccount={}, memo={}, status={}, matchedIds={}",
+            ledgerPayment.getId(),
+            ledgerTransaction.getHash(),
+            ledgerPayment.getTo(),
+            memo,
+            SepTransactionStatus.PENDING_SENDER,
+            ids);
+        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString()).increment();
+        return;
+      }
+      if (!sep31Txns.isEmpty()) {
+        JdbcSep31Transaction sep31Txn = sep31Txns.get(0);
         try {
           handleSep31Transaction(ledgerTransaction, ledgerPayment, sep31Txn);
           return;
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java [hunk 1]
```diff
@@ -156,24 +167,34 @@ && accountType(ledgerPayment.getTo()) == Muxed) {
     }
 
     try {
-      JdbcSep24Transaction sep24Txn =
-          sep24TransactionStore.findOneByWithdrawAnchorAccountAndMemoAndStatus(
```
JiahuiWho, line 160: `findOneByWithdrawAnchorAccountAndMemoAndStatus` methods stay on JdbcSep24/6TransactionRepo and stores but no caller remains. Let's delete them so nothing regresses to the unsafe path.

essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
```diff
@@ -342,7 +342,7 @@ class PaymentObserverTests {
   private suspend fun waitForEventsCoroutine(
     fromAccountId: String,
     listener: EventCapturingListener,
-    timeout: Long = 10000L,
+    timeout: Long = 60000L,
```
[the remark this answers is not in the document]
JiahuiWho: nit. revert this change?

platform/src/main/java/org/stellar/anchor/platform/observer/stellar/DefaultPaymentListener.java
```diff
@@ -112,21 +112,32 @@ void processAndDispatchLedgerPayment(
...
+        List<String> ids = sep31Txns.stream().map(JdbcSep31Transaction::getId).toList();
+        errorF(
+            "Ambiguous SEP-31 payment routing: paymentId={}, txHash={}, toAccount={}, memo={}, status={}, matchedIds={}",
+            ledgerPayment.getId(),
+            ledgerTransaction.getHash(),
+            ledgerPayment.getTo(),
+            memo,
+            SepTransactionStatus.PENDING_SENDER,
+            ids);
+        Metrics.counter(AnchorMetrics.PAYMENT_OBSERVER_AMBIGUOUS_ROUTING.toString()).increment();
```
JiahuiWho: `payment_observer.ambiguous_routing` increments for all three SEPs into one bucket. Splitting on tag (sep="24" etc.) would make on-call triage cheaper.

JiahuiWho:
The ERROR + metric turn a silent failure into a loud one, but the duplicate (`withdraw_anchor_account`, `memo`, `pending_user_transfer_start`) tuple can still be created via the `memo_type=id` path. By the time we alarm here, the victim's payment has already settled on chain and is stuck.                                         
 
 The structural fix would be a partial unique index, e.g.:                                                                   
```                                                                                                                          
   CREATE UNIQUE INDEX uq_sep24_pending_memo                                                                                   
     ON sep24_transaction (withdraw_anchor_account, memo_type, memo)                                                         
     WHERE status = 'pending_user_transfer_start';                                                                             
```
plus analogues for sep6/sep31, and mapping the violation to a 400 (memo_in_use) in `RequestOnchainFundsHandler`.              
                                                                                                                             
Happy either way, fold into this PR, or follow-up. Just flagging so we don't lose it, since it's the load-bearing fix.

JiahuiWho (approved):

platform/src/main/resources/db/migration/V28__memo_uniqueness_pending_transfer.sql
```diff
@@ -0,0 +1,11 @@
+CREATE UNIQUE INDEX uq_sep24_pending_memo
```
JiahuiWho, line 1: Just a note here: Worth a release-notes line warning operators to check for existing duplicates pre-upgrade (the unique index will refuse to create otherwise).

amandagonsalves: just added in the pr description, thank you!

Issue #5, opened 2021-12-23
Improve assertThrows in unit tests


lijamie98:
From @marcelosalloum :
FYI: assertThrows will return an instance of the thrown exception that you could use to verify the exception values, like in:
```
val thrown = assertThrows<HttpException> { request.block() }
assertEquals(HttpException(400, "Request body contains unprocessable entity.", "2"), thrown)
Also, assertEquals only works if the Exception implements the .equals(Object) method, which can be generated by @lombok.Data.
```

@lijamie98 would like to review all assertThrows in the tests.

Issue #4, opened 2021-11-30
Implement payments gateway interface for the Circle API


marcelosalloum:
- [x] Implement `ping()` method.
- [x] Implement `getDistributionAccountAddress()` method.
- [x] Implement `getAccount(String accountId)` method.
- [x] Implement `createAccount(String accountId)` method.
- [x] #40
  - [x] Payment history for `CircleWallet->CircleWallet` transfers.
  - [x] Payment history for `CircleWallet->Stellar` transfers.
  - [x] Payment history for `Stellar->CircleWallet` transfers.
  - [x] Payment history for `CircleWallet->BankWire` transfers.
  - [x] #87 
  - [x] Pagination
- [x] Implement `sendPayment(Account sourceAccount, Account destinationAccount, String currencyName, BigDecimal amount)` method.
  - [x] #3
  - [x] #21
- [x] #57
  - [x] Instructions to receive `CircleWallet<-CircleWallet` deposits.
  - [x] Instructions to receive `CircleWallet<-Stellar` deposits.
  - [x] Instructions to receive `CircleWallet<-BankWire` deposits.
    - [x] #86 
- [x] Polishes
  - [x] #2
  - [x] #26
  - [x] #27
  - [x] #84
  - [x] #88
- [x] #65

PR #1944, opened 2026-05-20
[ANCHOR-1943]: rpc observer DoS via malformed soroban transfer event

### LINKED ISSUE #1943: Anchor Platform Stellar-RPC payment observer permanent DoS via crafted Soroban transfer event with empty SCV_MAP
#### Vulnerability Summary
The Stellar-RPC payment observer in Anchor Platform parses the value field of every Soroban transfer event that matches the operator's distribution-account RPC filter. When the value is decoded as SCV_MAP, the code blindly indexes elements [0] and [1] of the map without size validation and casts them to specific SCVal types. The surrounding try/catch only catches IOException; runtime exceptions (ArrayIndexOutOfBoundsException from [0]/[1], IllegalArgumentException from Scv.fromInt128 / Scv.fromUint64 / new MuxedAccount(...), NullPointerException when the map's backing array is null) propagate up the stack.
The outer event loop (fetchEvents) catches Throwable but never advances the cursor past the poisoned event before the throw. The same window is therefore re-fetched on every 1-second tick, re-throws on the same event, and escalates through STREAM_ERROR → restartInternal() cycles with an exponential backoff timer that is documented as "not reset" between restarts. After ~10 minutes (default backoff schedule 5s → 10 → 20 → 40 → 80 → 160 → 300) the timer is maxed and the observer transitions to NEEDS_SHUTDOWN, at which point the watchdog calls shutdown() and the executor is permanently halted.
Triage-friendly impact summary: Any address on the Soroban network can spend a few cents in transaction fees to permanently freeze the on-chain payment observer of any Anchor-Platform deployment that uses Stellar-RPC mode (the modern Soroban-aware default). Once frozen, SEP-6, SEP-24 and SEP-31 withdrawals stop completing — the anchor receives the user's on-chain payment but never triggers the off-chain payout — leaving customer funds locked indefinitely until an operator restarts the process AND the malicious event ages out of the RPC's event-retention window.

#### Affected Code & Version
**Primary site — type-confusion + array OOB**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```
219    if (scValue.getDiscriminant() == SCValType.SCV_I128) {
220      amount = Scv.fromInt128(scValue).longValue();
221    } else if (scValue.getDiscriminant() == SCValType.SCV_MAP) {
222      amount = Scv.fromInt128(scValue.getMap().getSCMap()[0].getVal()).longValue();  // (1) AIOOBE / NPE / IAE
223      SCVal memoVal = scValue.getMap().getSCMap()[1].getVal();                        // (2) AIOOBE
224      eventMemo =
225          switch (memoVal.getDiscriminant()) {
226            case SCV_STRING -> memoVal.getStr().getSCString().toString();
227            case SCV_U64    -> memoVal.getU64().toString();
228            case SCV_BYTES  -> new String(Base64.getEncoder().encode(memoVal.getBytes().getSCBytes()));
229            default          -> null;
230          };
231      if (scValue.getMap().getSCMap()[1].getVal().getDiscriminant() == SCValType.SCV_U64) {
232        toAddr =
233            new MuxedAccount(
234                    Scv.fromAddress(to).toString(),
235                    Scv.fromUint64(scValue.getMap().getSCMap()[1].getVal()))           // (3) IAE if `to` is a C... contract address
236                .getAddress();
237      }
238    }
```
Four independent crash primitives are reachable on the SCV_MAP branch:
P1 — empty map → AIOOBE [0] at line 222: emit SCV_MAP with zero entries.
P2 — one-entry map → AIOOBE [1] at line 223 (and again at 231): emit SCV_MAP with one entry.
P3 — wrong first-entry type → IllegalArgumentException from Scv.fromInt128(...) at line 222: emit SCV_MAP{[0]: SCV_U64(0)}.
P4 — wrong second-entry type + contract-to → IllegalArgumentException from new MuxedAccount(...) at line 233-235 when to is a contract C… address: emit SCV_MAP{[0]: SCV_I128, [1]: SCV_U64} with topic[2] = SCV_ADDRESS containing an SC_ADDRESS_TYPE_CONTRACT.

**Narrow `catch` only handles `IOException`**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```
255  } catch (IOException ioex) {
256    warnF("Skip processing event: {}. ex={}", GsonUtils.getInstance().toJson(event), ioex.getMessage());
257    return builder.build();
258  }
```
The `try` opens at line 191 and contains the SCV_MAP indexing block above. `ArrayIndexOutOfBoundsException`, `NullPointerException`, and `IllegalArgumentException` are not subclasses of `IOException`, so they bypass the handler and propagate to `processEvents`.
Cursor never advances on poison:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java

**Cursor never advances on poison**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```
118  void fetchEvents() {
119    String cursor = (this.cursor != null) ? this.cursor : loadStellarRpcCursor();
120
121    try {
122      GetEventsResponse response = sorobanServer.getEvents(buildEventRequest(cursor));
123      lastActivityTime = Instant.now();
124      silenceTimeoutCount = 0;
125      metricLatestBlockRead.set(response.getLatestLedger());
126      if (response.getEvents() != null && !response.getEvents().isEmpty()) {
127        processEvents(response.getEvents());                                      // throws RuntimeException ...
128      }
129      // Save the cursor for the next request
130      cursor = response.getCursor();                                                // ... so we never reach here ...
131      try {
132        saveCursor(cursor);                                                         // ... or here.
133        metricLatestBlockProcessed.set(response.getLatestLedger());
134      } catch (Exception tex) { ... }
135    } catch (IOException ioex) { ... }
136    catch (NetworkException nex) { ... }
137    catch (Throwable t) {
138      errorEx("Unhandled error in RPC observer loop", t);
139      setStatus(ObserverStatus.STREAM_ERROR);                                       // → status watcher will retry forever
140    }
141  }
```
`saveCursor()` (line 369-372) is the only function that updates the in-memory `this.cursor field`. Because it is never called when `processEvents` throws, the next tick re-loads the previous cursor from DB, sends the same `getEvents(filter, cursor)` request, receives the same poisoned event, and crashes again.

**Self-shutdown escalation**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java
```
189  void checkStatus() {                              // runs every 1 second
190    switch (status) {
191      case NEEDS_SHUTDOWN:
192        infoF("shut down the observer");
193        shutdown();                                  // permanent halt; executors are terminated
194        break;
195      case STREAM_ERROR:
198        if (streamBackoffTimer.isTimerMaxed()) {
199          infoF("The streamer backoff timer is maxed. Shutdown the observer");
200          setStatus(NEEDS_SHUTDOWN);
201        } else {
204          streamBackoffTimer.backoff();              // doubles sleepSeconds until max
207          restartInternal();
208        }
```
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/AbstractPaymentObserver.java — Javadoc of `restartInternal`: "The timers are not reset." Combined with the cursor-non-advance, the backoff monotonically climbs to the configured max and the observer transitions to `NEEDS_SHUTDOWN`.

**Default timer configuration**:
stellar-org/anchor-platform/platform/src/main/resources/config/anchor-config-default-values.yaml
```
initial_stream_backoff_time: 5
max_stream_backoff_time: 300
```
stellar-org/anchor-platform/core/src/main/java/org/stellar/anchor/util/ExponentialBackoffTimer.java — `sleepSeconds = min(sleepSeconds * 2, maxSleepSeconds)`. Sequence: `5 → 10 → 20 → 40 → 80 → 160 → 300` (max reached on the 7th `backoff()` call). Total time before `isTimerMaxed()` is true and `NEEDS_SHUTDOWN` is set: ~10 minutes.

**Observer-wiring confirmation that StellarRpcPaymentObserver is the live observer in any Stellar-RPC deployment**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/component/observer/PaymentObserverBeans.java
```
69   AbstractPaymentObserver paymentObserver;
70   if (ledgerClient instanceof StellarRpc stellarRpc) {
71     paymentObserver = new StellarRpcPaymentObserver(stellarRpc, ...);    // <-- the affected observer
72   } else if (ledgerClient instanceof Horizon horizon) { ... }
```

**RPC event-filter (what the malicious event must match)**:
stellar-org/anchor-platform/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```
272    List<EventFilter> filters =
273        uniqueDistributionAccounts.stream()
274            .flatMap(distributionAccount -> Stream.of(
275                // Filter for transfers FROM the distribution account
276                EventFilter.builder()
277                    .type(EventFilterType.CONTRACT)
278                    .topic(List.of(
279                        Scv.toSymbol("transfer").toXdrBase64(),
280                        Scv.toAddress(distributionAccount).toXdrBase64(),
281                        "*", "*"))
282                    .build(),
283                // Filter for transfers TO the distribution account
284                EventFilter.builder()
285                    .type(EventFilterType.CONTRACT)
286                    .topic(List.of(
287                        Scv.toSymbol("transfer").toXdrBase64(),
288                        "*",
289                        Scv.toAddress(distributionAccount).toXdrBase64(),
290                        "*"))
291                    .build()))
292            .toList();
```
The filter is `type=contract` and does not restrict the emitter to a Stellar-Asset-Contract (SAC) instance; any Soroban contract that emits a topic-4 event with the right `transfer`-symbol + distribution-account address in topic[1] or topic[2] will be delivered to `shouldProcess`.

#### Steps to Reproduce
A self-contained PoC bundle is attached at
anchor-platform-stellar-rpc-observer-scvmap-poison-poc.
Run a single command (per the program's "use your local instance of
Horizon or Stellar-RPC and a separate network" rule, the script provisions
its own local standalone network — NOT testnet, NOT pubnet):
```
cd anchor-platform-stellar-rpc-observer-scvmap-poison-poc/
bash reproduce.sh
# (or:  bash reproduce.sh --down  to also tear the stack down at the end)
```
1. Sanity-checks toolchain (docker, docker compose, python3, jq,
curl, tar) and installs python3-cryptography if missing.
2. Generates three fresh ed25519 Stellar keypairs (SEP-10 signing,
distribution account, attacker account) and patches config/ + dev.env.
3. docker compose up -d brings up:
  - stellar/quickstart:testing — local standalone network with Horizon + Soroban-RPC,
  - postgres:15.2-alpine — anchor-platform's cursor store,
  - stellar/anchor-platform:latest — configured with STELLAR_NETWORK_TYPE=rpc so the live observer is StellarRpcPaymentObserver (the vulnerable class; see PaymentObserverBeans.java:70), and with the streamer backoff compressed to initial=2s, max=8s so the full escalation is observable in ~9 s instead of the default ~10 min.
4. Waits for /health to report stellar_rpc_payment_observer = GREEN
(asserts the pre-attack baseline).
5. Downloads stellar-cli 26.0.0 (Linux x86_64 release tarball) into
./bin/ and registers a local-network alias scoped to this PoC's
$XDG_CONFIG_HOME.
6. Funds the attacker account via friendbot (~10 000 XLM, of which only
a fraction of one XLM is consumed by the attack).
7. Uploads, deploys, and invokes the poison Soroban contract
(contract-prebuilt/poison.wasm, 875 bytes, source in contract/src/lib.rs).
The invocation emits one event:
```
Event: [{"symbol":"transfer"},
        {"address": <DIST>},
        {"address": <ATTACKER>},
        {"string": "native"}]
    value = {"map":[]}   ← empty SCV_MAP → AIOOBE at line 222
```
8. Polls /health and docker logs until all four signals are observed:

Signal | Source
-- | --
java.lang.ArrayIndexOutOfBoundsException at StellarRpcPaymentObserver.java:222 | anchor-platform logs
Setting status to STREAM_ERROR | anchor-platform logs
The streamer backoff timer is maxed. Shutdown the observer | anchor-platform logs
/health.stellar_rpc_payment_observer.status == RED | curl http://localhost:18083/health
9. Proves cursor-freeze by reading anchor-platform.stellar_payment_observer_page_token from postgres
twice (8 s and 16 s after permanent shutdown) and asserting they are
equal. A live observer would advance the cursor every second; a frozen
value means the observer is dead and cannot recover.
10. Dumps full evidence into ./evidence/ (anchor-platform.log,
observer-crash-summary.log, cursor-before.txt, cursor-after.txt,
cursor-final.txt, health-after-attack.json, keys.json,
invoke-output.log, SUMMARY.md).

Sample evidence captured by one run
evidence/SUMMARY.md from a real PoC run (full log in evidence/anchor-platform.log):
```
## Observer cursor freeze
| When                          | stellar_rpc_cursor             |
|-------------------------------|--------------------------------|
| Before attack                 | 0000000682899800063-4294967295 |
| 8s after permanent shutdown   | 0000000704374636543-4294967295 |
| 16s after permanent shutdown  | 0000000704374636543-4294967295 |

## Observer status (after attack)
{ "status": "RED" }

## Crash signature
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
    at org.stellar.anchor.platform.observer.stellar.StellarRpcPaymentObserver.shouldProcess(StellarRpcPaymentObserver.java:222)
    at org.stellar.anchor.platform.observer.stellar.StellarRpcPaymentObserver.processEvents(StellarRpcPaymentObserver.java:156)
    at org.stellar.anchor.platform.observer.stellar.StellarRpcPaymentObserver.fetchEvents(StellarRpcPaymentObserver.java:127)

## Status-transition timeline
15:52:34.092 INFO  AbstractPaymentObserver  Setting status to STREAM_ERROR
15:52:37.067 INFO  AbstractPaymentObserver  Setting status to RUNNING
15:52:37.075 INFO  AbstractPaymentObserver  Setting status to STREAM_ERROR
15:52:42.090 INFO  AbstractPaymentObserver  Setting status to RUNNING
15:52:42.141 INFO  AbstractPaymentObserver  Setting status to STREAM_ERROR
15:52:43.093 INFO  AbstractPaymentObserver  The streamer backoff timer is maxed. Shutdown the observer
15:52:43.094 INFO  AbstractPaymentObserver  Setting status to NEEDS_SHUTDOWN
15:52:44.095 INFO  AbstractPaymentObserver  Shutting down the payment observer
15:52:44.096 INFO  AbstractPaymentObserver  Setting status to SHUTDOWN
```
End-to-end runtime: ~3 minutes the first time (mostly image pull + wasm
download), ~45 seconds on subsequent runs.

Image
Image
Image
Image

** Variants reachable with the same primitive** 
Each of the following triggers the same uncaught-runtime-exception →
STREAM_ERROR escalation and is independently sufficient. The PoC contract
uses the first variant; the others can be reproduced by editing
contract/src/lib.rs and rebuilding (cd contract && cargo build --target wasm32v1-none --release):
  - Empty SCV_MAP (the PoC default) → AIOOBE at line 222 (scValue.getMap().getSCMap()[0]).
  - One-entry SCV_MAP → AIOOBE at line 223 (scValue.getMap().getSCMap()[1]) and again at line 231.
  - First entry not SCV_I128 (e.g. Map<u32, u32>) → IllegalArgumentException from Scv.fromInt128 at line 222.
  - to is a contract C… address with second map entry SCV_U64 → IllegalArgumentException from new MuxedAccount(C..., u64) at line 233-235 (MuxedAccount rejects non-G strkey).

#### Fix Recommendation
1. **Catch all unchecked exceptions in `shouldProcess` (and `processEvents`) so one malformed event never poisons the cursor**. 
```
// StellarRpcPaymentObserver.java:255
} catch (Exception ex) {                        // was: IOException ioex
  warnF("Skip processing event: {}. ex={}",
      GsonUtils.getInstance().toJson(event), ex.toString());
  return builder.build();
}
```
And, defensively, the same `Exception`-catch should wrap the per-event call in `processEvents`.
```
private void processEvents(List<EventInfo> events) {
  if (events == null || events.isEmpty()) return;
  for (EventInfo event : events) {
    try {
      ShouldProcessResult r = shouldProcess(event);
      if (r.shouldProcess) processTransferEvent(r);
    } catch (Exception ex) {
      warnF("Skip event due to error: {}. ex={}",
          GsonUtils.getInstance().toJson(event), ex.toString());
    }
  }
}
```
2. **Validate `SCMap` shape before indexing**.
```
} else if (scValue.getDiscriminant() == SCValType.SCV_MAP) {
  var entries = scValue.getMap() == null ? null : scValue.getMap().getSCMap();
  if (entries == null || entries.length < 2) return builder.build();   // bail to "skip"
  SCVal amountVal = entries[0].getVal();
  SCVal memoVal   = entries[1].getVal();
  if (amountVal.getDiscriminant() != SCValType.SCV_I128) return builder.build();
  amount = Scv.fromInt128(amountVal).longValue();
  // …
  if (memoVal.getDiscriminant() == SCValType.SCV_U64) {
    try {
      toAddr = new MuxedAccount(toAddr, Scv.fromUint64(memoVal)).getAddress();
    } catch (IllegalArgumentException iae) {
      // `to` was a contract address; leave toAddr as the unmuxed value.
    }
  }
}
```
3. **Make cursor progress poison-tolerant**. Restructure `fetchEvents` so a per-event exception cannot prevent the cursor from advancing past that batch. 
```
void fetchEvents() {
  String cursor = (this.cursor != null) ? this.cursor : loadStellarRpcCursor();
  GetEventsResponse response;
  try {
    response = sorobanServer.getEvents(buildEventRequest(cursor));
  } catch (Throwable t) {  // network / I/O only — these legitimately should retry without cursor advance
    errorEx("RPC getEvents failed", t);
    setStatus(ObserverStatus.STREAM_ERROR);
    return;
  }
  try {
    processEvents(response.getEvents());   // now per-event-try, can no longer poison
  } finally {
    try { saveCursor(response.getCursor()); }
    catch (Exception tex) { setStatus(ObserverStatus.DATABASE_ERROR); }
  }
}
```
4. **Reset `streamBackoffTimer` on a successful tick**. Today the timer is only ever reset implicitly via the `publishingBackoffTimer.reset()` call in `handleEvent` (which is for a different timer). Add `streamBackoffTimer.reset()` whenever a tick completes without throwing so a transient error does not progressively wind down to `NEEDS_SHUTDOWN`.
5. **(Defense-in-depth)** Restrict the RPC filter to contracts that are known SACs for the configured assets — emitter address can be checked against the operator's allow-list of SAC instance addresses. Today the filter is `type=contract` with no emitter-allowlist, so any Soroban account can deliver an event to the observer.

The first change alone closes the critical path. Changes 2–4 prevent reintroduction; change 5 hardens against an entire class of similar bugs.

amandagonsalves:
### Description

Before this change, the Stellar RPC payment observer could be permanently shut down by any Soroban contract emitting a `transfer` event with a malformed `value` field.

The `shouldProcess` method indexed `SCV_MAP` entries at `[0]` and `[1]` without bounds or type validation, and the surrounding `catch` only handled `IOException` — letting `ArrayIndexOutOfBoundsException`, `IllegalArgumentException`, and `NullPointerException` propagate.

Because `saveCursor` is called after `processEvents`, a single poison event stalled the cursor, causing the same event to be replayed on every 1-second tick until the exponential backoff timer maxed out (~10 min) and the observer called `shutdown()`. All SEP-6/24/31 withdrawals stopped completing until the process was manually restarted.

**Changes**
- [x] Validate `SCV_MAP` shape in `shouldProcess` before indexing: null-check `getMap()`, require `entries.length >= 2`, require `entries[0]` discriminant is `SCV_I128`; return a skip result early otherwise.
- [x] Catch `IllegalArgumentException` from `new MuxedAccount(C…, u64)` when `to` is a contract address; log a warning and fall back to the unmuxed address.
- [x] Broaden `catch (IOException)` to `catch (Exception)` in `shouldProcess` so all unchecked runtime exceptions are logged and skipped instead of propagating.
- [x] Wrap each event in `processEvents` with its own `try/catch(Exception)` so one bad event cannot abort the entire batch.
- [x] Move `saveCursor` into a `finally` block in `fetchEvents` so the cursor always advances when `getEvents` succeeds, preventing infinite replay of a poison event.

**Acceptance Criteria**
- [x] An empty `SCV_MAP` event does not crash the observer or set `STREAM_ERROR`.
- [x] A one-entry `SCV_MAP` event does not crash the observer.
- [x] An `SCV_MAP` whose first entry is not `SCV_I128` does not crash the observer.
- [x] A contract-address recipient with `SCV_U64` memo logs a warning and completes without crashing.
- [x] In all four cases above, the observer status remains `RUNNING` and the cursor advances past the poison event.
- [x] Existing SEP-24 / SEP-6 / SEP-31 happy paths and single-client callback flows continue to pass.

### Context

[#1943](https://github.com/stellar/anchor-platform/issues/1943)

### Testing
Four new unit tests in `StellarRpcPaymentObserverTest`:
- `fetchEvents skips empty SCV_MAP poison event without crashing or stalling` (P1)
- `fetchEvents skips one-entry SCV_MAP without crashing or stalling` (P2)
- `fetchEvents skips SCV_MAP with wrong first-entry type without crashing or stalling` (P3)
- `fetchEvents handles contract-address recipient with SCV_U64 memo without crashing` (P4)

Each test confirms `ObserverStatus.RUNNING` and cursor advancement after the poison event is processed.

**Integration tests**
- P1–P4 poison variants each confirm `ObserverStatus.RUNNING` and cursor advance on the live scheduler
- Health check returns `GREEN` after processing a poison event
- Mixed batch containing a poison and a valid event advances cursor past both
- Increased `waitForEventsCoroutine` default timeout from 10 s to 60 s to account for testnet ledger close time (~6 s) plus network latency
- `waitForEventsCoroutine` now fails with an explicit assertion message on timeout instead of silently returning
- `assertEventsPayment` / `assertEventsPathPayment` use `assertNotNull` before indexing so null events produce a readable failure

### Documentation
N/A

### Known limitations
N/A

### DIFF
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -235,7 +235,10 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
                 new MuxedAccount(Scv.fromAddress(to).toString(), Scv.fromUint64(memoVal))
                     .getAddress();
           } catch (IllegalArgumentException iae) {
-            warnF("Cannot build MuxedAccount for address '{}', using unmuxed value. ex={}", toAddr, iae.getMessage());
+            warnF(
+                "Cannot build MuxedAccount for address '{}', using unmuxed value. ex={}",
+                toAddr,
+                iae.getMessage());
           }
         }
       }
@@ -254,11 +257,9 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
           .eventMemo(eventMemo)
           .sep11Asset(asset.getStr().getSCString().toString())
           .build();
-    } catch (IOException ioex) {
+    } catch (Exception ex) {
       warnF(
-          "Skip processing event: {}. ex={}",
-          GsonUtils.getInstance().toJson(event),
-          ioex.getMessage());
+          "Skip processing event: {}. ex={}", GsonUtils.getInstance().toJson(event), ex.toString());
       return builder.build();
     }
   }
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -227,8 +227,16 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
       if (scValue.getDiscriminant() == SCValType.SCV_I128) {
         amount = Scv.fromInt128(scValue).longValue();
       } else if (scValue.getDiscriminant() == SCValType.SCV_MAP) {
-        amount = Scv.fromInt128(scValue.getMap().getSCMap()[0].getVal()).longValue();
-        SCVal memoVal = scValue.getMap().getSCMap()[1].getVal();
+        var entries = scValue.getMap() == null ? null : scValue.getMap().getSCMap();
+        if (entries == null || entries.length < 2) {
+          return builder.build();
+        }
+        SCVal amountVal = entries[0].getVal();
+        SCVal memoVal = entries[1].getVal();
+        if (amountVal.getDiscriminant() != SCValType.SCV_I128) {
+          return builder.build();
+        }
+        amount = Scv.fromInt128(amountVal).longValue();
         eventMemo =
             switch (memoVal.getDiscriminant()) {
               case SCV_STRING -> memoVal.getStr().getSCString().toString();
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -131,6 +131,7 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
       } finally {
         try {
           saveCursor(response.getCursor());
+          streamBackoffTimer.reset();
           metricLatestBlockProcessed.set(response.getLatestLedger());
         } catch (Exception tex) {
           warnF("Failed to persist RPC cursor. Will retry next tick. ex={}", tex.getMessage());
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -82,6 +82,7 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
   @Override
   void shutdownInternal() {
     task.cancel(true);
+    executorService.shutdownNow();
     status = ObserverStatus.SHUTDOWN;
   }
 
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -123,17 +123,18 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
       lastActivityTime = Instant.now();
       silenceTimeoutCount = 0;
       metricLatestBlockRead.set(response.getLatestLedger());
-      if (response.getEvents() != null && !response.getEvents().isEmpty()) {
-        processEvents(response.getEvents());
-      }
-      // Save the cursor for the next request
-      cursor = response.getCursor();
       try {
-        saveCursor(cursor);
-        metricLatestBlockProcessed.set(response.getLatestLedger());
-      } catch (Exception tex) {
-        warnF("Failed to persist RPC cursor. Will retry next tick. ex={}", tex.getMessage());
-        setStatus(ObserverStatus.DATABASE_ERROR);
+        if (response.getEvents() != null && !response.getEvents().isEmpty()) {
+          processEvents(response.getEvents());
+        }
+      } finally {
+        try {
+          saveCursor(response.getCursor());
+          metricLatestBlockProcessed.set(response.getLatestLedger());
+        } catch (Exception tex) {
+          warnF("Failed to persist RPC cursor. Will retry next tick. ex={}", tex.getMessage());
+          setStatus(ObserverStatus.DATABASE_ERROR);
+        }
       }
     } catch (IOException ioex) {
       warnF(
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -153,9 +153,16 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
     debugF("Processing {} 'transfer' events", events.size());
 
     for (EventInfo event : events) {
-      ShouldProcessResult result = shouldProcess(event);
-      if (result.shouldProcess) {
-        processTransferEvent(result);
+      try {
+        ShouldProcessResult result = shouldProcess(event);
+        if (result.shouldProcess) {
+          processTransferEvent(result);
+        }
+      } catch (Exception ex) {
+        warnF(
+            "Skip event due to unexpected error: {}. ex={}",
+            GsonUtils.getInstance().toJson(event),
+            ex.toString());
       }
     }
   }
--- a/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
+++ b/platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
@@ -229,12 +229,14 @@ public class StellarRpcPaymentObserver extends AbstractPaymentObserver {
                   new String(Base64.getEncoder().encode(memoVal.getBytes().getSCBytes()));
               default -> null;
             };
-        if (scValue.getMap().getSCMap()[1].getVal().getDiscriminant() == SCValType.SCV_U64) {
-          toAddr =
-              new MuxedAccount(
-                      Scv.fromAddress(to).toString(),
-                      Scv.fromUint64(scValue.getMap().getSCMap()[1].getVal()))
-                  .getAddress();
+        if (memoVal.getDiscriminant() == SCValType.SCV_U64) {
+          try {
+            toAddr =
+                new MuxedAccount(Scv.fromAddress(to).toString(), Scv.fromUint64(memoVal))
+                    .getAddress();
+          } catch (IllegalArgumentException iae) {
+            warnF("Cannot build MuxedAccount for address '{}', using unmuxed value. ex={}", toAddr, iae.getMessage());
+          }
         }
       }
 
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
@@ -11,7 +11,6 @@ import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.Assertions.assertNull
-import org.junit.jupiter.api.Assertions.fail
 import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 import org.stellar.anchor.asset.AssetService
@@ -359,7 +358,9 @@ class PaymentObserverTests {
       }
       delay(1000)
     }
-    fail("Timed out after ${timeout}ms waiting for event from account $fromAccountId")
+    throw AssertionError(
+      "Timed out after ${timeout}ms waiting for event from account $fromAccountId"
+    )
   }
 
   private fun sendTestPayment(fromKeyPair: KeyPair, toKeyPair: KeyPair): Transaction {
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
@@ -11,6 +11,7 @@ import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.fail
 import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 import org.stellar.anchor.asset.AssetService
@@ -358,9 +359,7 @@ class PaymentObserverTests {
       }
       delay(1000)
     }
-    assertNotNull(null) {
-      "Timed out after ${timeout}ms waiting for event from account $fromAccountId"
-    }
+    fail("Timed out after ${timeout}ms waiting for event from account $fromAccountId")
   }
 
   private fun sendTestPayment(fromKeyPair: KeyPair, toKeyPair: KeyPair): Transaction {
--- a/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
+++ b/essential-tests/src/testFixtures/kotlin/org/stellar/anchor/platform/integrationtest/PaymentObserverTests.kt
@@ -347,7 +347,7 @@ class PaymentObserverTests {
   private suspend fun waitForEventsCoroutine(
     fromAccountId: String,
     listener: EventCapturingListener,
-    timeout: Long = 30000L,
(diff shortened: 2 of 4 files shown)


platform/src/main/java/org/stellar/anchor/platform/observer/stellar/StellarRpcPaymentObserver.java
```diff
@@ -123,17 +124,18 @@ void fetchEvents() {
...
-      } catch (Exception tex) {
-        warnF("Failed to persist RPC cursor. Will retry next tick. ex={}", tex.getMessage());
-        setStatus(ObserverStatus.DATABASE_ERROR);
+        if (response.getEvents() != null && !response.getEvents().isEmpty()) {
+          processEvents(response.getEvents());
+        }
+      } finally {
+        try {
+          saveCursor(response.getCursor());
+          metricLatestBlockProcessed.set(response.getLatestLedger());
```
JiahuiWho, line 135: We need `streamBackoffTimer.reset();` here after a successful `saveCursor`. This is reporter's Fix #4

Without this, an operator could still observe slow walks toward self-shutdown that have nothing to do with the malformed-event attack.

JiahuiWho:
Note on report's Fix #5 (restrict EventFilter to known SAC addresses) — deferred

With the fixes here, the residual exposure of accepting events from any contract is log spam + RPC quota burn, not a service outage — defense-in-depth rather than load-bearing.                                                                      

Tracking as a follow-up. When picked up, SAC addresses for Stellar assets can be derived deterministically from (`asset code`, `issuer`) via `assetService.getStellarAssets()` (no operator config). Custom Soroban tokens would need an opt-in `stellar.soroban.token_contracts: [...]` allowlist that falls back to current behavior when empty.

JiahuiWho (approved):

PR #1946, opened 2026-06-01
[ANCHOR-1215]: Anchor Platform SEP-12 IDOR via unvalidated customer id allows cross-customer KYC disclosure and payout destination overwrite

amandagonsalves:
### Description

Before this change, \`Sep12Service.validateGetOrPutRequest\` enforced authorization on \`request.account\` and \`request.memo\` only. When a caller supplied \`request.id\` without \`account\`, the account check short-circuited entirely, allowing any valid SEP-10 session to \`GET\` or \`PUT\` any customer record by its \`id\` — regardless of who created it. The \`id\` was forwarded unmodified to the business-server callback, so any conforming business server would act on it.

The fix introduces a \`sep12_customer_ownership\` table. Every successful \`PUT /sep12/customer\` writes the caller's identity (\`stellar_account\`, \`memo\`) as the owner of the returned \`customer_id\`. Subsequent \`GET\` or \`PUT\` requests that supply \`id\` are checked against this table: if the stored owner does not match the token, the request is rejected with 403 before the business server is called. No business server contract changes are required.

**Changes**
- [x] \`Sep12CustomerOwnership\` / \`Sep12CustomerOwnershipStore\`: new core interfaces defining the ownership record model and its storage contract.
- [x] \`JdbcSep12CustomerOwnership\` / \`JdbcSep12CustomerOwnershipRepo\` / \`JdbcSep12CustomerOwnershipStore\`: platform JDBC implementation backed by the new \`sep12_customer_ownership\` table.
- [x] \`V29__add_sep12_customer_ownership.sql\`: Flyway migration creating the table (\`customer_id PK\`, \`stellar_account NOT NULL\`, \`memo\`).
- [x] \`Sep12Service.putCustomer\`: after every successful callback response, saves \`(customer_id, token_account, token_memo)\` to the ownership store, overwriting any prior record on update.
- [x] \`Sep12Service.validateGetOrPutRequest\`: when \`id\` is present and \`transactionId\` is absent, looks up the ownership record. If a record exists and the token's identity does not match, returns 403 before reaching the business server. If no record exists (customer predates the fix or SEP-31 first-time access), the request is allowed through with the caller's account injected, preserving backward compatibility.
- [x] \`SepBeans\` / \`DataBeans\`: wire \`Sep12CustomerOwnershipStore\` into the Spring context.
- [x] \`Sep12ServiceTest\`: mocked \`Sep12CustomerOwnershipStore\`; updated existing id-path tests; added IDOR unit tests covering mismatched account rejection, mismatched memo rejection, no-record allow-through, ownership save on PUT (plain and muxed), transactionId path bypass, and forward-account-to-callback behaviour.
- [x] \`Sep12Tests\`: added \`test cross-account id access is rejected\` integration test — registers a victim customer (ownership row written), attempts raw HTTP \`GET\` and \`PUT\` against that \`id\` from a different SEP-10 session, asserts 403 on both, and verifies the victim's record is unchanged.

**Acceptance Criteria**
- [x] \`GET /sep12/customer?id=<id>\` with a SEP-10 token that does not own that customer returns 403.
- [x] \`PUT /sep12/customer\` with \`{"id":"<id>", ...}\` from a different SEP-10 session returns 403.
- [x] The same \`GET\` and \`PUT\` succeed when issued by the token that owns the customer record.
- [x] No ownership record (first access or pre-existing customer) allows the request through with the caller's account injected.
- [x] The \`transactionId\` path is unaffected.
- [x] All existing \`Sep12ServiceTest\` tests pass.

### Context

[HackerOne #3735379](https://hackerone.com/reports/3735379)

### Testing

- Unit: \`./gradlew :core:test --tests "org.stellar.anchor.sep12.Sep12ServiceTest"\`
- Integration: \`./gradlew runEssentialTests\` — covers \`test cross-account id access is rejected\` in \`Sep12Tests\`

### Documentation
N/A

### Known limitations

Customers created before this migration has run have no ownership record. On their next \`PUT /customer\`, the record is written and all subsequent id-based requests are fully enforced. Until then, id-based access for those customers falls back to the caller's account being injected, relying on the business server's own id+account consistency check.


core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -204,10 +213,43 @@ void validateGetOrPutRequest(Sep12CustomerRequestBase requestBase, WebAuthJwt to
...
+        requestBase.setMemo(existing.getMemo());
+        requestBase.setMemoType(existing.getMemoType());
+        boolean tokenHasMemo = token.getAccountMemo() != null || token.getMuxedAccountId() != null;
+        if (existing.getAccount() == null && !tokenHasMemo) {
+          throw new SepNotAuthorizedException(ERR_CUSTOMER_ID_NOT_AUTHORIZED);
+        }
+      } catch (SepNotAuthorizedException e) {
+        throw e;
+      } catch (Exception e) {
+        throw new SepNotAuthorizedException(ERR_CUSTOMER_ID_NOT_AUTHORIZED);
```
JiahuiWho, line 245: This `catch (Exception e)` converts every prefetch failure (5xx, parsing errors, connectivity issues) into a silent 403. Add `Log.warnEx(e)` or `infoF` before the rethrow to preserve root cause without leaking it to the client.

api-schema/src/main/java/org/stellar/anchor/api/callback/GetCustomerResponse.java
```diff
@@ -24,6 +24,12 @@
 @SuperBuilder(toBuilder = true)
 public class GetCustomerResponse {
   String id;
+  String account;
+  String memo;
+
+  @SerializedName("memo_type")
+  String memoType;
+
```
JiahuiWho: The new `account`/`memo`/`memo_type` fields on `GetCustomerResponse` are a breaking change for any custom business server that doesn't return them, legitimate `id`-keyed `GET`/`PUT` will start 403'ing. 

Also these fields are deprecated per protocol: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -204,10 +213,43 @@ void validateGetOrPutRequest(Sep12CustomerRequestBase requestBase, WebAuthJwt to
       } catch (Exception e) {
         throw new SepNotAuthorizedException("The transaction specified does not exist");
       }
+    } else if (requestBase.getId() != null) {
```
JiahuiWho, line 216: To avoid the breaking change with `GetCustomerResponse`, same auth check can be done in reverse with no callback contract change: look up the customer by the token's account/memo (existing fields), and verify the returned id matches the requested one.
```
    } else if (requestBase.getId() != null) {
    isIdPath = true;
    String tokenAccount = token.getMuxedAccount() != null
        ? token.getMuxedAccount() : token.getAccount();
    String tokenMemo = token.getMuxedAccountId() != null
        ? token.getMuxedAccountId().toString() : token.getAccountMemo();
    GetCustomerResponse owned = customerIntegration.getCustomer(
        GetCustomerRequest.builder()
            .account(tokenAccount) 
            .memo(tokenMemo)
            .memoType(tokenMemo != null ? "id" : null)
            .type(requestBase.getType())
            .build());
    if (owned == null || !requestBase.getId().equals(owned.getId())) {
      throw new SepNotAuthorizedException(ERR_CUSTOMER_ID_NOT_AUTHORIZED);
    }
    requestBase.setAccount(tokenAccount);
    requestBase.setMemo(tokenMemo);
  }
```

JiahuiWho (dismissed):

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -204,10 +213,63 @@ void validateGetOrPutRequest(Sep12CustomerRequestBase requestBase, WebAuthJwt to
...
+                ? token.getMuxedAccountId().toString()
+                : token.getAccountMemo();
+
+        GetCustomerResponse owned =
+            customerIntegration.getCustomer(
+                GetCustomerRequest.builder()
+                    .id(requestBase.getId())
+                    .account(tokenAccount)
+                    .type(requestBase.getType())
+                    .build());
```
JiahuiWho, line 234: Remove `.id(requestBase.getId())` and add `.memo(tokenMemo).memoType(tokenMemo != null ? "id" : null)`. 

Reason: passing id makes us rely on the business server to enforce account matching, which the bug report says most production servers don't. Looking up by account + memo (no id) returns the caller's own customer; we then compare the returned id to the requested id. Works regardless of business-server behavior.

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -204,10 +213,63 @@ void validateGetOrPutRequest(Sep12CustomerRequestBase requestBase, WebAuthJwt to
...
+
+        if (owned == null || owned.getId() == null) {
+          throw new SepNotAuthorizedException(ERR_CUSTOMER_ID_NOT_AUTHORIZED);
+        }
+
+        if (tokenMemo != null) {
```
JiahuiWho: This if block is no longer needed because looking up by account + memo returns the caller's own customer

JiahuiWho (dismissed):

core/src/main/java/org/stellar/anchor/sep12/Sep12Service.java
```diff
@@ -204,10 +213,49 @@ void validateGetOrPutRequest(Sep12CustomerRequestBase requestBase, WebAuthJwt to
...
+        GetCustomerResponse owned =
+            customerIntegration.getCustomer(
+                GetCustomerRequest.builder()
+                    .memo(tokenMemo)
+                    .memoType(tokenMemo != null ? "id" : null)
+                    .account(tokenAccount)
+                    .type(requestBase.getType())
+                    .build());
+
+        if (owned == null || owned.getId() == null) {
```
JiahuiWho: This check passes as long as the business server returns any customer for the reverse-lookup, it doesn't verify the returned id matches the requested id. An attacker with their own customer record can still query/modify any other customer's by sending the victim's id, because their own record makes owned.getId() != null.
  
Change to:
```
if (owned == null || !requestBase.getId().equals(owned.getId())) {
  throw new  SepNotAuthorizedException(ERR_CUSTOMER_ID_NOT_AUTHORIZED);
}
```

JiahuiWho (approved):
LGTM! Don't forget to squash merge

PR #2021, opened 2026-09-14
[ANCHOR-1291]: SEP-10 multisig bypass on Horizon backend: getAccount treats non-404 4xx as account-not-found

ceciliaromao:
### Description

- `Horizon.getAccount()` now only takes the "account not found" fallback — which skips SEP-10's
  medium-threshold multisig check — on a genuine 404 response. Every other 4xx (401, 403, 400,
  etc.) now fails closed as a `LedgerException` instead.

### Context

[HackerOne #3967786](https://hackerone.com/reports/3967786)

The SEP-10 multisig fallback introduced by ANCHOR-1220 was meant to activate only for
confirmed-absent accounts. On the Horizon backend it actually caught every SDK
`BadRequestException`, which is also thrown for any 4xx other than 404/429 (e.g. a 401/403/400
from a WAF/CDN/proxy in front of Horizon). An existing multisig account whose lookup returned one
of those codes was silently downgraded to single master-key authentication with no threshold
check — re-arming the exact bypass ANCHOR-1220 fixed, on a different code path.

### Testing

- `./gradlew test`
- Added to `HorizonTest`: a genuine 404 still takes the fallback; 400/401/403/406 fail closed with
  a `LedgerException` that is specifically *not* an `AccountNotFoundException` (since that class
  extends `LedgerException`, a looser assertion would pass either way).
- Verified red/green manually: reverted the fix locally, confirmed the new non-404 test fails
  against the vulnerable code, then restored the fix and confirmed all tests pass.

### Documentation

N/A — internal fix, no public API or feature change.

### Known limitations

N/A — triage confirmed the gap is specific to the Horizon backend; the RPC backend and the SEP-45
login path already fail closed on the same class of error.


ceciliaromao (commented):
_(removed)_

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
```diff
@@ -98,7 +98,15 @@ public Account getAccount(String account) throws LedgerException {
...
-      throw new AccountNotFoundException(account);
+      // Only a genuine 404 means the account is confirmed absent. Every other 4xx (401, 403,
+      // 400, etc. — commonly emitted by a WAF/CDN/proxy in front of Horizon) must fail closed
+      // instead of being treated as "account not found," which would skip SEP-10's
+      // medium-threshold multisig check. See ANCHOR-1291.
+      Integer code = e.getCode();
+      if (code != null && code == 404) {
+        throw new AccountNotFoundException(account);
+      }
+      throw new LedgerException("Error getting account: " + account + " (HTTP " + code + ")", e);
```
ceciliaromao: 🔒 Security — Upstream HTTP status code now leaks into the client-facing SEP-10 error response

This new message format (`"Error getting account: " + account + " (HTTP " + code + ")"`) is a real behavior change, not just cosmetic: `Sep10Service.fetchAccount()` catches `LedgerException` and rethrows it as `SepValidationException("Failed to fetch account: " + ex.getMessage(), ex)`, and `SepControllerExceptionHandler` (a global `@RestControllerAdvice` extending `AbstractControllerExceptionHandler`) returns `ex.getMessage()` verbatim in the HTTP 400 body via `SepExceptionResponse`.

Before this fix, every 4xx from Horizon (including 401/403) was swallowed into the vulnerable `AccountNotFoundException("Account not found: " + account)` fallback, so this code path was never reached for those cases. Now that non-404 4xx codes correctly fail closed, the raw upstream HTTP status is newly surfaced to the (unauthenticated, pre-JWT) SEP-10 `/auth` caller — e.g. `"Failed to fetch account: Error getting account: GABC... (HTTP 403)"`. That leaks an internal implementation/infra detail (that a WAF/CDN/proxy in front of Horizon is blocking the request, and with what code) to anyone who can trigger the condition.

This isn't PII or a credential, but it's unnecessary internal diagnostic detail exposed to an unauthenticated caller, and it could aid reconnaissance against the anchor's Horizon-fronting infrastructure.

**Recommendation:** Keep the HTTP code out of the message that ends up in the client-visible exception chain — e.g. log the code server-side (`Log.warn`/`Log.error`) and throw `LedgerException` with a generic message (`"Error getting account: " + account`), or have `Sep10Service.fetchAccount()` map `LedgerException` to a generic "unable to verify account" message before it reaches `SepValidationException`.

ceciliaromao: 💡 — LedgerException message drops the response body that the sibling method includes

Rule: sibling-code consistency (`getTransaction()`, ~18 lines below in this same file)

`getTransaction()` wraps its `NetworkException` into a `LedgerException` with a structured message that includes both the HTTP code and the response body:
```java
throw new LedgerException(
    String.format(
        "Error getting transaction: %s. Code: %s, body: %s",
        txnHash, nex.getCode(), nex.getBody()),
    nex);
```
`BadRequestException` extends the same `NetworkException` and also exposes `getBody()`. The new code here instead uses plain string concatenation and omits the body:
```java
throw new LedgerException("Error getting account: " + account + " (HTTP " + code + ")", e);
```
Given this catch block's own rationale (a WAF/CDN/proxy may be emitting the non-404 4xx), the response body is exactly the diagnostic signal an operator would want to tell a real Horizon 401/403 apart from a proxy block page — but it's discarded here.

**Recommendation:** Match the `getTransaction()` pattern for consistency and diagnostics:
```java
throw new LedgerException(
    String.format(
        "Error getting account: %s. Code: %s, body: %s", account, code, e.getBody()),
    e);
```
Not a blocker — just a minor consistency/diagnostics gap between two methods in the same file.

ceciliaromao (commented):
_(removed)_

core/src/main/java/org/stellar/anchor/ledger/Horizon.java
ceciliaromao: [RESOLVED] Fixed in 28916b6d — the HTTP status/body are now logged server-side (`errorF`) instead of being embedded in the `LedgerException` message that reaches the unauthenticated SEP-10 caller via `Sep10Service.fetchAccount()`.

amandagonsalves (approved):

